PR #315, opened 2017-04-20
Add Table of Contents to FAQ (Fixes #124)

### LINKED ISSUE #124: Add a table of contents to the FAQ page

arya-s:


2Pacalypse-:
It seems like the browser just doesn't remember the location for the back button to work properly, even when using a regular anchor tag. Probably because of the client-side routing?

In any case, I feel like it's working reasonably well now. Since the react-router history gets updated fine by clicking the #hash links, we just scroll into view the element whenever the hash changes. And if the hash changes to an empty one (by pressing the back button), we scroll to the top of FAQ table of contents.

The table of contents works both on web and electron versions, except the back button doesn't work in the electron one, but that's a separate issue :d

tec27:
Does this really work in Electron? Sort of surprising to me considering we use hash routing there. It's not really a big deal if it doesn't, I guess, I've always planned to strip out the splash/FAQ stuff from the electron app but just haven't yet.

client/beta/beta.css
```diff
@@ -169,13 +169,19 @@
   padding-right: 48px;
 }
 
-.faqFeature {
+.faqFeature, .faqTOC {
```
tec27: nit: `Toc`

client/beta/beta.css
```diff
@@ -169,13 +169,19 @@
...
   composes: row itemsCenter;
   composes: pageWidth;
   padding: 48px 0;
   border-bottom: 1px solid colorDividers;
 }
 
+.faqTOC {
```
tec27: same as above, acronyms only get the first letter capitalized (e.g. `Rpc`, `Http`, etc.)

client/beta/faq.jsx
```diff
@@ -118,35 +120,82 @@ const questions = [
...
-    <div className={styles.faqFeature}>
-      <div className={styles.faqText}>
-        <QuestionIcon className={styles.faqQuestionIcon} />
-        <h3 className={styles.faqQuestion}>
-          {question}
-        </h3>
-        <div className={styles.faqAnswer}>
-          {answer}
+const makeQuestionId = question => {
+  return question.replace(/\s/g, '-')
```
tec27: should probably URL encode this

2Pacalypse-:
> Does this really work in Electron? Sort of surprising to me considering we use hash routing there. It's not really a big deal if it doesn't, I guess, I've always planned to strip out the splash/FAQ stuff from the electron app but just haven't yet.

It does work in Electron. I guess one of the router-related libraries (`history`, `react-router`, `react-router-redux`) properly sets the `location.hash` property, even in hash-based routing.

tec27:
You can fix conflicts and merge this if you want

PR #337, opened 2017-04-28
Rework the UI for joining channels.

### LINKED ISSUE #270: Add a more "content-ful" UI for joining channels
Eventually we'd want a UI with dynamic recommendations and typeahead for channel names and such, but I think for now, a basic implementation can just contain a hardcoded list of recommended channels, and a free-form input (without any sort of search) for creating/joining a channel by name.

2Pacalypse-:
Instead of a dialog, we now display an overlay with an input box where
you can enter a channel to join. It also displays a hard-coded list of
featured channels. For now that is only #ShieldBattery!

Fixes #270


2Pacalypse-:
Outdated now. Will need to be done from scratch.

PR #362, opened 2017-07-17
Fix and improve lobby tests.

### LINKED ISSUE #361: Server lobby test is broken/outdated
UMS tests mainly, but also some serialization stuff that needs to be updated

2Pacalypse-:
This commit fixes all the errors in our lobby tests, as well as
refactors some of the tests to be more readable.

There are also a few more tests added to test some of the edge cases
around the lobbies with observers and hidden slots (UMS-related).

As a part of these lobby tests fixes and improvements, there were a
couple of things changed in our lobby code as well:
- fixed a bug when finding the available slot in observer team
- make the presence of observeres in melee lobbies configurable
- make the observer slots 'closed' by default

Fixes #361

### DIFF
--- a/server/lib/wsapi/lobbies.js
+++ b/server/lib/wsapi/lobbies.js
@@ -143,7 +143,7 @@ export class LobbyApi {
     }),
   )
   async create(data, next) {
-    const { name, map, gameType, gameSubType } = data.get('body')
+    const { name, map, gameType, gameSubType, allowObservers } = data.get('body')
     const user = this.getUser(data)
     const client = this.getClient(data)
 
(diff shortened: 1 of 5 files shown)


server/lib/wsapi/lobbies.js
```diff
@@ -160,7 +160,16 @@ export class LobbyApi {
...
 
-    const lobby = Lobbies.create(name, mapData, gameType, gameSubType, numSlots, client.name)
+    const lobby = Lobbies.create(
+      name,
+      mapData,
+      gameType,
+      gameSubType,
+      numSlots,
+      client.name,
+      undefined,
```
tec27: put a comment for which parameter this is, i.e.:
`undefined, /* hostRace */`

Probably at this point we're getting to where a params object makes more sense than an ordered list, though :)

server/lib/wsapi/lobbies.js
```diff
@@ -143,7 +143,7 @@ export class LobbyApi {
     }),
   )
   async create(data, next) {
-    const { name, map, gameType, gameSubType } = data.get('body')
+    const { name, map, gameType, gameSubType, allowObservers } = data.get('body')
```
tec27: So we need to make a separate change to actually make this enable-able on the client, correct?

2Pacalypse-: Yeah, I was thinking our (new and improved!) create-lobby-ui should have a checkbox to enable this.

I guess we could make it `true` by default on server?

tec27: Or just pass it through on the client for now

server/lib/wsapi/lobbies.js
2Pacalypse-: Next time I add something to it, I'll make a params object. Promise! :d

PR #363, opened 2017-07-18
Add and set up nodemon to enable live reloading of the server.

2Pacalypse-:


tec27:
I'm not really a fan of this personally. If you want it, you can have it, but it needs to be on a command other than `start`

2Pacalypse-:
I guess I overestimated the benefits of this. I don't think they are enough to justify adding yet another script either, so just gonna give up on this.

PR #365, opened 2017-07-18
Allow resizing the observer team

neivv:
Some thoughts:

This functionality will delete/add slots in the slot list of a team, which leads to any indices becoming invalidated. It feels fragile, but I didn't see any code which uses something but the slot uids to refer slots. The tests are difficult to follow due to this though.

Additionally, when a slot gets deleted, the server has to obviously send a diff for any slots that were located after the deleted one. On the other hand, diff code hardly had to be changed at all and everything works nicely.

Good? Bad?

### DIFF
--- a/app/common/lobbies/index.js
+++ b/app/common/lobbies/index.js
@@ -165,12 +165,14 @@ export function isInObserverTeam(lobby, slot) {
 
 // Checks if the lobby has any slots that can be made observers.
 export function canAddObservers(lobby) {
+  if (!hasObservers(lobby)) return false
   const [, observerTeam] = getObserverTeam(lobby)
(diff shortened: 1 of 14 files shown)


neivv:
Oh, also, I refactored lobby creation a bit when thinking how to make slot creation/deletion simpler for UMS, even though we don't support UMS observing anyways. I think the changes are good regardless.

2Pacalypse-:
I'll take a closer look tomorrow, but the general approach seems good.

I was worried about code being fragile as well when I was planning on working this before, since this feature invalidates some of the assumptions we had, eg. team size in a lobby being fixed etc. We'll need to test this code around the edges to make sure there aren't any looming bugs lurking around :d

PR #371, opened 2017-07-25
Copy the local.dll file from StarCraft folder to our downgrade folder.

2Pacalypse-:
Since 1.19 version, local.dll file is not located in the root StarCraft
folder anymore, but instead in one of the subfolders. To make it easier
for us to use it, we copy it over to our downgrade folder and we'll make
StarCraft look for it there, which will come in a separate PR.

### DIFF
--- a/client/settings/check-starcraft-path.js
+++ b/client/settings/check-starcraft-path.js
@@ -29,6 +29,15 @@ async function checkHash(path, validHashes) {
   return validHashes.includes(hash)
 }
 
+async function checkFileExists(path) {
+  try {
+    await accessAsync(path, fs.constants.R_OK)
+    return true
+  } catch (err) {
+    return false
+  }
+}
+
 // Returns whether or not a StarCraft path is valid, along with whether or not the versions of
 // files contained in it are the expected versions. A path is valid if it contains:
 //   - StarCraft.exe
@@ -53,13 +62,12 @@ export async function checkStarcraftPath(dirPath, downgradePath) {
   }
 
   // Due to 1.19 version moving local.dll to a separate folder, we need to handle it separately
-  let matches = await globAsync(`${dirPath}/**/local.dll`)
-  if (matches.length < 1) {
+  let localDllValid = await checkFileExists(path.join(dirPath, 'local.dll'))
+  const matches = await globAsync(`${dirPath}/locales/*/local.dll`)
+  if (!localDllValid && matches.length < 1) {
     return { path: false, version: false, downgradePath: false }
   }
 
-  matches = matches.map(match => path.normalize(match))
-  let localDllValid = matches.includes(path.normalize(`${dirPath}/local.dll`))
   let [starcraftValid, stormValid] = await Promise.all([
     checkHash(path.join(dirPath, 'starcraft.exe'), EXE_HASHES_1161),
     checkHash(path.join(dirPath, 'storm.dll'), STORM_HASHES_1161),
@@ -69,13 +77,11 @@ export async function checkStarcraftPath(dirPath, downgradePath) {
     return { path: true, version: true, downgradePath: false }
   }
 
+  localDllValid = await checkFileExists(path.join(downgradePath, 'local.dll'))
   ;[starcraftValid, stormValid] = await Promise.all([
     checkHash(path.join(downgradePath, 'starcraft.exe'), EXE_HASHES_1161),
     checkHash(path.join(downgradePath, 'storm.dll'), STORM_HASHES_1161),
   ])
-
-  matches = await globAsync(`${downgradePath}/local.dll`)
-  localDllValid = matches.length > 0
 
   if (starcraftValid && stormValid && localDllValid) {
     return { path: true, version: true, downgradePath: true }
--- a/client/settings/check-starcraft-path.js
+++ b/client/settings/check-starcraft-path.js
@@ -6,7 +6,6 @@ import glob from 'glob'
 import thenify from 'thenify'
 import logger from '../logging/logger'
 import getFileHash from '../../app/common/get-file-hash'
-import { streamEndPromise, streamFinishPromise } from '../../app/common/async/stream-promise'
 
 const accessAsync = thenify(fs.access)
 const globAsync = thenify(glob)
@@ -28,19 +27,6 @@ async function checkHash(path, validHashes) {
   }
 
   return validHashes.includes(hash)
-}
-
-async function maybeCopyLocalDll(dllPath, downgradePath) {
-  const outFile = path.join(downgradePath, 'local.dll')
-  // Check if the local.dll file already exists in our downgradePath
-  const matches = await globAsync(outFile)
-  if (matches.length > 0) return
-
-  // local.dll file doesn't exist; copy it over
-  const inStream = fs.createReadStream(dllPath)
-  const outStream = fs.createWriteStream(outFile)
-  inStream.pipe(outStream)
-  await Promise.all([streamEndPromise(inStream), streamFinishPromise(outStream)])
 }
 
 // Returns whether or not a StarCraft path is valid, along with whether or not the versions of
@@ -67,19 +53,19 @@ export async function checkStarcraftPath(dirPath, downgradePath) {
   }
 
   // Due to 1.19 version moving local.dll to a separate folder, we need to handle it separately
-  const matches = await globAsync(`${dirPath}/**/local.dll`)
-  if (matches.length > 0) {
-    maybeCopyLocalDll(matches[0], downgradePath)
-  } else {
+  let matches = await globAsync(`${dirPath}/**/local.dll`)
+  if (matches.length < 1) {
     return { path: false, version: false, downgradePath: false }
   }
 
+  matches = matches.map(match => path.normalize(match))
+  let localDllValid = matches.includes(path.normalize(`${dirPath}/local.dll`))
   let [starcraftValid, stormValid] = await Promise.all([
     checkHash(path.join(dirPath, 'starcraft.exe'), EXE_HASHES_1161),
     checkHash(path.join(dirPath, 'storm.dll'), STORM_HASHES_1161),
   ])
 
-  if (starcraftValid && stormValid) {
+  if (starcraftValid && stormValid && localDllValid) {
     return { path: true, version: true, downgradePath: false }
   }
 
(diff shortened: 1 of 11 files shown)


client/settings/check-starcraft-path.js
```diff
@@ -42,14 +58,22 @@ async function checkHash(path, validHashes) {
...
   try {
     await Promise.all(requiredFiles.map(f => accessAsync(path.join(dirPath, f), fs.constants.R_OK)))
   } catch (err) {
     return { path: false, version: false, downgradePath: false }
   }
 
+  // Due to 1.19 version moving local.dll to a separate folder, we need to handle it separately
+  const matches = await globAsync(`${dirPath}/**/local.dll`)
+  if (matches.length > 0) {
+    maybeCopyLocalDll(matches[0], downgradePath)
```
tec27: You're performing a filesystem mutation in what was previously pure, this isn't good. On top of that, I don't think we should be copying this unless we need it (that is, unless the file is in the wrong spot, or the other files require us to downgrade). Thus, this method should be updated to simply check for local.dll and mark the version as wrong if it's not found (as an added check to the existing ones here), and the file that deals with downgrading should actually perform any necessary copying.

2Pacalypse-: Hmm, I thought you said we should copy the local.dll file to our downgrade folder for all BW versions. I assumed that meant even for 1.16.1 - 1.18 where the local.dll was in the correct spot (BW's root folder). Are you saying we shouldn't copy the local.dll to our downgrade folder if it's located in the BW's root folder?

tec27: We don't do anything for 1.16.1 installs, we should continue not doing anything for those. I meant that you should copy it *any time we downgrade*, not always. And you definitely *should not* be copying as part of something that is just meant to check whether an install is valid

client/settings/check-starcraft-path.js
```diff
@@ -42,20 +44,28 @@ async function checkHash(path, validHashes) {
...
-  const requiredFiles = ['starcraft.exe', 'storm.dll', 'local.dll', 'stardat.mpq', 'broodat.mpq']
+  const requiredFiles = ['starcraft.exe', 'storm.dll', 'stardat.mpq', 'broodat.mpq']
 
   try {
     await Promise.all(requiredFiles.map(f => accessAsync(path.join(dirPath, f), fs.constants.R_OK)))
   } catch (err) {
     return { path: false, version: false, downgradePath: false }
   }
 
+  // Due to 1.19 version moving local.dll to a separate folder, we need to handle it separately
```
tec27: Just do 2 checks:
- In the root folder (aka valid for 1.16)
- At locales/*/local.dll

Globbing everything under StarCraft's folder is potentially expensive, people have large map collections and such. This setup also avoids the need to check for the first case in the matches, which seems very error prone

client/settings/check-starcraft-path.js
```diff
@@ -64,7 +74,10 @@ export async function checkStarcraftPath(dirPath, downgradePath) {
     checkHash(path.join(downgradePath, 'storm.dll'), STORM_HASHES_1161),
   ])
 
-  if (starcraftValid && stormValid) {
+  matches = await globAsync(`${downgradePath}/local.dll`)
```
tec27: Why are you globbing this, exactly? There's no pattern here whatsoever, just access it like we do above

client/settings/patch-starcraft.js
```diff
@@ -89,11 +95,23 @@ async function patchFile(dirPath, outPath, filename, validHashes) {
   }
 }
 
+async function copyLocalDll(dirPath, downgradePath) {
+  const matches = await globAsync(`${dirPath}/**/local.dll`)
```
tec27: Apply the same advice in the check function to here as far as finding the DLL.

This also potentially causes problems if they have multiple locales installed (and we pick the local.dll for a locale they don't want). I think that's okay for now, especially if we plan on letting people select locales down the road and replace local.dll, but something to be aware of.

client/settings/patch-starcraft.js
```diff
@@ -61,10 +70,7 @@ async function patchFile(dirPath, outPath, filename, validHashes) {
   const [inValid, inHash] = await checkFileHash(inFile, validHashes)
   if (inValid) {
     // the one in the folder is already valid, we can just copy it over
-    const inStream = fs.createReadStream(inFile)
-    const outStream = fs.createWriteStream(outFile)
-    inStream.pipe(outStream)
-    await Promise.all([streamEndPromise(inStream), streamFinishPromise(outStream)])
+    copyFile(inFile, outFile)
```
tec27: So if you don't await this, this function's gonna return before it finishes ;)

client/settings/patch-starcraft.js
```diff
@@ -89,11 +95,23 @@ async function patchFile(dirPath, outPath, filename, validHashes) {
...
 
+async function copyLocalDll(dirPath, downgradePath) {
+  const matches = await globAsync(`${dirPath}/**/local.dll`)
+  if (matches.length < 1) {
+    throw new Error("local.dll doesn't exist in StarCraft directory")
+  }
+  const inFile = matches[0]
+  const outFile = path.join(downgradePath, 'local.dll')
+
+  copyFile(inFile, outFile)
```
tec27: Same as above, you need to await this

game/common/win_helpers.cpp
```diff
@@ -131,4 +131,13 @@ wstring GetDocumentsPath() {
   return documents_path;
 }
 
+bool EndsWith(const string checked, const string suffix) {
```
tec27: This has nothing to do with Windows, move it to its own file

game/shieldbattery/shieldbattery.cpp
```diff
@@ -384,13 +386,31 @@ DWORD __stdcall GetModuleFileNameAHook(
   }
 }
 
+HMODULE __stdcall LoadLibraryAHook(_In_ LPCTSTR filename) {
+  if (EndsWith(filename, "local.dll")) {
+    wchar_t starcraft_dir[MAX_PATH];
+    DWORD count = GetCurrentDirectoryW(sizeof(starcraft_dir), starcraft_dir);
```
tec27: We have a path for this, its `current_dir_on_inject` or `real_starcraft_path`. You shouldn't rely on the current directory just happening to be the StarCraft path, I very particularly wrote this such that we save it off at the earliest possible time (before anything has an opportunity to change it). It is absolutely not okay to re-check it at this point in time.

This code is needlessly complex, really, you can just:
1) Check if they're loading local.dll
2) Call GetModuleFileNameW(NULL) [returns the filepath for starcraft.exe]
3) Replace the file name in that path with 'local.dll'
4) Return LoadLibraryW result of that path

There's only two cases here:
- We're using a downgrade version, in which case local.dll is beside it in the downgrade folder
- We're not using a downgrade version, in which case local.dll is beside it in the StarCraft folder

As you can see, in both cases, local.dll is always directly beside StarCraft.exe. If we didn't need the GetModuleFileNameA hook, hooking LoadLibrary here wouldn't be necessary at all, as the code that StarCraft runs would already try to load the right path.

client/settings/check-starcraft-path.js
```diff
@@ -42,29 +53,37 @@ async function checkHash(path, validHashes) {
...
     checkHash(path.join(dirPath, 'starcraft.exe'), EXE_HASHES_1161),
     checkHash(path.join(dirPath, 'storm.dll'), STORM_HASHES_1161),
   ])
 
-  if (starcraftValid && stormValid) {
+  if (starcraftValid && stormValid && localDllValid) {
     return { path: true, version: true, downgradePath: false }
   }
 
+  localDllValid = await checkFileExists(path.join(downgradePath, 'local.dll'))
```
tec27: This makes this a 2 step process instead of executing in parallel, you need to put this in the Promise.all call so they all get kicked off at once

client/settings/check-starcraft-path.js
```diff
@@ -42,29 +53,37 @@ async function checkHash(path, validHashes) {
...
 
   try {
     await Promise.all(requiredFiles.map(f => accessAsync(path.join(dirPath, f), fs.constants.R_OK)))
   } catch (err) {
     return { path: false, version: false, downgradePath: false }
   }
 
+  // Due to 1.19 version moving local.dll to a separate folder, we need to handle it separately
+  let localDllValid = await checkFileExists(path.join(dirPath, 'local.dll'))
+  const matches = await globAsync(`${dirPath}/locales/*/local.dll`)
```
tec27: Only do this if localDllValid is false

client/settings/patch-starcraft.js
```diff
@@ -89,11 +95,23 @@ async function patchFile(dirPath, outPath, filename, validHashes) {
   }
 }
 
+async function copyLocalDll(dirPath, downgradePath) {
+  const matches = await globAsync(`${dirPath}/locales/*/local.dll`)
+  if (matches.length < 1) {
+    throw new Error("local.dll doesn't exist in StarCraft directory")
```
tec27: No, this isn't the only possible case. There are 2 cases:
- local.dll is in the root folder
- local.dll is in the locales/*/ folder

The first will happen for all version < 1.19, the latter will happen for 1.19 or later. We support all such versions, so you need to handle both cases.

game/shieldbattery/shieldbattery.cpp
```diff
@@ -384,13 +386,26 @@ DWORD __stdcall GetModuleFileNameAHook(
   }
 }
 
+HMODULE __stdcall LoadLibraryAHook(_In_ LPCTSTR filename) {
+  if (EndsWith(filename, "local.dll")) {
+    wchar_t starcraft_path[MAX_PATH];
+    GetModuleFileNameW(NULL, starcraft_path, sizeof(starcraft_path));
+    wstring local_dll_path = wstring(starcraft_path);
+    local_dll_path.replace(local_dll_path.end() - strlen("starcraft.exe"), local_dll_path.end(), L"local.dll");
```
tec27: You can't assume the exe name like this. You also can't assume that the filename won't appear in the filepath leading up to the filename. There is code *in this file* that does this properly. Hint: StartNode

game/shieldbattery/shieldbattery.cpp
```diff
@@ -384,13 +386,26 @@ DWORD __stdcall GetModuleFileNameAHook(
   }
 }
 
+HMODULE __stdcall LoadLibraryAHook(_In_ LPCTSTR filename) {
+  if (EndsWith(filename, "local.dll")) {
+    wchar_t starcraft_path[MAX_PATH];
```
tec27: `exe_path`, `starcraft_path` is very overloaded in meanings here

PR #373, opened 2017-07-27
Hook the LoadLibraryA function to properly load local.dll library.

2Pacalypse-:
When BW tries to load the local.dll library, we need to figure out
whether to load it from StarCraft's root directory (if exists), or from
our downgrade folder (for 1.19+ versions).


tec27:
Can you just put everything related to this in a single PR, instead of creating a chain of dependent  ones?

2Pacalypse-:
Sure, I moved this commit to #371

PR #374, opened 2017-07-27
Fade the atmosphere sound if the lobby has canceled loading.

2Pacalypse-:


client/lobbies/socket-handlers.js
```diff
@@ -230,6 +230,7 @@ const eventToAction = {
   },
 
   cancelLoading: (name, event) => dispatch => {
+    clearCountdownTimer()
```
tec27: fadeAtmosphere seems like the more correct call, the countdown would never actually be running here

PR #377, opened 2017-07-31
Implement the server side necessary for email verification.

2Pacalypse-:
Everytime a new account is created, we send an account verification
email to the user. There are two new routes added in this commit; one to
actually verify an user's email, and one to resend the verification
email.

### DIFF
--- a/server/lib/api/users.js
+++ b/server/lib/api/users.js
@@ -1,13 +1,16 @@
 import bcrypt from 'bcrypt'
+import cuid from 'cuid'
 import httpErrors from 'http-errors'
 import thenify from 'thenify'
 import createThrottle from '../throttle/create-throttle'
 import throttleMiddleware from '../throttle/middleware'
 import users from '../models/users'
 import initSession from '../session/init'
+import sendAccountVerificationEmail from '../verifications/send-account-verification-email'
 import setReturningCookie from '../session/set-returning-cookie'
 import { checkAnyPermission } from '../permissions/check-permissions'
 import { usePasswordResetCode } from '../models/password-resets'
+import { addEmailVerificationCode, useEmailVerificationCode } from '../models/email-verifications'
 import { isValidUsername, isValidEmail, isValidPassword } from '../../../app/common/constants'
 import { UNIQUE_VIOLATION } from '../db/pg-error-codes'
 import transact from '../db/transaction'
@@ -27,6 +30,8 @@ export default function(router) {
       throw new httpErrors.ImATeapot()
     })
     .post('/:username/password', resetPassword)
+    .post('/:id/emailVerification', verifyEmail)
+    .post('/:id/resendVerification', resendVerificationEmail)
 }
 
 async function find(ctx, next) {
@@ -71,6 +76,10 @@ async function createUser(ctx, next) {
   await ctx.regenerateSession()
   initSession(ctx, result.user, result.permissions)
   setReturningCookie(ctx)
+
+  const code = cuid()
+  await addEmailVerificationCode(result.user.id, email, code, ctx.ip)
+  await sendAccountVerificationEmail(email, code)
   ctx.body = result
 }
 
@@ -95,4 +104,34 @@ async function resetPassword(ctx, next) {
     await user.save()
     ctx.status = 204
   })
+}
+
+async function verifyEmail(ctx, next) {
+  const { id } = ctx.params
+  const { code } = ctx.query
+  const { email } = ctx.request.body
+
+  if (!id || !code || !isValidEmail(email)) {
+    throw new httpErrors.BadRequest('Invalid parameters')
+  }
+
+  const verifiedEmail = await useEmailVerificationCode(id, email, code)
+  if (!verifiedEmail) {
+    throw new httpErrors.BadRequest('Email verification code is invalid')
+  }
+  ctx.body = verifiedEmail
+}
+
+async function resendVerificationEmail(ctx, next) {
+  const { id } = ctx.params
+  const { email } = ctx.request.body
+
+  if (!id || !isValidEmail(email)) {
+    throw new httpErrors.BadRequest('Invalid parameters')
+  }
+
+  const code = cuid()
+  await addEmailVerificationCode(id, email, code, ctx.ip)
+  await sendAccountVerificationEmail(email, code)
+  ctx.status = 204
 }
--- a/server/lib/api/users.js
+++ b/server/lib/api/users.js
@@ -18,6 +18,8 @@ import {
 import { isValidUsername, isValidEmail, isValidPassword } from '../../../app/common/constants'
 import { UNIQUE_VIOLATION } from '../db/pg-error-codes'
 import transact from '../db/transaction'
+import redis from '../redis'
+import sessionStore from '../session/session-store'
 
 const accountCreationThrottle = createThrottle('accountcreation', {
   rate: 1,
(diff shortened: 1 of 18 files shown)


server/lib/api/users.js
```diff
@@ -97,3 +106,33 @@ async function resetPassword(ctx, next) {
     ctx.status = 204
   })
 }
+
+async function verifyEmail(ctx, next) {
```
tec27: This needs to be rate limited to prevent brute-force attempts

server/lib/api/users.js
```diff
@@ -97,3 +106,33 @@ async function resetPassword(ctx, next) {
...
+  }
+
+  const verifiedEmail = await useEmailVerificationCode(id, email, code)
+  if (!verifiedEmail) {
+    throw new httpErrors.BadRequest('Email verification code is invalid')
+  }
+  ctx.body = verifiedEmail
+}
+
+async function resendVerificationEmail(ctx, next) {
```
tec27: This needs to be rate-limited very heavily to prevent using it as a spam vector. We may want to consider some maximum amount of times we'd send email to a particular address without verification (e.g. after N attempts you are required to specify a different email address, rather than being able to re-send it).

server/mail-templates/email-verification.html
```diff
@@ -0,0 +1,205 @@
...
+      <div class="content">
+        <table class="main" width="100%" cellpadding="0" cellspacing="0">
+          <tr>
+            <td class="content-wrap">
+              <table width="100%" cellpadding="0" cellspacing="0">
+                <tr>
+                  <td class="content-block">
+                    <h1>ShieldBattery Email Verification</h1>
+                    <p>
+                      Thank you for signing up to the ShieldBattery!
```
tec27: `to ShieldBattery` or `for ShieldBattery`

server/lib/api/users.js
```diff
@@ -97,3 +106,33 @@ async function resetPassword(ctx, next) {
...
+  const { code } = ctx.query
+  const { email } = ctx.request.body
+
+  if (!id || !code || !isValidEmail(email)) {
+    throw new httpErrors.BadRequest('Invalid parameters')
+  }
+
+  const verifiedEmail = await useEmailVerificationCode(id, email, code)
+  if (!verifiedEmail) {
+    throw new httpErrors.BadRequest('Email verification code is invalid')
```
tec27: When you implement this on the client, you'll probably want to mention that the code might be expired.

2Pacalypse-: Since we don't have a sensible error system yet, how do I check it was this error on the client side that happened?

tec27: You could check the error code + message contents. Or, since we know the client validates the presence of most of things, you could just always act as if this error (by "this error" I mean a 400/Bad Request) means invalid email or verification code for now.

All I mean is that the message to the client should be something like:
'The provided email or verification code is not valid. If the verification code matches the one you were emailed, it may have expired. Please request a new verification email and try again.'

server/lib/models/email-verifications.js
```diff
@@ -0,0 +1,44 @@
...
+  try {
+    await client.query(query, params)
+  } finally {
+    done()
+  }
+}
+
+export async function useEmailVerificationCode(id, email, code) {
+  const query = `
+    UPDATE users u
```
tec27: This query should ensure that the user's email matches the one we're verifying for

server/lib/api/users.js
2Pacalypse-: I went to do this, but I have no idea if there's any rationale behind choosing the rate/burst/window values. All of these values seem like they're pulled out of a hat to me.

tec27: They are all arbitrary, yes. Generally you can find/choose similar values for rate-limiting similar things. Things that cause emails to be sent should be rate-limited most stringently, things that are mainly to prevent brute-forcing (and have consequences for other users) can allow far more requests. It's a judgment call: how much do you expect a particular entity to call something in the normal case? How many calls would be considered very exceptional? Would we expect the exceptional case to continue for a prolonged period, or only in a single burst (and then return to the normal rate)?

tec27: And in cases where we limit on IP and not user, you'd want to consider what effects multiple machines behind the same NAT would have (but for this particular case, limiting on user+email combo seems fine tbh)

server/lib/api/users.js
```diff
@@ -96,3 +139,71 @@ async function resetPassword(ctx, next) {
...
+
+  ctx.status = 204
+
+  // If the user is not logged in, our job is done here. They will receive the updated data with
+  // `emailVerified` set to `true` next time they log in.
+  if (!ctx.session.userId) return
+
+  // If the user is logged in when verifying their email, update all of their sessions to indicate
+  // that their email is now indeed verified.
+  const userSessionsKey = 'user_sessions:' + ctx.session.userId
```
tec27: You should probably pull all this session updating stuff out into a separate function so that this code doesn't proliferate all over the codebase. Something like `updateAllSessions(userId, updatedValues)` (the `sessionStore.set` call can then just pass `{ ...session, ...updatedValues }`)

server/lib/api/users.js
```diff
@@ -96,3 +139,71 @@ async function resetPassword(ctx, next) {
...
+  // unupdated. So we update it manually here.
+  ctx.session.emailVerified = true
+
+  // Last thing to do is to notify all of the user's opened sockets that their email is now verified
+  // NOTE(2Pac): With the way the things are currently set up on client (their socket is not
+  // connected when they open the email verification page), the client actually making the request
+  // won't get this event. Thankfully, that's easy to deal with on the client-side.
+  const user = userSockets.getByName(ctx.session.userName)
+  if (user) {
+    user.subscribe('/auth/emailVerification', () => ({ action: 'emailVerified' }))
```
tec27: This is a really absurd way to do this, users should all be subscribed to a relevant path by default (one that has an identifying piece of information like their username or ID), and you should publish to that route when it changes. Don't abuse subscriptions as a way to publish

server/lib/api/users.js
```diff
@@ -96,3 +139,71 @@ async function resetPassword(ctx, next) {
...
+  const emailVerified = await useEmailVerificationCode(id, email, code)
+  if (!emailVerified) {
+    throw new httpErrors.BadRequest('Email verification code is invalid')
+  }
+
+  ctx.status = 204
+
+  // If the user is not logged in, our job is done here. They will receive the updated data with
+  // `emailVerified` set to `true` next time they log in.
+  if (!ctx.session.userId) return
```
tec27: This feels like a possible attack vector to me, perhaps you should enforce that you have to be logged in to verify an email?

server/lib/api/users.js
2Pacalypse-: Yeah, sorry about that; I tried to cheat a bit here :d

server/lib/api/users.js
2Pacalypse-: Hmm, I thought it was usual to allow email verification without being logged in. What exact attack vector do you see potentially happening here?

server/lib/api/users.js
2Pacalypse-: I tried to subscribe a user to `/users/:userId` path for this, but then I saw we're already using that route for user's subscriptions: https://github.com/ShieldBattery/ShieldBattery/blob/master/client/loading/socket-handlers.js#L42

Should I reuse that path for notifying user's their email has been verified, or should I create a new path? I assume the latter, since we could probably reuse this path for all user's profile updates. Just not sure what would the appropriate path be :d

PR #380, opened 2017-10-15
Upgrade react-router and refactor the app to work with new version

2Pacalypse-:
New version of react-router is a complete rewrite from previous
versions, so a lot of things changed.

Most notably, react-router's new philosophy is dynamic routes, which
means we don't have a file where we statically define routes
configuration in advance. Routes are declared throughout the site,
wherever you need something rendered based on a path.

Another big change is in the Route component itself and the way route
matching works. Route nesting is now discouraged, which does make
figuring out what should render more straight-forward, even if it
requires a bit more work to use.


client/auth/signup.jsx
```diff
@@ -181,10 +179,10 @@ export default class Signup extends React.Component {
   }
 
   onLogInClick = () => {
-    const query = {
+    const search = queryString.stringify({
```
2Pacalypse-: Note to self: this is unnecessary since `this.props.location.search` is already string (oh and it's `search`, not `query` anymore).

Will fix when I come home from work.

package.json
```diff
@@ -65,15 +65,17 @@
...
     "rally-point-player": "^1.0.0",
     "react": "^15.4.1",
     "react-addons-css-transition-group": "^15.4.1",
     "react-custom-scrollbars": "^4.0.0",
     "react-dom": "^15.4.1",
     "react-ga": "^2.1.2",
     "react-redux": "^5.0.1",
-    "react-router": "^2.6.0",
-    "react-router-redux": "^4.0.5",
+    "react-router": "^4.2.0",
```
2Pacalypse-: Note to self: Can remove this dependency since we're using `react-router-dom` to import things.

client/app.jsx
```diff
@@ -1,16 +1,41 @@
...
+}
+
 export default class App extends React.Component {
   initialized = false
   onUpdate = () => {
     if (!this.initialized) {
       return
     }
 
+    // TODO(2Pac): Make GA work.
```
tec27: Should file an issue for this if we don't have one already

client/app.jsx
```diff
@@ -31,9 +56,23 @@ export default class App extends React.Component {
...
+      <Switch>
+        <Route path="/splash" component={Splash} />
+        <Route path="/faq" component={Faq} />
+        <Route path="/download" component={DownloadPage} />
+        <LoginRoute path="/forgot-password" component={ForgotPassword} />
+        <LoginRoute path="/forgot-user" component={ForgotUser} />
+        <LoginRoute path="/login" component={Login} />
+        <LoginRoute path="/reset-password" component={ResetPassword} />
+        <LoginRoute path="/signup" component={Signup} />
+        {devRoutes}
```
tec27: Would be nice to just put these all under `/dev/` or something and have a Dev component we pull in that decides whether or not to expose them. Then you'd just need logic to decide whether to add that here, and not deal with building the actual list of dev routes in this file.

I can see you've *sort of* done that by putting them in different subject areas, but it'd probably be better to just have one Dev component at the root that we include (or don't) depending (or perhaps that just either returns a Switch or null depending? Would make the logic here simpler).

client/whispers/whisper.jsx
```diff
@@ -90,11 +90,11 @@ class Whisper extends React.Component {
 function isClosingCurrentWhisperSession(oldProps, newProps) {
```
tec27: Not something you changed here, but these sorts of functions seems to be a prime refactoring target, to me. We duplicate them for basically every subject area with slight differences, and have to check them in a number of spots. Seems like we could make a wrapper component that would do this check (and receive the data from the store), leaving a pure component inside that receives that data on props and could not deal with that stuff?

Dunno, just something to think about for the future

PR #382, opened 2017-10-29
Move the Dialog component to Portals.

### LINKED ISSUE #299: Change the click-intercepting logic for portals to always take effect regardless of scrim setting
Right now when you have a select component opened inside a dialog or an overlay, doing click outside the dialog/overlay will close them as well, rather than just closing the select component.

Fixing this probably involves moving the dialogs and overlays to use the portal component as well, but I'm not sure.

2Pacalypse-:
Note: Modal dialogs are not working in this commit. Need to rethink how
the modal dialogs/portals should work and fix it.

Fixes #299

### DIFF
--- a/client/dialogs/connected-dialog-overlay.jsx
+++ b/client/dialogs/connected-dialog-overlay.jsx
@@ -16,52 +16,39 @@ import AcceptMatch from '../matchmaking/accept-match.jsx'
 import { closeDialog } from './dialog-action-creator'
 
 const transitionNames = {
-  appear: styles.enter,
-  appearActive: styles.enterActive,
   enter: styles.enter,
   enterActive: styles.enterActive,
   leave: styles.leave,
   leaveActive: styles.leaveActive,
 }
 
-const CLOSE_TIME = 250
-
 @connect(state => ({ dialog: state.dialog }))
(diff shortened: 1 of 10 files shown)


client/material/portal.css
```diff
@@ -5,6 +5,7 @@
 .portal {
   contain: layout style;
   position: relative;
+  z-index: 0;
```
tec27: probably wanna comment about stacking here as well (like you did in the global CSS)

client/dialogs/connected-dialog-overlay.jsx
```diff
@@ -27,39 +29,39 @@ class ConnectedDialogOverlay extends React.Component {
     this._focusable = elem
   }
 
-  getDialogComponent(dialogType) {
+  getDialog(dialogType) {
     switch (dialogType) {
       case 'acceptMatch':
-        return AcceptMatch
+        return { component: AcceptMatch, modal: true }
```
tec27: A little sad we lost the ability to specify modality on the dialog component itself. I feel like there's still a reasonable way to do that with render props or something, but oh well, not worth trying to fix as part of this PR. We can refactor it later.

PR #385, opened 2017-12-23
Update the color scheme.

2Pacalypse-:
This commit changes the color scheme slightly, and also reduces the
amount of colors we have defined, as per new material design
specification.

Another thing that was changes is the nomenclature of our color names,
which is now: `primary` and `secondary`


2Pacalypse-:
Gonna close this for now since the app is already updated to the latest mocks. Also, the colors defined here need some tweaking anyway.

PR #393, opened 2018-04-01
This commit updates the design of the app to match the new mocks.

2Pacalypse-:
Few notes:
- App bar has been reworked to accomodate new design
- The title and user count in the app bar are selectable text,
everything else acts like a top bar of an actual window
- Window Controls are now rendered on top of everything, even
scrim/loading filters
- A bunch of stuff has been rewritten in styled-components, most notably
the avatar stuff

TODO:
- Logo in the lockup needs to get updated to match the app bar color
- color scheme needs to get updated to match the new mocks (requires
experimentation)
- Profile link can only be accessed through the avatar button for now,
because making the username clickable as well felt weird (requires
experimentation)


client/admin/app-bar-title.jsx
```diff
@@ -0,0 +1,26 @@
...
+
+import { AppBarTitle } from '../app-bar/app-bar.jsx'
+
+@connect(state => ({ routing: state.routing }))
+export default class AdminTitle extends React.Component {
+  render() {
+    const { routing: { location: { pathname } } } = this.props
+
+    let appBarTitle
+    if (pathname === '/admin') {
```
tec27: You've probably already thought about this, but it'd be really nice if this could be specified on the route entries in some way

PR #400, opened 2019-04-03
Make starcraft.exe rusty

neivv:
Things that still need to be done (the very least)
- [x] Figuring out how to make the building / distribution simple (Right now it just accesses debug build directory)
- [x] Instructions for however the above step turns out
- [x] Observing
- [x] UMS ? Not sure if it works already (Update: It did)
- [x] Team melee?
- [x] Review other patches to starcraft.exe that the old dll applied
- [x] Moving the game thread code away from lib.rs
- [x] Decide if OpenGL support is necessary (IMO it's not worth the pain)
- [x] Clean up the async code I wrote first :p
- [x] Check for any leftover TODO/unimplemented
- [x] Delete the old code

---
How to build:
https://github.com/ShieldBattery/ShieldBattery/blob/2c3a87b77156f4bb6124dc1298eeb93e32cdfe06/docs/GETTING_STARTED.md#rust


client/active-game/game-server.js
```diff
@@ -1,80 +1,97 @@
...
+        this.idToSocket.set(gameId, socket)
+        activeGameManager.handleGameConnected(gameId)
+      }
+    })
   }
 
-  async function onReplaySave(data, next) {
-    const { path } = data.get('body')
-    activeGameManager.handleReplaySave(data.get('gameId'), path)
+  sendCommand(socket, command, payload) {
```
2Pacalypse-: Maybe put underscore before the name of this method to indicate it should stay private? Otherwise, this method should probably be the one that receives the `gameId` and fetches socket based on that and checks if the socket is still open/valid.

client/active-game/game-server.js
```diff
@@ -1,80 +1,97 @@
...
-  })
 }
 
 function makeGameServer() {
   if (!IS_ELECTRON) {
     return null
   }
 
-  const nydus = require('nydus').default
+  const WebSocket = require('ws/index.js')
```
2Pacalypse-: This import could really use a comment on why we need to import `index.js` directly (to workaround the `browser` field in `ws`'s `package.json` which webpack uses for Electron apps; *smh*).

game/rust-dll/src/client_socket.rs
```diff
@@ -0,0 +1,272 @@
...
+use tokio::prelude::*;
+use tokio::sync::mpsc;
+use tokio::timer::Delay;
+use websocket::{self, OwnedMessage};
+use websocket::r#async::client::{ClientNew, TcpStream};
+
+use crate::{AsyncSenders, AsyncMessage, box_future};
+use crate::game_state::{GameStateMessage};
+
+fn connect_to_client() -> ClientNew<TcpStream> {
```
2Pacalypse-: Isn't the Electron app the one hosting the server, so this should be `connect_to_server`?

game/rust-dll/src/client_socket.rs
```diff
@@ -0,0 +1,272 @@
...
+                        .map_err(|_| ());
+                    let (sink, stream) = client.split();
+                    let stream_done = stream
+                        .filter_map(|message| {
+                            match message {
+                                OwnedMessage::Text(text) => {
+                                    match handle_client_message(text) {
+                                        Ok(o) => Some(o),
+                                        Err(e) => {
+                                            error!("Error handling message: {}", e);
```
2Pacalypse-: That's a lot of indent :smile: Hope it gets better in _future_.

client/active-game/active-game-manager.js
```diff
@@ -197,7 +198,11 @@ function silentTerminate(proc) {
   }
 }
 
-const injectPath = path.resolve(remote.app.getAppPath(), '../game/dist/shieldbattery.dll')
+// TODO(neive): how to select between local dev/installed dll?
+const injectPath = path.resolve(
+  remote.app.getAppPath(),
+  '../game/rust-dll/target/i686-pc-windows-msvc/debug/shieldbattery_rust.dll',
```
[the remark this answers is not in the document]
neivv: Right, adding a .bat script orsuch to copy the dll to `dist` would likely be the simplest way to do things

game/rust-dll/src/client_socket.rs
neivv: Yeah, probably it should be changed to `app` or such - I was thinking in the sense of "The electron *client* to shieldbattery service is what we're connecting to" And avoided calling anything a client/server outside of what the library calls used.

client/settings/settings.jsx
```diff
@@ -94,10 +94,6 @@ class SettingsForm extends React.Component {
           disabled={!this.isFullscreen()}
           inputProps={{ tabIndex: 0 }}
         />
-        <Select {...bindCustom('renderer')} label='Renderer' tabIndex={0}>
```
2Pacalypse-: I think we should leave this select in the settings for now, and just remove the `OpenGL` option from it. I know that having a select with only one option is a bit weird, but this whole settings dialog will need to be revamped pretty soon depending on what we decide to do with regards to SC:R support.

You can also remove any OpenGL specific code that we had in forge, eg.
https://github.com/ShieldBattery/ShieldBattery/blob/rust-dll/game/rust-dll/src/forge/renderer.rs#L26

Another thing that you should take care of is handle the case for users that have OpenGL saved in their settings. Guess it's already handled if you ignore that setting in forge and are just creating the DirectX renderer? We also need a migration to update those user's settings file, but I guess we can postpone that for when we revamp the settings.

neivv: I feel that leaving the select as something that just sits there saying 'This is not something you can change' is weird, but sure, I don't have strong preference either way.

And right, forge can be cleaned up as well - I was thinking if it's worth saving the details on what kind of hacks opengl needs to do in order to get acceptable rendering, but I also think it still didn't work properly all the time with Windows 10's taskbar?

2Pacalypse-: Yeah, I'm not sure either which is better; completely remove the renderer option for now, or just remove the OpenGL option from it. If we go with former option, then we should also publish the new version of settings and add a migration for it here:
https://github.com/ShieldBattery/ShieldBattery/blob/master/app/local-settings.js#L98

Btw, I can't really figure out how the settings are handled now in the Rust code, but the game code should handle the case of undefined/invalid setting values and apply the default values, where applicable, ie. what was being done here:
https://github.com/ShieldBattery/ShieldBattery/blob/master/game/node-bw/wrapped_brood_war.cpp#L467

And as far as the OpenGL hacks go, I can't really remember what was the state of things there; but I don't think we get any benefit of keeping those hacks around if we're not planning on implementing OpenGL renderer, especially if we move to SC:R (in any case, there's always git history that we can turn to if we really need to dig those in future :D).

neivv: The settings are technically handled here: https://github.com/ShieldBattery/ShieldBattery/blob/328b57c035bd4dff0ecd8a0ae1577a4a13e5d310/game/rust-dll/src/game_state.rs#L223
but since it turned out that all settings we currently have are only relevant to forge, it currently doesn't do anything else than forward to forge which reads what it needs https://github.com/ShieldBattery/ShieldBattery/blob/328b57c035bd4dff0ecd8a0ae1577a4a13e5d310/game/rust-dll/src/forge/mod.rs#L1123
(Invalid/undefined values are handled same as the c++ did code except it doesn't read the renderer field)

2Pacalypse-: Oh, I missed that; that's cool.

neivv:
This is now in a state where I believe (=hope) everything from the old version should be implemented and working =)

client/settings/settings.jsx
neivv: I removed the renderer/opengl from everywhere and added settings migration for new version; it would not be useful with SCR anyway (Or rather adding custom renderer for SCR would be a completely new, unrelated and huge task)

2Pacalypse-:
Great work! A few things:
1. Shouldn't `sign` be `%sign%` here? I'm not really familiar with `.bat` syntax though so I might be wrong :D - https://github.com/ShieldBattery/ShieldBattery/blob/rust-dll/game/build.bat#L52
2. You can probably remove these two lines (I think?):
https://github.com/ShieldBattery/ShieldBattery/blob/rust-dll/prod.yml#L7
https://github.com/ShieldBattery/ShieldBattery/blob/rust-dll/staging.yml#L7
3. Remove `game` from this comment:
https://github.com/ShieldBattery/ShieldBattery/blob/master/babel-register-hook.js#L3
4. Update this: https://github.com/ShieldBattery/ShieldBattery/blob/master/.gitattributes#L26
(This is not really that important, it just tells Github which folders/files to ignore when calculating language % on their page ^^)
5. Don't think we need git submodule anymore(?), so you can remove this:
https://github.com/ShieldBattery/ShieldBattery/blob/rust-dll/docs/GETTING_STARTED.md#repository-setup
6. The opening paragraph of `GETTING_STARTED` guide needs to include Rust as well ^^
7. I'm not sure, but I think some of our dependencies require a newer version of Node than is indicated here (you can just replace it with the current latest to be sure):
https://github.com/ShieldBattery/ShieldBattery/blob/rust-dll/docs/GETTING_STARTED.md#javascript
8. Main `README` needs to be updated as well. Mainly "Running the game section". Also, the `game` section in the "Project structure". If you feel like it, it would be cool to add a `README` to the `game` folder itself and describe a code structure + architecture of the Rust code there. Doesn't have to be anything too long, but would help some of us that are just starting with Rust :D
9. Not sure if [Travis CI](https://travis-ci.org/) supports Rust, but it would be nice if it could run some kind of a linter (using `rustfmt` rules?) on all open PRs and report an error if it doesn't pass. Travis config is in this file: https://github.com/ShieldBattery/ShieldBattery/blob/master/.travis.yml

game/build.bat
```diff
@@ -0,0 +1,60 @@
...
+)
+if not exist %scriptroot%\dist\bspatch.exe xcopy "%scriptroot%\..\tools\bspatch.exe" "%scriptroot%\dist" /y /f /c
+if not exist %scriptroot%\dist\ICSharpCode.SharpZipLib.dll xcopy "%scriptroot%\..\tools\ICSharpCode.SharpZipLib.dll" "%scriptroot%\dist" /y /f /c
+
+
+@rem build the DLL
+cargo build %cargoflags%
+if errorlevel 1 goto exit
+
+if %cargoflags%==--release (
```
2Pacalypse-: This is throwing a syntax error when I run this script. This works though:
```
if [%cargoflags%]==[--release] (
  ...
) else (
  ...
)
```

2Pacalypse-:
A couple of more things I've noticed when I tested it:
- The cursor is not centered when you enter the game
- For some reason it takes a while for StarCraft process to exit after the game (well, at least until the `app` receives the `/game/end` message), and even more weirdly, it seems to be correlated with how long the game went. For example, I opened a replay and left immediately and it took a couple of seconds to get the `/game/end` message and close the active game on the client. After that, I opened the same replay but watched it until the end (on x4-x8 though), and after the replay finished it took considerably longer to receive the `/game/end` message (somewhere around 10 seconds maybe).

2Pacalypse-:
One other small thing while we're at it :D
Could you perhaps add some blank line(s) in the log file before the first log line / after the last log line, so it's easier to distinguish different games in the log file.

game/src/forge/mod.rs
```diff
@@ -278,7 +278,10 @@ unsafe extern "system" fn wnd_proc(window: HWND, msg: u32, wparam: usize, lparam
                     });
                     if let Some(ref window) = forge.window {
                         // Move the cursor to the middle of the window
-                        SetCursorPos(window.width / 2, window.height / 2);
+                        SetCursorPos(
+                            window.client_x + window.width / 2,
```
2Pacalypse-: Not sure this does what it should be doing. In any case, it doesn't center the cursor for me. It only centers it *after* I exit the game, and only if I didn't move it until then.
Why not just hard code it to 320/240 like it was before? - https://github.com/ShieldBattery/ShieldBattery/blob/master/game/forge/forge.cpp#L594

neivv: Oh, right, you are correct.

It only works currently if the mouse sensitivity setting is set to leftmost value, which I had set.
Note that the C++ code calls the hook, which does the conversion from BW coords to screen coords, while this called the real function directly.

The reason don't want to call the hook is that Rust doesn't allow getting mutable access to a single variable simultaneously from two separate points, e.g.
```rust
fn parent() {
  // Access the global forge variable
  with_forge(|forge| {
    // do things...
    something(320, 240);
  });
}

fn something(x: i32, y: i32) {
  // Accesses the global forge variable, but parent function is already doing that :(
  with_forge(|forge| {
    // ...
  });
}
```
is not going to work.

Just calling that child function outside `with_forge` would obviously work, but trying to keep track of which helper functions can be called and which can not is likely going to end up being a nightmare. So I've kept any `with_forge` call directly at the hook functions themself, not calling any of those hooks myself, and any helper function needing forge state takes it as an `&mut Forge` parameter.

So I figured it's not worth it to create a helper function to share the logic between SetCursorPos hook and this one time we explicitly center screen at 320, 240 - the center of window, simple enough :) . ..But yeah, I didn't think it all the way through, there's enough logic needed that having a function `forge.set_cursor_to_bw_coords(320, 240)`, and having the hook call that as well seems to be the best way to do it.

neivv:
>     1. Shouldn't `sign` be `%sign%` here? I'm not really familiar with `.bat` syntax though so I might be wrong :D - https://github.com/ShieldBattery/ShieldBattery/blob/rust-dll/game/build.bat#L52
That works, I tested that part of the .bat =)

>     2. You can probably remove these two lines (I think?):
>        https://github.com/ShieldBattery/ShieldBattery/blob/rust-dll/prod.yml#L7
>        https://github.com/ShieldBattery/ShieldBattery/blob/rust-dll/staging.yml#L7
> 
>     3. Remove `game` from this comment:
>        https://github.com/ShieldBattery/ShieldBattery/blob/master/babel-register-hook.js#L3
> 
>     4. Update this: https://github.com/ShieldBattery/ShieldBattery/blob/master/.gitattributes#L26
>        (This is not really that important, it just tells Github which folders/files to ignore when calculating language % on their page ^^)
> 
>     5. Don't think we need git submodule anymore(?), so you can remove this:
>        https://github.com/ShieldBattery/ShieldBattery/blob/rust-dll/docs/GETTING_STARTED.md#repository-setup
> 
>     6. The opening paragraph of `GETTING_STARTED` guide needs to include Rust as well ^^
> 
>     7. I'm not sure, but I think some of our dependencies require a newer version of Node than is indicated here (you can just replace it with the current latest to be sure):
>        https://github.com/ShieldBattery/ShieldBattery/blob/rust-dll/docs/GETTING_STARTED.md#javascript
> 
>     8. Main `README` needs to be updated as well. Mainly "Running the game section". Also, the `game` section in the "Project structure". If you feel like it, it would be cool to add a `README` to the `game` folder itself and describe a code structure + architecture of the Rust code there. Doesn't have to be anything too long, but would help some of us that are just starting with Rust :D
Right, agree that those should be done

>     9. Not sure if [Travis CI](https://travis-ci.org/) supports Rust, but it would be nice if it could run some kind of a linter (using `rustfmt` rules?) on all open PRs and report an error if it doesn't pass. Travis config is in this file: https://github.com/ShieldBattery/ShieldBattery/blob/master/.travis.yml
Yeah, it should be doable. I still have to setup rustfmt and check if it butchers anything I wrote ^^

Travis also seems to have some [experimental support for windows](https://travis-ci.community/c/environments/windows), I don't know if Windows Server causes some issues (not being able to run tests due to binary being linked to d3d11.dll?), but running the few actual tests there are would also be nice. I think verifying that things compile may be doable from Linux side as well the very least.

client/active-game/game-server.js
```diff
@@ -41,15 +41,28 @@ class GameServer {
...
           this.onMessage(gameId, message)
         })
-        this.idToSocket = this.idToSocket.set(gameId, socket)
+        socket.on('error', e => {
+          log.error(`Game socket error ${e}`)
+        })
+        const pingInterval = setInterval(() => {
+          socket.ping()
+        }, 20000)
+        this.idToSocket = this.idToSocket.set(gameId, { socket, pingInterval })
```
2Pacalypse-: I'd prefer to use two different maps for this; or at least give a better name for this new, combined map. Which would hopefully eliminate the usage of `socket.socket` above which looks pretty weird and confusing ^^
Also, on line 44 you're shadowing the `socket` variable as well. Either using two different maps or using a differently named map should fix those issues.

game/README.md
```diff
@@ -0,0 +1,72 @@
...
+following files contain rest of the async side:
+
+- `app_messages.rs` Definitions for messages sent to/from app, and helper code for parsing them.
+- `app_socket.rs` Task for handling the websocket connection to app.
+- `cancel_token.rs` Helper code for cleaning up tasks when their parent gets dropped.
+- `game_state.rs` The most complex part of this DLL, manages game initialization. It may
+    handle other game-related async functionality in future, but right now we don't have any :)
+    Rest of the async code avoids directly calling BW functions at all
+    (as the BW thread is doing its own thing simultaneously, which could be bad),
+    but we do it in `GameState::init_game` as the older C++/JS code worked the same, and during
```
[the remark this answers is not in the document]
neivv: I'm not saying the current implementation is clearly wrong or unstable, but during the initialization  we have to let forge run Win32 message loop on the main thread to handle lobby networking. My understanding is that BW handling the lobby networking messages may touch some of the lobby state, but it ends up being fine anyway?

I definitely haven't seen anything in the older code which paused the message loop while the JS thread called BW functions to act on things, so guess they don't conflict anywhere?

game/src/forge/renderer.rs
```diff
@@ -0,0 +1,103 @@
...
+            None
+        }
+    }
+}
+
+impl Renderer {
+    pub fn new() -> Renderer {
+        // If we can't get display frequency, or it is below 60hz for some reason,
+        // still assume 60hz.
+        let display_frequency = device_display_frequency().unwrap_or(60).max(60);
```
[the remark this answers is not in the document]
neivv: It's `x.max(60)` as in `std::cmp::max(x, 60)` as in "greater of two values `x` and `60`"

(I don't think it min/max ever read well, regardless of what way they are written :l )

PR #401, opened 2019-04-07
Track the last focused state in the File Browser component.

2Pacalypse-:
This allows users to continue using the File Browser where they last
left off. The last focused state is saved in the local storage, so as
long as that is not cleared, it should be remembered. This PR also
implements a bunch of shortcuts, which should make using the File
Browser easier, eg. Up Arrow, Down Arrow, Page Up, Page Down, Home and
End.

Note: List virtualization has not been implemented yet, so the
performance in large folders might be questionable. This might come in
the later commit.


client/actions.js
```diff
@@ -88,6 +88,14 @@ export const CHAT_UPDATE_USER_OFFLINE = 'CHAT_UPDATE_USER_OFFLINE'
 export const DIALOG_OPEN = 'DIALOG_OPEN'
 export const DIALOG_CLOSE = 'DIALOG_CLOSE'
 
+// We have clicked on a folder inside the file browser
+export const FILE_BROWSER_CHANGE_PATH = 'FILE_BROWSER_CHANGE_PATH'
+// We are starting the process of getting the list of files (and folders) from a user's computer
+export const FILE_BROWSER_GET_BEGIN = 'FILE_BROWSER_GET_BEGIN'
+// We have received a response to our files retrieval (error or the list of the files for that
+// particular folder)
+export const FILE_BROWSER_GET = 'FILE_BROWSER_GET'
```
tec27: feel like it's a bit odd to just have an action named `GET` with this prefix, maybe `GET_FILE_LIST` or `GET_LIST` or something along those lines?

client/file-browser/browse-files.css
```diff
@@ -78,9 +82,11 @@
 .filesScrollableView {
 }
 
-.fileList {
-  padding-top: 8px;
-  padding-bottom: 8px;
+.fileList,
+.fileList > div {
```
tec27: Generally you probably want to avoid selectors like this if you can help it, better to give the things a unique class (I realize you might not be able to since this is controlled by the virtualized list?)

client/file-browser/browse-files.jsx
```diff
@@ -0,0 +1,526 @@
...
+      const initialPath = focusedPath ? pathApi.parse(focusedPath).dir : root
+
+      this.props.dispatch(changePath(browseId, initialPath))
+    } else {
+      this.props.dispatch(getFiles(browseId, path))
+    }
+
+    // Focus *something* when browser is opened, because if we don't, whatever was focused before
+    // the browser was opened will still have focus and will mess with our keyboard events.
+    this.focusTimeout = setTimeout(() => {
```
tec27: I'd probably prefer `Promise.resolve().then(() => { ... })` over a `setTimeout(0)`

(I'm guessing you probably took this from something I wrote, soooo :P)

IIRC we may have had some kind of actual timeout before focusing anything because a focus animation caused jank in panel opening? But maybe that doesn't matter here. If it doesn't matter, you could probably just focus an element when the ref is set

client/file-browser/browse-files.jsx
```diff
@@ -0,0 +1,526 @@
...
+
+  onBreadcrumbNavigate = path => {
+    const { root } = this.props
+    const pathWithoutRoot = path.slice(this.props.rootFolderName.length + 1)
+    this.props.dispatch(changePath(this.props.browseId, pathApi.join(root, pathWithoutRoot)))
+  }
+
+  onUpLevelClick = () => {
+    const { path } = this.props.fileBrowser[this.props.browseId]
+    const prevPath = path.lastIndexOf('\\') !== -1 ? path.slice(0, path.lastIndexOf('\\')) : ''
```
tec27: do we need to worry about different types of slashes here? would it be better to use the `path` module to do these kind of things?

client/file-browser/browse-files.css
2Pacalypse-: Yeah, this is a `div` inside `react-virtualized`which you can't attach class to, only inline style. And you can't select pseudo classes with inline style :/

client/file-browser/browse-files.jsx
2Pacalypse-: Yeah, using a callback ref instead of `createRef` API might be better here, as we can focus the element as soon the React calls the ref setter.

client/file-browser/browse-files.jsx
2Pacalypse-: Well, we only support Windows for now :P
But yeah, we use the `path` module to initially [set the path](https://github.com/ShieldBattery/ShieldBattery/blob/f699775c38c5a903bff61db484f3d2677043945c/client/replays/get-files.js#L11), so the path separator will be platform-specific. I'll update the handling of separators in this file to use the `path` module then, to be more future proof.

PR #403, opened 2019-04-24
Fix map uploading

neivv:
koa-body had breaking changes which weren't noticed when it was updated ( https://github.com/dlau/koa-body/blob/master/CHANGELOG.md#breaking-changes-in-v34 )

### DIFF
--- a/server/lib/api/maps.js
+++ b/server/lib/api/maps.js
@@ -26,8 +26,8 @@ function uploadPermissionCheck() {
 }
 
 async function upload(ctx, next) {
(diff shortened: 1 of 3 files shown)


2Pacalypse-:
Oops, that was my bad for not noticing ><
Could you also fix this in the `patches` API as well please? Thanks!
https://github.com/ShieldBattery/ShieldBattery/blob/master/server/lib/api/patches.js#L26-L27

neivv:
Right, fixed patches as well

PR #406, opened 2019-07-30
Upgrade the lobby creation workflow, both in functionality and in design.

2Pacalypse-:


2Pacalypse-:
Couple of things to note for this PR:
- Can't upload official maps atm. Admin panel that used to do it is broken for now, will add it in a later commit, or possibly even separate PR (need to introduce map ownership as a concept)
- Activity routing (or lack thereof) still feels a bit iffy to me. For now all the activity overlays are separate and the only way they can communicate is through a store. Activity reducer does keep the history of opened overlays in its store, so the "back" button can work. One thing that we could do is extend the "openOverlay" action to initialize the overlay with some data.
- Not sure if we should make activities a bit more like dialogs, by having their "title" area and "actions" area defined on an activity level instead of each individual overlay implementing its own. This would help a bit with some standardized actions, eg. "go back" action.

Other than those things, everything seems to work fine to me and I'm generally happy with the way things are implemented. Could always use some feedback though 😄

neivv:
Recent map list has a bit unexpected behaviour:
If I select maps A first, followed by B, they are ordered with B being first (leftmost) on the list and A second - That makes sense.
However, after hosting a game on map A, default selected map on create menu is still B and A doesn't move to the top of the list - I'd expect that a map hosted last would be moved to the first.
Selecting A again from map browse makes A the default focused map when opening the menu later, however, it doesn't move to start of the list like it would if it hadn't been on the last 5 selected maps.

I also had to upgrade postgres from 9.4, should update `docs/GETTING_STARTED.md` to whatever is the current minimum version

2Pacalypse-:
> However, after hosting a game on map A, default selected map on create menu is still B and A doesn't move to the top of the list - I'd expect that a map hosted last would be moved to the first.

Hmm, not sure how this can happen. Last selected map should always be remembered when the "Create lobby" overlay closes. So if you hosted a game on map A, when you open the overlay again, map A should be selected as a default, not map B.

But yeah, I get your point about "unexpected" behaviour. I debated a bit with myself on preserving the behaviour from Battle.net, and decided not to do it atm. So our recent maps list doesn't get reordered when a map from it is hosted. The reason why I did it like this for now is that I initially thought of pulling the recent maps list from the `games` table (which holds the list of all the games for each user - so we can pull last 5 games that a particular user hosted), but unfortunately we haven't yet implemented the full functionality of recording games :smile: - this is probably next on my TODO list, ref. branch: https://github.com/ShieldBattery/ShieldBattery/tree/game_results

But after thinking about it a bit more, I actually realized that the `games` table wouldn't even help me here, because I really wanted to preserve a map in the recent maps list after a user selects it, even it they don't host/play a game on it. This is another thing that is different from how Battle.net does it (I think?).

So yeah, I can change it so the hosted map always gets pushed to the top of the recent list; it's really no problem at all. I just thought it was not a big deal to have it like this because we have a visual indicator showing the user which map is selected.

> I also had to upgrade postgres from 9.4, should update `docs/GETTING_STARTED.md` to whatever is the current minimum version

I really don't understand which change in this PR requires later postgres version. Can you tell me which error(s) did you get with 9.4 version, so I could try to guess what would be the new minimum version :smile:

I guess it's also possible that the `node-postgres` requires the new version of postgres too?

neivv:
```
syntax error at or near \"ON\"","name":"error","stack":"error: syntax error at or near \"ON\"
    at Connection.parseE (C:\\Jotakin\\src\\manner-pylon\\server\
ode_modules\\pg\\lib\\connection.js:602:11)
    at Connection.parseMessage (C:\\Jotakin\\src\\manner-pylon\\server\
ode_modules\\pg\\lib\\connection.js:399:19)
    at Socket.<anonymous> (C:\\Jotakin\\src\\manner-pylon\\server\
ode_modules\\pg\\lib\\connection.js:121:22)
    at Socket.emit (events.js:203:13)
    at addChunk (_stream_readable.js:294:12)
    at readableAddChunk (_stream_readable.js:275:11)
    at Socket.Readable.push (_stream_readable.js:210:10)
    at TCP.onStreamRead (internal/stream_base_commons.js:166:17)","code":"42601"},"msg":"server error","time":"2019-08-06T15:28:36.032Z","v":0}
```
So likely from the INSERT INTO .... ON CONFLICT - obviously cannot promise it is the only syntax error, but at least everything works on postgres 11

--

I don't really care about recent map ordering otherwise, but it seems really unintuitive that even if you host a single map almost constantly, every time when hosting something else the favourite map is pushed closer to being to dropping off the list.

> Hmm, not sure how this can happen. Last selected map should always be remembered when the "Create lobby" overlay closes. So if you hosted a game on map A, when you open the overlay again, map A should be selected as a default, not map B.

Seems to be caused by `state.localMaps.selectedMap` always overriding the selection once it has been set.

client/lobbies/lobby-preferences-reducer.js
```diff
@@ -0,0 +1,71 @@
...
+  [LOBBY_PREFERENCES_GET_BEGIN](state, action) {
+    return state.set('isRequesting', true)
+  },
+
+  [LOBBY_PREFERENCES_GET](state, action) {
+    if (action.error) {
+      return state.set('isRequesting', false).set('lastError', action.payload)
+    }
+
+    const { selectedMap } = action.meta
```
neivv: (This thing here)

2Pacalypse-:
> So likely from the INSERT INTO .... ON CONFLICT - obviously cannot promise it is the only syntax error, but at least everything works on postgres 11

Oh yeah, that makes sense. We actually already used `ON CONFLICT` for matchmaking preferences too, but I guess it was missed back then ^^. Anyway, seems like version 9.5 is the one that added `ON CONFLICT`, so I guess that's our minimum version now: https://www.postgresql.org/docs/9.5/sql-insert.html



> I don't really care about recent map ordering otherwise, but it seems really unintuitive that even if you host a single map almost constantly, every time when hosting something else the favourite map is pushed closer to being to dropping off the list.

That's a really good point that I haven't considered previously. And I agree, having a map that you play pretty frequently on "sliding" out of the recent map list is a pretty bad UX, so I'll change the recent map list to always push the hosted map to the top of the list.



> Seems to be caused by `state.localMaps.selectedMap` always overriding the selection once it has been set.

God dammit, I could've guessed that's what's causing the bug. I hated writing that piece of code, but it was basically a workaround to the way our overlays work. Since we don't have a way to open an overlay with some init data (in this case the `LOCAL_MAPS_SELECT` action would open the `create-lobby` overlay with the map it just uploaded), I had to use store to relay information between those two overlays.

While this particular bug is easy to fix, I wonder if there's a better way to deal with overlay navigation and overlays opening other overlays. I did spend quite some time thinking about this, and even implementing some proof-of-concepts, but nothing that I was really happy with.

Issue #139, opened 2016-07-20
Chat messages should never include client-side timestamps


tec27:
This causes problems both in ordering/display (it can look like a message is out of order because the client's clock is off), and when we go to request more messages from the server (if the client's clock is ahead of the server, they can end up with duped/out-of-order messages). Anything we send from the server that we expect the client to display as a message should include a server timestamp.

For now, we're working around this issue by not adding messages with client-side generated timestamps to the chat unless it has had its history loaded at least a bit (and thus will guarantee its first message has a server-side timestamp)

PR #409, opened 2019-08-28
Add a message indicating a new day change in chat channels.

2Pacalypse-:
The message is added between any two subsequent messages that were not
posted on the same day. Also, the message is posted at the top of the
chat channel if there is no more history to load.

### DIFF
--- a/client/chat/chat-reducer.js
+++ b/client/chat/chat-reducer.js
@@ -24,6 +24,7 @@ import {
   JoinChannelMessage,
   LeaveChannelMessage,
   NewChannelOwnerMessage,
+  NewDayMessage,
   SelfJoinChannelMessage,
   UserOnlineMessage,
   UserOfflineMessage,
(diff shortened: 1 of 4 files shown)


2Pacalypse-:
Can't remember the exact details now, but this needs a bit more thought put into it. The most basic approach is to insert a new day message between any two messages where their date differs. However, special consideration needs to be taken around the "edges". When loading the chat history dynamically, a new day can happen between the two batches of history fetched. Also, we should probably display _something_ that jots down the date when the channel was first created.

There's a few edge cases that needs to be taken care of here, like filtering non-user messages when determining if the day changed, and making sure there are no surprises with client-side timestamps, which is closely related to this issue #139.

Anyway, last time I tried implementing this, I got a bit overwhelmed with trying to think of all the edge cases, and also testing the implementation of this proved to be really tricky; especially when you want to test insertion around "edges" when dynamically loading. That's why before attempting on implementing this again, we should probably see if there's anything we can do to reduce the amount of things we need to think about in our chat system, as well as figure out a way to write some kind of an automated tests for all the chat logic we do in the reducers and elsewhere.

tec27:
Feels like you were potentially very close but just holding it a bit wrong. I would try to think about the problem as if you were processing the entire history of messages at once, then figure out what is necessary to make message retrieval work as a pure subproblem of that (for example, maybe you need to load the previous oldest message into some state variable before processing a new retrieval).

Alternatively we could look at this purely as a display problem instead of putting things in the reducer state (I honestly think this is probably a better approach, but not totally sure).

2Pacalypse-:
Closed in favor of https://github.com/ShieldBattery/ShieldBattery/pull/669

PR #410, opened 2019-09-01
Build on node 12

neivv:
(Also make the script for building app's native module display msbuild output)


neivv:
I don't think anything here should make it unbuildable on the node/v8 versions that we previously supported, but I also don't have them to verify it.

PR #414, opened 2019-09-15
Add a Docker integration to streamline the deployment process.

2Pacalypse-:
There is a deployment guide in this PR with more information on how our deployment works.


2Pacalypse-:
I just want to note a couple more things that are needed to complete this PR, before I completely forget them.
#### Determine whether to run `nginx` inside a container or outside of it
The reason why I think it would be ok to run `nginx` outside the container is that you probably want to have `nginx` config files outside the container anyway (you don't really want to make them publicly available). And if you have `nginx` config files already set up on your server, it's not that much more effort to make sure you have `nginx` installed either (they usually go together, anyway).

In any case, I actually added a commented out config for running `nginx` inside the container, so if we want to run `nginx` as a service, just comment out this code:
https://github.com/ShieldBattery/ShieldBattery/pull/414/files#diff-4e5e90c6228fd48698d074241c2ba760R33-R44

#### Setup map system
As explained in our [getting started](https://github.com/ShieldBattery/ShieldBattery/blob/master/docs/GETTING_STARTED.md#set-up-map-system) guide, we require some files for server to be able to generate thumbnails for map images. This is pretty easy to do, just need to determine how to get those files into the server. Two possible options are:
- Have the files already available on the server and make them available to the Docker service by using volumes, similarly to how `uploaded_files` are handled: https://github.com/ShieldBattery/ShieldBattery/pull/414/files#diff-4e5e90c6228fd48698d074241c2ba760R13
- Have the files hosted somewhere that's reliable and publicly available and then bake them into the Docker image when building the image, similarly to how we download the `wait_for_it.sh` script: https://github.com/ShieldBattery/ShieldBattery/pull/414/files#diff-3254677a7917c6c01f55212f86c57fbfR30

#### Uploading the image
Currently I made our deploy script try to login to the Docker hub with the `shieldbattery` account (which I made) and then push the image there: https://github.com/ShieldBattery/ShieldBattery/pull/414/files#diff-b56215cd158613cc0deeddedf599193aR42-R56

However, we might want to push the image somewhere else (directly to our server?), but yeah, this is really not a big problem.

tec27:
Haven't tried running through this stuff yet, but a couple notes:

1) I think it probably is preferable to run nginx in a container. I also don't really see any issue with exposing an nginx config publicly, there's nothing really secret about it. May need to put some more thought into integrating certbot, I'm not sure that the provided config gets us all the way there.

2) We're not gonna be able to legally host the files for map rendering for direct consumption by anything (even though they are part of a game that's free, it has to be installed from Blizzard). I think we should probably work on the getting started guide for this piece, as I don't think a current install of SC:R will give you the necessary MPQ files at all (they've moved onto CASC). Ideally we could have a script that would output a correct directory structure given a current install, but barring that we should at least make sure the instructions still work for modern installs.

I can probably try and adjust for these things as I run through this, so no need to do any work on it atm. I'll let you know if I run into any issues I can't figure out

docs/DEPLOYMENT.md
```diff
@@ -0,0 +1,141 @@
...
+### Setup environment variables
+
+Docker allows for a couple of ways in which the `docker-compose.yml` file can be configured when running the containers, and all of them are centered around environment variables. In short, they are:
+- The `.env` file - a single file which can hold all of the environment variables you need; needs to be in the same directory as the `docker-compose.yml` file.
+- Shell environment variables - you can set the environemnt variables directly in shell before running the docker compose (they take priority over the `.env` file).
+
+You can read more about the environment variables in docker compose at [this link](https://docs.docker.com/compose/environment-variables/).
+
+The following environment variables are available to be configured (those without a default value are required):
+- `SERVER_NAME` - Must be either `production` or `staging` which is used to fetch the image from docker hub.
```
tec27, line 69: Maybe I'm missing something, but I don't really understand the purpose of this. Generally the staging server should be running the same stuff as the production one, it's just a later version (before that version gets promoted to production). If we have to change things other than configuration between staging and production then staging isn't a particularly good test. Is there something I'm overlooking there?

tec27: I assume from doing more reading that this is purely for the electron app side of things. I wonder if it might be better to upload versions the same package for staging/production server URLs. From a quick look at Docker Hub it seems like these versions can include strings and the like: https://hub.docker.com/_/sonarqube/?tab=tags

So we could upload like a 1.0.0-production and 1.0.0-staging at the same time. This still feels quite odd to me though given that both images would now be including a URL to our actual servers, rather than being deployable to any random location. If we instead mounted the installer location (or made a separate service for delivering the installers) perhaps all that weirdness goes away. That'd also remove all the stuff preventing us from easily building these docker images remotely.

2Pacalypse-: I can't remember my exact motivation now, and this is something I did very early in my process of learning Docker, but I _think_ my idea of this environment variable was to use it as just a decorator in the image name; I don't think it's used in any more meaningful capacity than that. For some reason, I thought back then that we needed different images for different flavors of the server.

But thinking about it again now, I see what you mean that this doesn't make much sense. The (only?) difference between our production and staging servers is located in the config file, which is loaded externally into the image, so I think you're right, we could probably deploy only one image.

I think all notions of deploying a particular server can be removed, except one: running `dist` script before deploying the production server, and running `diststaging` before deploying the staging server. So we probably still need two npm scripts for deploying, but we don't need to send the `server` argument anymore, just the version here:
https://github.com/ShieldBattery/ShieldBattery/pull/414/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R20-R21

2Pacalypse-: Took me a while, but I actually just understood what you mean by mounting the installer location. And yeah, I guess the only thing holding us back from having server-agnostic deployment is the way we build electron stuff with `electron-builder`. If we can somehow consolidate `prod.yml` and `staging.yml` into one build file, then deployment, along with building, won't depend on the actual server.

server/lib/db/init-scripts/install-extensions.sql
```diff
@@ -0,0 +1,4 @@
+-- Sql script which installs necessary extensions for the shieldbattery server
```
2Pacalypse-, line 1: There's something I wanted to note about this file (which I somehow alluded to in the migration guide). Which is that it feels weird to even have it as it's not that useful. The way our docker service is running now is that it will run this script _only_ when the database is empty, which usually happens only when running the service the first time. And that means that if we want to install another extension once the database is already set up, it can't be put into this file, as it won't run again.

I feel like it might be more useful to have dedicated migrations for extension installing, as migrations will be run every time a service starts. I don't know if there are any downsides to installing extensions through migrations though. And also, even if add a migration to install these two extensions now, they would still need to be installed manually when setting up a server for the first time since our previous migrations depend on them.

Dockerfile
```diff
@@ -0,0 +1,60 @@
...
+WORKDIR /shieldbattery
+
+# Copy the files necessary to install dependencies
+COPY server/package.json server/
+COPY server/yarn.lock server/
+COPY package.json ./
+COPY yarn.lock ./
+
+# Install only the server and main dependencies (`app` dependencies should be built into the Electron app)
+# TODO(2Pac): Figure out why some `devDependencies` are needed for the server to run in production
```
tec27, line 24: Seems to me like these are just incorrectly tagged as dev dependencies, they're actually used in production as well (because we don't precompile with babel, although I guess we could? Might make debugging server errors a bit harder though).

If we were willing to move to `.mjs` files and required that the production server ran webpack separately (e.g. it expected that the compiled client JS was already there rather than compiling it on startup) we could probably do away with babel as a production dep for the server. Not sure how big of a gain it really is though, the `@babel` folder in my node_modules is like 13MB. (It *is* potentially nice from a runtime speed and reproducible build perspective, just not sure it's worth the effort atm)

Dockerfile
```diff
@@ -0,0 +1,68 @@
...
+# really matter as it will not be present in the final image, but keeping the path short makes
+# copying from it easier in the next stage.
+WORKDIR /shieldbattery
+
+# Copy the whole repository to the image, *except* the stuff marked in the `.dockerignore` file
+COPY . .
+
+# Install only the server and main dependencies (`app` dependencies should be built into the
+# Electron app) and install only production dependencies to make the image smaller.
+RUN cd server && yarn && cd .. && yarn
```
2Pacalypse-, line 26: Shouldn't the `ENV NODE_ENV=production` be declared above this line if we only want to install the production dependencies?

EDIT: I see we prune the server deps to only the production ones below. Maybe a comment outlining why that's necessary would help here.

deploy.bat
```diff
@@ -0,0 +1,107 @@
+@rem This script should be called with both the current version number of the
```
2Pacalypse-, line 1: Typo, extra word `both`.

deploy.bat
```diff
@@ -0,0 +1,107 @@
...
+)
+
+SET NAME=shieldbattery/shieldbattery
+SET IMG=%NAME%:%gitsha%
+SET NPM_TAG=%NAME%:%version%
+SET LATEST_TAG=%NAME%:latest
+
+@rem The docker file will take whatever files are currently built and create a docker image out of
+@rem them. This means that this script doesn't care which branch you're trying to deploy. It only
+@rem cares whether you're trying to deploy to the main production server or the staging server.
```
2Pacalypse-, line 58: This sentence can also be removed I believe.

docker-compose.yml
```diff
@@ -0,0 +1,88 @@
...
+    image: shieldbattery/${SERVER_NAME}:${SERVER_VERSION:-latest}
+    container_name: sb-${SERVER_NAME}-server
+    # Required for proper handling of kernel signals (eg. ctrl-c); for more info, see:
+    # https://github.com/nodejs/docker-node/blob/master/docs/BestPractices.md#handling-kernel-signals
+    init: true
+    ports:
+     - "80:5555"
+    volumes:
+      - uploaded_files:${UPLOADED_FILES_PATH:-./server/uploaded_files}
+      - type: bind
```
notenbrood, line 14: This syntax is pretty much never used for bind volumes, and even the short version has a way to specify read only (although the bash evaluation makes it slightly less readable):

`- ${SERVER_CONFIG_PATH:-./server/config.js}:/home/node/shieldbattery/server/config.js:ro
- ${SERVER_DB_CONFIG_PATH:-./server/database.json}:/home/node/shieldbattery/server/config.js:ro`

docker-compose.yml
```diff
@@ -0,0 +1,88 @@
+version: "3.7"
+
+services:
+  server:
+    image: shieldbattery/${SERVER_NAME}:${SERVER_VERSION:-latest}
+    container_name: sb-${SERVER_NAME}-server
```
notenbrood, line 6: sb- prefix is probably redundant because starting these containers will use a Docker-Compose project name, even if you dont set one. See: https://docs.docker.com/compose/reference/overview/#use--p-to-specify-a-project-name

Dockerfile
tec27: Yeah, we need the non-production deps so that we can use webpack, then we prune it back out after doing so. I'll add a comment

deploy.bat
tec27: I think it's still relevant, it is indeed branch-agnostic

2Pacalypse-: I meant the last sentence of this comment, i.e.
```
It only cares whether you're trying to deploy to the main production server or the staging server.
```

I don't think this script cares about that anymore :d

tec27: Ah, I was thinking about a different comment elsewhere anyway, I don't think this one is particularly useful so I'll just remove it.

docker-compose.yml
tec27: I think this was intended to not be a bind volume, I just misunderstood when I was initially using this. It is/was at least declared in `volumes` and is intended to be filled by the server when people upload things, so I think this usage is probably fine.

docker-compose.yml
tec27: I removed most of the `container_name` usage in this file, as it seemed fairly redundant with the service names anyway. I only left it for `nginx_proxy` because other containers need it to have a consistent ID.

Unsure if that is a typical usage of `container_name` or not, but it seems reasonable to me atm

server/lib/db/init-scripts/install-extensions.sql
tec27: Can't use migrations for installing extensions because it needs superuser privileges. I tried many approaches, none of them really worked, so I just put it in the init scripts. If more extensions are ever need it, they can just be run one-off using `docker exec` prior to deploying that version (example in the init script itself). New extensions should be added to the init script so that brand new deploys get them by default.

docker-compose.yml
```diff
@@ -0,0 +1,169 @@
...
+    restart: always
+    environment:
+      - POSTGRES_USER=postgres
+      - POSTGRES_PASSWORD=${POSTGRES_SUPER_PASSWORD}
+      - POSTGRES_DB=postgres
+      - SB_DB=${SB_DB:-shieldbattery}
+      - SB_DB_USER=${SB_DB_USER:-shieldbattery}
+      - SB_DB_PASSWORD=${SB_DB_PASSWORD}
+    volumes:
+      - ./server/deployment/db/init-scripts:/docker-entrypoint-initdb.d:ro
```
tec27, line 130: This would ideally be provided from `app_server`'s filesystem, might be a weird order of operations here though, given the dependencies? Need to think about it more

PR #423, opened 2019-12-04
Add server-side support for updating maps.

2Pacalypse-:
For now, only `name`, `description` and `visibility` can be updated.
In future, users will also be able to update the map file itself.

### DIFF
--- a/server/lib/models/maps.js
+++ b/server/lib/models/maps.js
@@ -248,7 +248,7 @@ export async function getMaps(
   }
 }
 
-export async function updateMap(mapId, name, description, visibility) {
+export async function updateMap(mapId, favoritedBy, name, description, visibility) {
   let setStatement = 'SET'
   const params = []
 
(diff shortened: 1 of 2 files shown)


server/lib/models/maps.js
```diff
@@ -248,6 +248,43 @@ export async function getMaps(
...
+    ${setStatement}
+    WHERE id = $${params.length + 1}
+    RETURNING *;
+  `
+  params.push(mapId)
+
+  const { client, done } = await db()
+  try {
+    const result = await client.query(query, params)
+    return createMapInfo(result.rows[0])
```
neivv: Isn't this going to fail since `result` here doesn't have the columns from the maps table? Maybe just not returning anything would be fine?

2Pacalypse-: Yeah, you're right.
I think we still want to return the updated `MapInfo` model, so the client can update its internal state as well. Will fix it.

PR #424, opened 2019-12-06
Add server-side support for deleting maps.

2Pacalypse-:


server/lib/models/maps.js
```diff
@@ -301,6 +302,40 @@ export async function updateMap(mapId, favoritedBy, name, description, visibilit
...
+
+    // Delete all maps that no one is referencing anymore, if any
+    query = `
+      DELETE FROM maps AS m
+      WHERE m.hash NOT IN (
+        SELECT map_hash
+        FROM uploaded_maps
+      )
+      RETURNING *;
+    `
```
neivv: Would having the first deletion query return map hash and limiting this query to only operate on map with that hash be better?

(To be clear, I have no idea how this currently performs relatively or absolutely, or if we have any reason to believe that this query may need to delete more than one map ever)

server/lib/file-upload/local-filesystem.js
```diff
@@ -50,15 +56,22 @@ export default class LocalFsStore {
...
   }
 
-  async url(filename) {
-    const normalized = path.posix.normalize(filename)
-    if (path.isAbsolute(normalized) || normalized[0] === '.') {
-      throw new Error('Invalid directory')
+  async delete(filename) {
+    const full = this._getFullPath(filename)
+    try {
+      await access(full)
```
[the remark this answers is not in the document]
2Pacalypse-: Yeah, you're right on both counts. Will add a log message.

server/lib/models/maps.js
2Pacalypse-: Hmm, I'm having trouble imagining how that would work?
We need the check if the hash of the deleted map exists at all in the `uploaded_maps` table before deleting an entry in the `maps` table.

Granted, this doesn't look like the most optimal way to do it, but I'm really failing to see how else it could be done at the moment.

2Pacalypse-:
> Seems good as far as I can tell.
> 
> That being said, if we an user deletes their uploaded map like this, anyone that has favorited the map ends up just noticing that a map had disappeared from their favorites without any explanation. Is that ideal?
> 
> Obviously we want to let people delete the maps they uploaded, but if a map uploaded by someone gets popular, then that gives a single user quite a lot of power to be an ass by deleting the map and requiring people to find another copy of it.

Yeah, I did think about this a bit and decided to leave it as a future improvement for now. I was thinking of maybe sending a message in some way to the users if the map they've favorited has been deleted. 

Also, about users deleting popular maps; this is actually a pretty serious problem in content creation platforms. TeamLiquid had the same problem where people would delete their threads which were really popular, and then the "solution" was implemented that editing/deleting of threads becomes locked after a few months of inactivity and an admin has to be messaged if you want to do it. Of course, we would need to have some kind of terms and conditions document which says we can actually do that :d

server/lib/models/maps.js
2Pacalypse-: Oh nvm, I see what you mean now (I think :d).

2Pacalypse-: Nvm again, I don't think I know what you mean; or at least not enough to make it work ^^

neivv: Yeah, I don't have too clear concept of it to give any concrete idea how to do it. Current way is fine to me

PR #433, opened 2020-01-29
Organize game code

neivv:
No functionality changes, just modifying how we interact with BW

### DIFF
--- a/game/src/bw_1161.rs
+++ b/game/src/bw_1161.rs
@@ -2,6 +2,7 @@
 
 mod observing;
 mod snp;
+mod storm;
 
 use std::io;
 use std::ffi::{CStr, OsStr};
@@ -75,7 +76,6 @@ impl bw::Bw for Bw1161 {
         game_info: &mut bw::JoinableGameInfo,
         map_path: &[u8],
     ) -> Result<(), u32> {
-        use crate::storm;
         let ok = join_game(game_info);
         if ok == 0 {
             return Err(storm::SErrGetLastError());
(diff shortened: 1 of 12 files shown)


game/src/bw_1161.rs
```diff
@@ -0,0 +1,612 @@
...
+    }
+}
+
+unsafe fn create_lobby(
+    map_path: &Path,
+    name: &str,
+    game_type: bw::GameType,
+) -> Result<(), bw::LobbyCreateError> {
+    let map = find_map_entry(map_path)?;
+    // Password must be null for replays to work
```
2Pacalypse-: Is this correct? I thought when watching replays through a lobby, password must *not* be null?
Or is this related to something else?

2Pacalypse-:
Btw, does this bit in description still holds?
>The code of the game DLL has two major parts: asynchronous code that focuses on handling network traffic, and 'synchronous' code that executes inside of hooks into BW's code. They currently are mixed in src/ directory, but it may be worth separating one/both of them in separate modules, even though most of the files aren't 100% async or 100% BW.

neivv:
It kind of still holds? 1.16.1-specific code is now under bw_1161/, and that's obviously almost all of bw-side code, but there's still bunch of files whose code is run on bw's thread. I'll revise the documentation a bit once scr code is added =)

game/src/bw_1161.rs
neivv: You cannot set password anyway in LAN, but for some reason hosting replay will specifically check that there isn't password.. I think? Anyway it's been like that always here

2Pacalypse-:
Alright, sounds good. As far as I'm concerned you can merge this into the master. "Squash and merge" option on github should probably be used here.

PR #448, opened 2020-04-13
Create a minidump file every time the inject fails.

2Pacalypse-:
This commit also fixes the script that builds our native deps to build
them for the current version of Electron, so there's no mismatch between
node.js versions in Electron and the host environment.

### DIFF
--- a/app/native/process/wrapped_process.cpp
+++ b/app/native/process/wrapped_process.cpp
@@ -122,22 +122,6 @@ ScopedVirtualAlloc::~ScopedVirtualAlloc() {
   }
 }
 
-static WindowsError CreateMiniDump(HANDLE process, const string& path) {
-  HANDLE handle = CreateFileA(path.c_str(), GENERIC_READ | GENERIC_WRITE, 0, NULL, CREATE_ALWAYS,
-    FILE_ATTRIBUTE_NORMAL, NULL);
-  if (handle == INVALID_HANDLE_VALUE) {
-    return WindowsError("CreateMiniDump -> CreateFile", GetLastError());
-  }
-  WinHandle file(handle);
-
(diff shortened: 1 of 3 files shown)


app/native/process/wrapped_process.cpp
```diff
@@ -1159,6 +1135,30 @@ WindowsError Process::GetExitCode(uint32_t* exit_code) {
   return WindowsError();
 }
 
+void Process::CreateMiniDump(const string& error_dump_path) {
+  HANDLE handle = CreateFileA(error_dump_path.c_str(), GENERIC_READ | GENERIC_WRITE, 0, NULL,
+    CREATE_ALWAYS, FILE_ATTRIBUTE_NORMAL, NULL);
+  WindowsError err = WindowsError();
+
+  if (handle == INVALID_HANDLE_VALUE) {
+    err = WindowsError("CreateMiniDump -> CreateFile", GetLastError());
```
neivv, line 1144: This should be reorganized so that the code won't call MiniDumpWriteDump if an error occurs here

neivv: Well, now we're losing the error without logging it.
It probably won't ever be an important error path, so if you have had enough of revising this code then feel free to just merge it =) But this possibly should have an inner function which returns the error that is then checked. (Or a try-catch, but maybe there was a reason why this code doesn't use exceptions)

2Pacalypse-: Good point.

I fixed it now, although in a bit uglier way than you suggested. I think it's fine for now as it's unlikely this will have to change much in future, if at all?

neivv: Yup that's perfectly fine now

PR #481, opened 2020-06-05
Refactor the game loading code.

2Pacalypse-:
This PR takes the code that was inside a lobby service and pulls it into
its own thing. Now we have two new pieces to deal with the game loading:
- Game loader
- games API

Game loader is a service which is (currently) in charge of keeping
track of games that are loading, as well as setting up the routes for
all players. In the (near) future, it will also be used to start
tracking results for each loaded game. Another thing that might be
useful to have is to track the currently active game for all users, even
for single-player games (like watching a replay, or demo-ing a map).
This would allow us to display the information next to the user's name
on the client-side if they're currently in the game, and what exactly is
the context of their game.

Games API is currently serving a purpose to allow clients to report when
their game has successfully loaded, or when it has failed loading. In
(near) future, it will be used to report game results as well. Also,
even though the API is currently called from the client, it might make
more sense to call the API directly from the game in future.


client/active-game/socket-handlers.js
```diff
@@ -14,12 +15,10 @@ export default function ({ siteSocket }) {
...
     })
 
-    // TODO(tec27): Only invoke the lobby API for lobby loads, need to split this out for e.g.
-    // replays and AMM and such.
     if (status.state === 'playing') {
-      siteSocket.invoke('/lobbies/gameLoaded')
+      fetch('/api/1/games/' + encodeURIComponent(status.id), { method: 'put' })
     } else if (status.state === 'error') {
-      siteSocket.invoke('/lobbies/loadFailed')
+      fetch('/api/1/games/' + encodeURIComponent(status.id), { method: 'delete' })
```
tec27, line 21: This feels weird semantically. With the way this works, it's like we're creating an entry for this ID if the game loads. But here, we're deleting an entity that was basically not created?

I think it'd feel less weird if the assumption was the server created this entity, and the `/gameLoaded` replacement here actually `PUT` some new data into that entry (e.g. `loaded: true` or something)

server/lib/games/game-loader.js
```diff
@@ -0,0 +1,185 @@
...
+    })
+
+    return gameLoad
+  }
+
+  // A function which is called to register the (now loaded) game into the system for each player.
+  // Currently, it only cleans up the loading game state, but in future it will be used to accept
+  // the game results and perhaps registering the user to the gameplay activity registry, if not
+  // already registered.
+  registerGame(gameId, playerName) {
```
tec27, line 98: this naming feels a bit odd to me given that the information passed in is something that this service already "knows", it's just registering the fact that the loading completed. Maybe `registerGameAsLoaded` or something along those lines?

PR #482, opened 2020-06-17
Make the matchmaking actually launch the game when all players accept

2Pacalypse-:
it.

### DIFF
--- a/client/active-game/active-game-manager.js
+++ b/client/active-game/active-game-manager.js
@@ -2,7 +2,6 @@ import { remote } from 'electron'
 import path from 'path'
 import { promises as fsPromises } from 'fs'
 import { EventEmitter } from 'events'
-import deepEqual from 'deep-equal'
 import { Set } from 'immutable'
 import { checkStarcraftPath } from '../starcraft/check-starcraft-path'
 import log from '../logging/logger'
@@ -46,30 +45,24 @@ export default class ActiveGameManager extends EventEmitter {
 
   setGameConfig(config) {
     const current = this.activeGame
-    if (current) {
-      if (deepEqual(config, current.config)) {
-        // Same config as before, no operation necessary
-        return current.id
-      }
-      if (current.id !== config.gameId) {
-        // Quit the currently active game so we can replace it
-        this.emit('gameCommand', current.id, 'quit')
-      }
+    if (current && current.id !== config.gameId) {
+      // Means that a previous game left hanging somehow; quit it
+      this.emit('gameCommand', current.id, 'quit')
     }
     if (!config.setup) {
       this._setStatus(GAME_STATUS_UNKNOWN)
       this.activeGame = null
       return null
     }
-
     if (current && current.routes && config.setup) {
       const routesIds = new Set(current.routes.map(r => r.for))
       const slotIds = new Set(config.setup.slots.map(s => s.id))
 
       if (!slotIds.isSuperset(routesIds)) {
-        this._setStatus(GAME_STATUS_UNKNOWN)
+        this._setStatus(GAME_STATUS_ERROR)
         this.activeGame = null
-        return null
+
+        throw new Error("Slots and routes don't match")
       }
     }
 
@@ -101,12 +94,15 @@ export default class ActiveGameManager extends EventEmitter {
       return
     }
 
-    if (current && current.config && routes) {
+    if (current && current.config) {
       const routesIds = new Set(routes.map(r => r.for))
       const slotIds = new Set(current.config.setup.slots.map(s => s.id))
 
       if (!slotIds.isSuperset(routesIds)) {
-        return
+        this._setStatus(GAME_STATUS_ERROR)
+        this.activeGame = null
+
+        throw new Error("Slots and routes don't match")
       }
     }
 
--- a/client/active-game/active-game-manager.js
+++ b/client/active-game/active-game-manager.js
@@ -3,6 +3,7 @@ import path from 'path'
 import { promises as fsPromises } from 'fs'
 import { EventEmitter } from 'events'
 import deepEqual from 'deep-equal'
+import { Set } from 'immutable'
 import { checkStarcraftPath } from '../starcraft/check-starcraft-path'
 import log from '../logging/logger'
 import { REMASTERED } from '../../common/flags'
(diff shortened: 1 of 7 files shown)


client/active-game/active-game-manager.js
```diff
@@ -50,15 +51,28 @@ export default class ActiveGameManager extends EventEmitter {
         // Same config as before, no operation necessary
         return current.id
       }
-      // Quit the currently active game so we can replace it
-      this.emit('gameCommand', current.id, 'quit')
+      if (current.id !== config.gameId) {
```
tec27, line 54: I wonder if this should come before the `deepEqual` check (and that should be an `else if`)? Unsure of what the semantics are here as far as whether a different config would always mean a different ID or whatever. I also don't remember where these IDs get generated now, but if it's a server thing, maybe the config check is completely unnecessary (because the server would handle config changes by making a new ID).

client/active-game/active-game-manager.js
```diff
@@ -50,15 +51,28 @@ export default class ActiveGameManager extends EventEmitter {
...
       this._setStatus(GAME_STATUS_UNKNOWN)
       this.activeGame = null
       return null
     }
 
+    if (current && current.routes && config.setup) {
+      const routesIds = new Set(current.routes.map(r => r.for))
+      const slotIds = new Set(config.setup.slots.map(s => s.id))
+
+      if (!slotIds.isSuperset(routesIds)) {
```
tec27, line 69: Feels like we should probably be logging something in this case, or this will require some work to track down "why the game status suddenly became `UNKNOWN`" when this error is encountered. I also think throwing an exception would potentially be fine, given that this is a programmer error

client/active-game/active-game-manager.js
```diff
@@ -82,13 +96,32 @@ export default class ActiveGameManager extends EventEmitter {
...
-    if (!this.activeGame || this.activeGame.id !== gameId) {
+    const current = this.activeGame
+    if (current && current.id !== gameId) {
       return
     }
 
-    this.activeGame.routes = routes
-    this.emit('gameCommand', gameId, 'routes', routes)
-    this.emit('gameCommand', gameId, 'setupGame', this.activeGame.config.setup)
+    if (current && current.config && routes) {
```
tec27, line 104: Is there a place where `routes` would be undefined/null here? Feels kinda weird and maybe like we should fix that to pass an empty array/list if so, instead of doing this check

client/matchmaking/socket-handlers.js
```diff
@@ -90,15 +96,63 @@ const eventToAction = {
...
+    } = getState()
+
+    const {
+      hash,
+      mapData: { format },
+      mapUrl,
+    } = event.mapInfo
+    // NOTE(2Pac): We can't download map any sooner since we don't know which map we'll play until
+    // all players accept the game
+    mapStore.downloadMap(hash, format, mapUrl)
```
tec27, line 121: might wanna file a bug for this, I *believe* my plan was to just download all the matchmaking maps before they could enter the queue (although I guess we still might need to double-check here if they deleted it in the meantime)

server/lib/models/maps.js
```diff
@@ -143,6 +143,16 @@ export async function mapExists(hash) {
 }
 
 export async function getMapInfo(mapIds, userId) {
+  const favoritedJoin = userId
```
tec27, line 146: convert this to use the sql template tag thing maybe?

should make the variable number of parameters easier to use, I think :)

client/active-game/active-game-manager.js
2Pacalypse-: I assume this used to handle the case where a game process would leave hanging for whatever reason, to give it a chance to exit the game one last time? We cleanup the active game (by setting it to `null`) both when a game exits cleanly and when there's a launch error, so not sure if it's worth having this check anymore, but I don't think it's hurting anything.

As for the `deepEqual` check, I don't think it's necessary anymore. The server is the one that generates the `gameId` now, and it will do so every time some service calls the `loadGame` function:
https://github.com/ShieldBattery/ShieldBattery/blob/master/server/lib/games/game-loader.js#L83

The only time when the `deepEqual` check would return `true` is if the `setGameConfig` function is called twice during the loading of a same game? I don't see at the moment how that would be possible.

client/active-game/active-game-manager.js
2Pacalypse-: No, `routes` should always be set to something here. This was just a case of defensive programming that's not needed; I'll remove it.

client/active-game/active-game-manager.js
tec27: Well before, the ID was generated on the client (and was a way of sending messages to specific clients). You can kind of think of it as a hash of the game config before though (as if the game config was the same, we'd simply keep using the ID we had). So yeah, I think now that that's gone, no `deepEqual` check is needed, you can simply check the ID for the same purpose

2Pacalypse-: One thing that crossed my mind by having this check if the `gameId` is different (and exiting the "old" game if it is), is that the users might have some unintended consequences. Since we currently don't have any checks that stops a user from starting a game while they're already in another game (and not sure that we should?), it's possible for a user to, for example, click on a replay while they're in another game and their game will abruptly exit and be replaced with a replay.

So I would do either of the following two things:
- remove this check that sends a `quit` command to the old game if a user is already in a game
- don't unsubscribe users from gameplay activity registry until _after_ the game finishes, not just until they leave the lobby/queue

I'm personally leaning towards the former, but I'm not sure if there are any unintended consequences of allowing users to be in multiple games at once.

tec27: I'd prefer something like the latter, although I'm not actually sure if what you mentioned is the right approach. The code we have here has always assumed we only had one active game, and any additional active games were just things we needed to shut down. I don't see why we'd want to let players be in two multiplayer games at once, and watching a replay while playing a ranked game doesn't seem like something we'd want to support either (not to mention that with SC:R, the resource usage of doing that is quite high)

2Pacalypse-: Yeah, thinking about it a bit further, I think I agree with the latter option as well now.

I'll leave the check which exits the old game then for now, and eventually we'll probably need to extend the game loader to deal with registering/unregistering the gameplay activities, as well as handling other malfunctions and stuff as you mentioned.

server/lib/wsapi/matchmaking.js
```diff
@@ -109,8 +110,12 @@ export class MatchmakingApi {
 
   async _onGameSetup(matchInfo, clients, players, setup = {}) {
     // TODO(2Pac): Select map intelligently based on user's preference
-    const mapInfo = (await getMapInfo(['4cd22c4f-2924-42f3-91ae-0e85ddaace3d']))[0]
+    const mapPool = await getCurrentMapPool(matchInfo.type)
+    if (!mapPool) {
+      throw new errors.NotFound('invalid map pool')
```
tec27, line 115: Feels like a 5xx-ish error to me, given that it shouldn't generally be possible to get to this point if a map pool isn't around?

server/lib/wsapi/matchmaking.js
```diff
@@ -109,8 +110,12 @@ export class MatchmakingApi {
...
+    }
 
+    const mapInfo = (await getMapInfo(mapPool.maps))[0]
```
tec27, line 118: The error below this (github won't let me add a comment there :( ) should probably also be 5xx, it implies some bug in the map selection process

PR #487, opened 2020-07-02
Start loading the game in lobbies as soon as the countdown starts.

2Pacalypse-:
This should reduce the loading time of SC:R games considerably. Or 5
seconds to be more exact.

Once the countdown finishes, the loading screen is shown for a minimum
of 2 seconds before launching the users into the game.

### DIFF
--- a/client/active-game/active-game-manager.js
+++ b/client/active-game/active-game-manager.js
@@ -25,6 +25,7 @@ export default class ActiveGameManager extends EventEmitter {
     this.mapStore = mapStore
     this.activeGame = null
     this.serverPort = 0
+    this.allowStartSent = false
   }
 
   getStatus() {
@@ -129,6 +130,7 @@ export default class ActiveGameManager extends EventEmitter {
     this._setStatus(GAME_STATUS_STARTING)
     setTimeout(() => {
       this.emit('gameCommand', this.activeGame.id, 'allowStart')
+      this.allowStartSent = true
     }, waitTime)
   }
 
@@ -153,6 +155,12 @@ export default class ActiveGameManager extends EventEmitter {
     if (game.routes) {
       this.emit('gameCommand', id, 'routes', game.routes)
       this.emit('gameCommand', id, 'setupGame', game.config.setup)
+    }
+
+    // If the `allowStart` command was already sent by this point, it means it was sent while the
+    // game wasn't even connected; we resend it here, otherwise the game wouldn't start at all.
+    if (this.allowStartSent) {
+      this.emit('gameCommand', this.activeGame.id, 'allowStart')
     }
   }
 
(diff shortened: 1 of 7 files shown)


2Pacalypse-:
Currently, there are following defects:
- active-game in the navbar is shown as soon as the countdown starts (will have to refactor how the active-game navbar is shown anyway to account for matchmaking etc.)
- loading can fail while the countdown is still going on, in which case both the countdown and loading are canceled (this might be a bit unexpected, but potentially this might be fine?)
- when a user leaves the lobby during a countdown, the loading doesn't get cancelled for the leaving player, while it does for all the other players in the lobby (this is probably the trickiest thing to fix, since the leaving player gets unsubscribed from the lobby events, they don't receive the cancelLoading message; still not sure how to fix this)

client/active-game/active-game-manager.js
```diff
@@ -117,10 +119,21 @@ export default class ActiveGameManager extends EventEmitter {
...
 
+  allowStart(gameId, waitTime = 2000) {
+    if (this.activeGame && this.activeGame.id !== gameId) {
+      return
+    }
+
+    this._setStatus(GAME_STATUS_STARTING)
+    setTimeout(() => {
+      this.emit('gameCommand', gameId, 'allowStart')
+      this.allowStartSent = true
```
tec27, line 133: I never see this get set back to false anywhere, so it seems like once this happens once, this condition is no longer correct? I think it'd make more sense to put this condition inside the `this.activeGame`, that way we're guaranteed it gets erased if we're launching a new game

client/active-game/active-game-manager.js
```diff
@@ -117,10 +119,21 @@ export default class ActiveGameManager extends EventEmitter {
     if (current && current.config) {
       this.emit('gameCommand', gameId, 'routes', routes)
       this.emit('gameCommand', gameId, 'setupGame', current.config.setup)
-      this.emit('gameCommand', gameId, 'allowStart')
     }
   }
 
+  allowStart(gameId, waitTime = 2000) {
```
tec27, line 125: Not quite clear what the wait time here is intended to accomplish, is it meant to be like a minimum amount of time to show the loading screen for? Perhaps we should keep that value elsewhere so we're sure it's the same for all clients if so (passed from the server? or at least stored in a common constant somewhere?)

Could even consider structuring this such that the 'allowStart' is triggered directly from the server for all clients (that is, the server emits another event when it's okay for clients to start)

client/active-game/active-game-manager.js
2Pacalypse-: Oops, you're right. Will fix.

client/active-game/active-game-manager.js
2Pacalypse-: Yeah, the original intention of wait time was to show the loading screen for a minimum amount of time. I didn't want to make it a constant because I thought it might be possible for different gameplay activities to display a loading screen in a different way.

Your point about keeping it same for all clients stands though; and I actually initially planned on making the server in charge of dealing with it, but decided against it as this seemed more straightforward and I didn't see any downsides. But yeah, I'll move it to the server since it makes more sense for server to keep all clients in sync about starting a game.

PR #491, opened 2020-07-21
Implement the UX/UI of the matchmaking loading process.

2Pacalypse-:
The matchmaking loading process starts once all the players accept the
game, and is executed in the following order:
- The players are moved to the "map selection" screen, where the map
  pool for that match is shown and the chosen map is highlighted. That
  screen is shown for 5 seconds and then...
- The players are moved to the "countdown" screen, which is counting
  down from 5 down to a 0, at which point...
- The `allowStart` command is sent to the game (game process was started
  at the beginning of the loading process), and if the game process
  didn't manage to start yet, a "Game is starting..." message is shown
  to the user until it does, after which...
- The game has fully loaded and the users are shown the "Game in
  progress..." message, while being moved to the game to start playing.

Couple of notes about this PR:
- The left nav entry for the matchmaking game while it's loading /
  in-progress is currently broken. It will be fixed in the next PR.
- The map selection screen only highlights the chosen map for the 5
  seconds. It might be more fun to have a highlight indicator randomly
  switch between all of the maps, before settling on the chosen one. PRs
  welcome for this ;)
- The way we display the players will have to change, soon. Currently
  we're only sending player *slots* when the match is ready (since it's
  needed for the game to actually start), but eventually we should send
  actual player's info as well (e.g. their rank, profile image, etc.).

### DIFF
--- a/server/lib/wsapi/matchmaking.js
+++ b/server/lib/wsapi/matchmaking.js
@@ -173,10 +173,10 @@ export class MatchmakingApi {
       })
 
       const [preferredMaps, randomMaps] = await Promise.all([
-        await getMapInfo(preferredMapsHashes.toJS()),
-        await getMapInfo(randomMapsHashes),
+        getMapInfo(preferredMapsHashes.toJS()),
+        getMapInfo(randomMapsHashes),
       ])
-      if (!preferredMaps.length && !randomMaps.length) {
+      if (!(preferredMaps.length + randomMaps.length)) {
         throw new Error('no maps found')
       }
 
--- a/server/lib/wsapi/matchmaking.js
+++ b/server/lib/wsapi/matchmaking.js
@@ -1,4 +1,4 @@
-import { List, Map, Record } from 'immutable'
+import { List, Map, Range, Record, Set } from 'immutable'
 import errors from 'http-errors'
 import { Mount, Api, registerApiRoutes } from '../websockets/api-decorators'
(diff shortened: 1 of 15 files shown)


server/lib/wsapi/matchmaking.js
```diff
@@ -138,50 +155,119 @@ export class MatchmakingApi {
...
+
+      const randomMapsHashes = []
+      Range(preferredMapsHashes.size, 4).forEach(() => {
+        const availableMaps = mapPool.subtract(preferredMapsHashes.concat(randomMapsHashes))
+        const randomMap = availableMaps.toList().get(getRandomInt(availableMaps.size))
+        randomMapsHashes.push(randomMap)
+      })
+
+      const [preferredMaps, randomMaps] = await Promise.all([
+        await getMapInfo(preferredMapsHashes.toJS()),
```
tec27, line 176: since you're awaiting the outer Promise, you don't want to await these inner ones (or you're ending up blocking on them here anyway)

server/lib/wsapi/matchmaking.js
```diff
@@ -138,50 +155,119 @@ export class MatchmakingApi {
...
+        await getMapInfo(randomMapsHashes),
+      ])
+      if (!preferredMaps.length && !randomMaps.length) {
```
tec27, line 179: Seems like we might want to check that *all* the potential maps were found

PR #496, opened 2020-08-10
Add a basic implementation of an Aws file store.

2Pacalypse-:
Aws file store can be used by any service that is compatible with
aws-sdk. For now, we're only allowing the usage of DigitalOcean Spaces,
but it shouldn't be much of a problem to support Amazon S3 as well, if
needed.

### DIFF
--- a/sample.env
+++ b/sample.env
@@ -50,15 +50,11 @@ SB_RALLY_POINT_SERVERS={"local":{"desc":"Local","address":"::ffff:127.0.0.1","po
 #SB_ROUTE_CREATOR_PORT=14099
 
 
-# A string denoting which file store implementation to use. Must match one of
-# the implementations below (either "filesystem" or "dospaces" for now).
-SB_FILE_STORE="filesystem"
 # JSON string configuring the storage of uploaded files. Currently there are
 # two options, storing to the filesystem and storing to the DigitalOcean Spaces.
-# For the filesystem option, you can change the path (relative to the server
-# directory) to change where it writes to. For the Spaces option, you *must* set
-# the URL to the correct endpoint, and matching access keys.
-SB_FILE_STORE_IMPLEMENTATIONS={"filesystem":{"path":"server/uploaded_files"},"dospaces":{"endpoint":"region.digitaloceanspaces.com","accessKeyId":"ACCESS_KEY_ID","secretAccessKey":"SUPER_SECRET_ACCESS_KEY","bucket":"shieldbattery"}}
+SB_FILE_STORE={"filesystem":{"path":"server/uploaded_files"}}
+# Example DigitalOcean Spaces configuration
+#SB_FILE_STORE={"doSpaces":{"endpoint":"region.digitaloceanspaces.com","accessKeyId":"ACCESS_KEY_ID","secretAccessKey":"SUPER_SECRET_ACCESS_KEY","bucket":"shieldbattery"}}
 
 
 # Configuration for BW sprite data used for map thumbnail generation
--- a/sample.env
+++ b/sample.env
@@ -50,10 +50,15 @@ SB_RALLY_POINT_SERVERS={"local":{"desc":"Local","address":"::ffff:127.0.0.1","po
 #SB_ROUTE_CREATOR_PORT=14099
 
 
-# JSON string configuring the storage of uploaded files. Currently there is only
-# one option, storing to the filesystem. You can change the path (relative to
-# the server directory) to change where it writes to.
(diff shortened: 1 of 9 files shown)


sample.env
```diff
@@ -50,10 +50,15 @@ SB_RALLY_POINT_SERVERS={"local":{"desc":"Local","address":"::ffff:127.0.0.1","po
...
-SB_FILE_STORE={"filesystem":{"path":"server/uploaded_files"}}
+# A string denoting which file store implementation to use. Must match one of
+# the implementations below (either "filesystem" or "dospaces" for now).
+SB_FILE_STORE="filesystem"
+# JSON string configuring the storage of uploaded files. Currently there are
+# two options, storing to the filesystem and storing to the DigitalOcean Spaces.
+# For the filesystem option, you can change the path (relative to the server
+# directory) to change where it writes to. For the Spaces option, you *must* set
+# the URL to the correct endpoint, and matching access keys.
+SB_FILE_STORE_IMPLEMENTATIONS={"filesystem":{"path":"server/uploaded_files"},"dospaces":{"endpoint":"region.digitaloceanspaces.com","accessKeyId":"ACCESS_KEY_ID","secretAccessKey":"SUPER_SECRET_ACCESS_KEY","bucket":"shieldbattery"}}
```
tec27, line 61: I don't really understand the point of having two config vars here, this one just contains a key (`dospaces`) that isn't even used for anything? Why not just keep the original key and if it contains a `filesystem` key, we store to the file system, and if it stores a `doSpaces` key, we store to Spaces? I think that was the original intent

If you're trying to give an example of the configuration for both, just include that in a commented out line

server/lib/file-upload/aws.js
```diff
@@ -0,0 +1,53 @@
...
+  async delete(filename, options = {}) {
+    const normalized = this._getNormalizedPath(filename)
+    const params = { Key: normalized, Bucket: this.bucket, ...options }
+    return this.client.deleteObject(params).promise()
+  }
+
+  async url(filename, options = {}) {
+    const normalized = this._getNormalizedPath(filename)
+    const params = { Key: normalized, Bucket: this.bucket, ...options }
+    const getSignedUrlAsPromise = util.promisify(this.client.getSignedUrl.bind(this.client))
```
tec27, line 50: promisifying a bound method every time this is called just seems like a bad idea to me, but I don't have performance data to back that up. Suffice to say, I would probably write out the relevant code instead (or bind this method once in the constructor and promisify it then)

server/app.js
```diff
@@ -74,7 +75,28 @@ if (!process.env.SB_FILE_STORE) {
   throw new Error('SB_FILE_STORE must be specified')
 }
 const fileStoreSettings = JSON.parse(process.env.SB_FILE_STORE)
-setStore(new LocalFileStore(fileStoreSettings.filesystem))
+if (!(fileStoreSettings.filesystem || fileStoreSettings.doSpaces)) {
```
tec27, line 78: nit: I'd probably just check that this is truthy here, and throw a different error in an `else` clause below (if no valid key could be found in it). That will make it slightly easier to add/change the keys in the future (even though that's probably rare :) )

server/lib/file-upload/aws.js
```diff
@@ -0,0 +1,76 @@
...
+  _getNormalizedPath(filename) {
+    // Force posix path separators on aws-compatible services which use it to create faux folders.
+    const normalized = path.posix.normalize(filename)
+    if (path.isAbsolute(normalized) || normalized[0] === '.') {
+      throw new Error('Invalid directory')
+    }
+    return normalized
+  }
+
+  async write(filename, stream, options = {}) {
```
tec27, line 53: Can you add a comment above this that specifies what valid options are/what their format is? (same for the other methods that take options). Doesn't need to be completely exhaustive, but at least what we convert from our style to AWS-API capitalization would be good

server/lib/file-upload/local-filesystem.js
```diff
@@ -6,9 +6,7 @@ import path from 'path'
 import thenify from 'thenify'
 import util from 'util'
 
-// How long browsers can cache resources for (in milliseconds). These resources should all be pretty
-// static, so this can be a long time
-const FILE_MAX_AGE_MS = 14 * 24 * 60 * 60 * 1000
+import { FILE_MAX_AGE_MS } from './index.js'
```
tec27, line 9: This feels like a pretty weird import to me (I'd expect imports to go in the other direction), maybe this constant should be in a different file?

PR #497, opened 2020-08-19
Implement a promise-based queue and use it for map parsing.

2Pacalypse-:


server/lib/maps/store.js
```diff
@@ -1,16 +1,18 @@
 import childProcess from 'child_process'
 import fs from 'fs'
+import bl from 'bl'
 import { addMap } from '../models/maps'
 import { writeFile } from '../file-upload'
-import bl from 'bl'
+import Queue from '../../../common/async/promise-queue'
 
 const BW_DATA_PATH = process.env.SB_SPRITE_DATA || ''
+const mapQueue = new Queue(Number(process.env.SB_MAP_PARSER_MAX_CONCURRENT))
```
tec27, line 9: This has the effect of making this environment variable required, but you don't actually check things in that way so you'd result in some very bad failure cases (namely, you're going to propagate NaN through a bunch of places and the checks against it aren't going to do what you expect).

You should either fail very loudly if this isn't set, or only do the `Number` conversion if it's not undefined (and in the undefined case, the default value would be used)

PR #498, opened 2020-08-27
Implement ability to delete all maps.

2Pacalypse-:
This feature enables users with a special permission to delete all maps
from the database, and their associated files. Use with care.

### DIFF
--- a/client/admin/permissions-reducer.js
+++ b/client/admin/permissions-reducer.js
@@ -14,6 +14,7 @@ export const Permissions = new Record({
   banUsers: false,
   manageMaps: false,
   manageMapPools: false,
+  deleteMaps: false,
 
   lastUpdated: 0,
   isRequesting: false,
--- a/client/admin/permissions-reducer.js
+++ b/client/admin/permissions-reducer.js
@@ -14,7 +14,7 @@ export const Permissions = new Record({
   banUsers: false,
   manageMaps: false,
   manageMapPools: false,
-  deleteMaps: false,
+  massDeleteMaps: false,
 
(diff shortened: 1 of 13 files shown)


client/admin/permissions-reducer.js
```diff
@@ -14,6 +14,7 @@ export const Permissions = new Record({
   banUsers: false,
   manageMaps: false,
   manageMapPools: false,
+  deleteMaps: false,
```
tec27, line 17: I wonder if this should be called like `deleteAllMaps` or `massDeleteMaps` or something, since we might want to allow a larger group of people delete/deactivate an individual map? Maybe doesn't matter, not sure

tec27: And I guess looking further it sorta seems like the case to me that these should be separate (since deleting a single map would probably just be flagging it as such, rather than actually removing the row/references elsewhere?)

server/lib/models/permissions.js
```diff
@@ -9,57 +10,56 @@ class Permissions {
...
 export async function updatePermissions(userId, perms) {
-  const query = `
-    UPDATE permissions SET edit_permissions = $2, debug = $3, accept_invites = $4,
-        edit_all_channels = $5, ban_users = $6, manage_maps = $7, manage_map_pools = $8
-    WHERE user_id = $1
-    RETURNING *
+  const query = sql`
+    UPDATE permissions
+    SET
+      edit_permissions = ${!!perms.editPermissions},
```
tec27, line 48: god I'm so happy we moved to this template string library for this stuff :)

PR #499, opened 2020-08-30
Generate a smaller sized image for all maps and use it as a thumbnail.

2Pacalypse-:
This PR assumes we're only gonna generate two images for each map, so
it doesn't bother making the code too generic to work for N amount of
images. That seems like a reasonable assumption for now. If it ever
breaks, some kind of an array format will need to be used to work with
the image URLs, and their saving path might need to be rethought.

### DIFF
--- a/server/lib/maps/map-parse-worker.js
+++ b/server/lib/maps/map-parse-worker.js
@@ -52,8 +52,7 @@ function generateImage(map, bwDataPath, width = 1024) {
     return Promise.resolve()
   }
 
-  const aspectRatio = map.size[0] / map.size[1]
-  const height = Math.floor(width / aspectRatio)
+  const height = Math.round((width * map.size[1]) / map.size[0])
 
   return map.image(Chk.fsFileAccess(bwDataPath), width, height, { melee: true }).then(imageRgb => {
     const rgbaBuffer = Buffer.alloc(width * height * 4)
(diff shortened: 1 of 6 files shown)


tec27:
I think this is missing the 2x density stuff we had talked about (I think the general structure of it otherwise still works fine, it just needs to generate images with a 1x/2x suffix in the filename or something similar, and the client-side components need to be updated to reference both in a `srcset`)

server/lib/maps/map-parse-worker.js
```diff
@@ -46,84 +46,80 @@ function createLobbyInitData(chk) {
...
-      height = 1024
-      width = map.size[0] * Math.floor(1024 / map.size[1])
+// Creates an image with the provided width and calculates height so the aspect ratio is preserved.
+function generateImage(map, bwDataPath, width = 1024) {
+  if (!bwDataPath) {
+    return Promise.resolve()
+  }
+
+  const aspectRatio = map.size[0] / map.size[1]
+  const height = Math.floor(width / aspectRatio)
```
tec27, line 56: My gut instinct says a `round` is probably more appropriate here (although it probably makes a small difference)

tec27: I also wonder if it might be less error-inducing to do this via a multiplication + divide, rather than 2 divides of smaller numbers. i.e. rearrange the formulas:

```
aspectRatio = actualWidth / actualHeight;
newHeight = desiredWidth / aspectRatio;
=>
newHeight = desiredWidth / (actualWidth / actualHeight)
=>
newHeight = desiredWidth * actualHeight / actualWidth
```

tec27 (approved):

PR #501, opened 2020-09-11
Create a MapImage component that should be used as a primitive for

2Pacalypse-:
displaying map images.

Also uses it throughout the app. There are two components that are a
level higher than MapImage component, which are MapThumbnail and
MapPreview. They can be used in places where some common functionality
is required to make things easier (though MapPreview is pretty bare
atm).

This PR also updates the naming scheme of the map image names and URLs
to be a bit more cleaner and consistent. They are also saved in the same
folder now.

### DIFF
--- a/client/lobbies/loading.jsx
+++ b/client/lobbies/loading.jsx
@@ -242,7 +242,7 @@ const GameTypeMapBridge = styled(Display1)`
   color: ${colorTextSecondary};
 `
 
-const MapThumbnail = styled.div`
+const MapImageContainer = styled.div`
   ${shadow1dp};
   width: 256px;
   height: auto;
@@ -301,9 +301,9 @@ export default class LoadingScreen extends React.Component {
           <GameTypeMapBridge as='span'> on </GameTypeMapBridge>
           <Display1 as='span'>{lobby.map.name}</Display1>
         </div>
-        <MapThumbnail>
-          <MapImage map={lobby.map} showNotAvailableText={true} />
-        </MapThumbnail>
+        <MapImageContainer>
+          <MapImage map={lobby.map} />
+        </MapImageContainer>
         <Players>{playerElems}</Players>
         <LoadingMessage />
       </Content>
--- a/client/lobbies/loading.jsx
+++ b/client/lobbies/loading.jsx
@@ -4,6 +4,7 @@ import { getPlayerSlots } from '../../common/lobbies'
 import gameTypeToString from './game-type-to-string'
 import styled from 'styled-components'
 
+import MapImage from '../maps/map-image.jsx'
 import PlayerCard from './player-card.jsx'
 
 import { colorTextSecondary } from '../styles/colors'
@@ -241,12 +242,13 @@ const GameTypeMapBridge = styled(Display1)`
   color: ${colorTextSecondary};
 `
 
-const MapThumbnail = styled.img`
+const MapThumbnail = styled.div`
   ${shadow1dp};
   width: 256px;
   height: auto;
   border-radius: 2px;
   margin-top: 16px;
+  overflow: hidden;
 `
 
 const Players = styled.div`
@@ -299,9 +301,9 @@ export default class LoadingScreen extends React.Component {
           <GameTypeMapBridge as='span'> on </GameTypeMapBridge>
           <Display1 as='span'>{lobby.map.name}</Display1>
         </div>
-        <div>
-          <MapThumbnail src={lobby.map.imageUrl} />
-        </div>
+        <MapThumbnail>
+          <MapImage map={lobby.map} showNotAvailableText={true} />
+        </MapThumbnail>
         <Players>{playerElems}</Players>
         <LoadingMessage />
       </Content>
--- a/client/lobbies/lobby-reducer.js
+++ b/client/lobbies/lobby-reducer.js
@@ -24,6 +24,7 @@ import {
   LOBBY_UPDATE_LOADING_START,
   LOBBY_UPDATE_LOADING_CANCELED,
   LOBBY_UPDATE_CHAT_MESSAGE,
+  MAPS_TOGGLE_FAVORITE,
   NETWORK_SITE_CONNECTED,
 } from '../actions'
 
@@ -152,6 +153,12 @@ const infoReducer = keyedReducer(undefined, {
 
   [NETWORK_SITE_CONNECTED](state, action) {
     return new LobbyInfo()
+  },
+
+  [MAPS_TOGGLE_FAVORITE](state, action) {
+    const { map } = action.meta
+
+    return state.setIn(['map', 'isFavorited'], !map.isFavorited)
   },
 })
 
--- a/client/lobbies/view.jsx
+++ b/client/lobbies/view.jsx
@@ -7,6 +7,10 @@ import styled from 'styled-components'
 import Lobby from './lobby.jsx'
 import LoadingScreen from './loading.jsx'
 import LoadingIndicator from '../progress/dots.jsx'
+import MapPreview from '../maps/map-preview.jsx'
+
+import { openSimpleDialog } from '../dialogs/action-creators'
+import { toggleFavoriteMap } from '../maps/action-creators'
 
 import {
   addComputer,
@@ -33,6 +37,7 @@ const mapStateToProps = state => {
     lobby: state.lobby,
     gameClient: state.gameClient,
     activeGame: state.activeGame,
+    maps: state.maps,
   }
 }
 
@@ -116,7 +121,7 @@ export default class LobbyView extends React.Component {
 
   renderLobby = () => {
     const routeLobby = this.props.match.params.lobby
-    const { lobby, user } = this.props
+    const { lobby, maps, user } = this.props
 
     let content
     if (!lobby.inLobby) {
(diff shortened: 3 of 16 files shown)


client/lobbies/view.jsx
```diff
@@ -129,6 +134,7 @@ export default class LobbyView extends React.Component {
           lobby={lobby.info}
           chat={lobby.chat}
           user={user}
+          isFavoritingMap={maps.favoriteStatusRequests.has(lobby.info.map.id)}
```
tec27, line 137: I don't think the `favoriteStatusRequests` thing is new but probably a better way to handle this is to update the map's state to favorited, then set it back to unfavorited if that request fails (and vice versa). Makes the handling a lot simpler here and everywhere else that has a favorite icon

client/maps/map-image.jsx
```diff
@@ -0,0 +1,74 @@
...
+  render() {
+    const { map, size, showNotAvailableText } = this.props
+    const srcSet = `
+      ${map.image256Url} 256w,
+      ${map.image512Url} 512w,
+      ${map.image1024Url} 1024w,
+      ${map.image2048Url} 2048w
+    `
+
+    // TODO(2Pac): Actually check if the image URL is available instead of just checking it's set.
```
tec27, line 53: More accurately, probably "handle 404s"

client/maps/map-image.jsx
```diff
@@ -0,0 +1,74 @@
...
+import PropTypes from 'prop-types'
+import styled from 'styled-components'
+
+import ImageIcon from '../icons/material/baseline-image-24px.svg'
+
+import { grey800 } from '../styles/colors'
+import { Subheading } from '../styles/typography'
+
+const ImgElement = styled.img`
+  display: block;
```
tec27, line 11: I'm not sure these styles really make sense for the generally usable component, it'd feel more appropriate for the particular usage in the map browser or whatever

client/maps/map-image.jsx
```diff
@@ -0,0 +1,74 @@
...
+  width: 100%;
+  object-fit: cover;
+`
+
+const NoImage = styled.div`
+  display: flex;
```
tec27, line 17: These styles (especially the sizes) feel *very* particular to the map browser, and I'm not sure they should really be in here

client/maps/map-image.jsx
```diff
@@ -0,0 +1,74 @@
...
+    const srcSet = `
+      ${map.image256Url} 256w,
+      ${map.image512Url} 512w,
+      ${map.image1024Url} 1024w,
+      ${map.image2048Url} 2048w
+    `
+
+    // TODO(2Pac): Actually check if the image URL is available instead of just checking it's set.
+    return (
+      <>
```
tec27, line 55: This component now returns either:
- An image element that is styleable
- A div that is not at all styleable

That doesn't feel like a good API to me, ideally this component would have the same degree of styleability regardless of whether the image is present. It's going to get *especially* bad if you start handling 404s, where that can change on the fly. Since you feel like the NoImage case needs to have an option for text anyway, maybe a better option is to allow the caller to provide the element to use if the image isn't available? That'd fix the issues I noted above, and also remove the need for that configuration option.

client/maps/map-image.jsx
```diff
@@ -0,0 +1,74 @@
...
+        {map.image256Url ? (
+          <ImgElement
+            className={this.props.className}
+            srcSet={srcSet}
+            sizes={`${size}px`}
+            src={map.image256Url}
+            alt={map.name}
+            draggable={false}
```
tec27, line 63: I think this is already covered by our CSS and you don't need to set this here

client/maps/map-image.jsx
```diff
@@ -0,0 +1,74 @@
...
+    // TODO(2Pac): Actually check if the image URL is available instead of just checking it's set.
+    return (
+      <>
+        {map.image256Url ? (
+          <ImgElement
+            className={this.props.className}
+            srcSet={srcSet}
+            sizes={`${size}px`}
+            src={map.image256Url}
+            alt={map.name}
```
tec27, line 62: Might want to make this customizable since the map name might not be the most appropriate description for the image in a particular UI (although falling back to the map name seems fine to me)

tec27 (changes requested):

client/lobbies/loading.jsx
```diff
@@ -241,12 +242,13 @@ const GameTypeMapBridge = styled(Display1)`
...
 `
 
-const MapThumbnail = styled.img`
+const MapThumbnail = styled.div`
   ${shadow1dp};
   width: 256px;
   height: auto;
   border-radius: 2px;
   margin-top: 16px;
+  overflow: hidden;
```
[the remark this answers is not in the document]
2Pacalypse-: I actually added it as a way to show the rounded corners on the map image (since this is now a container that has MapImage as a child). It was either this or styling the map image directly; not sure which way is preferable?

client/lobbies/lobby-reducer.js
```diff
@@ -153,6 +154,12 @@ const infoReducer = keyedReducer(undefined, {
   [NETWORK_SITE_CONNECTED](state, action) {
     return new LobbyInfo()
   },
+
+  [MAPS_TOGGLE_FAVORITE](state, action) {
```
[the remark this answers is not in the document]
2Pacalypse-: Well, right now the lobby state contains the whole MapRecord of the chosen map. I guess we could update it to only contain a map ID or something, and keep the actual map info in the maps reducer, but I feel like this would complicate things a lot. Currently, the maps reducer is only used by the server-maps-browser (I actually thought of renaming that reducer to server-maps-reducer, since we also have a local-maps-reducer), and every other place that fetches maps keep them in their own reducers, e.g. find-match, preferred-maps, matchmaking-map-pools, create-lobby, lobby, etc.

When I was working on this, I did try to think of ways to keep the maps state in one place, but it proved too hard for me. Maybe I was missing something though :/

client/lobbies/view.jsx
2Pacalypse-: The only reason why I added this is to disable the favorite/unfavorite button while the request hasn't finished yet. Not sure what would be the best way to deal with the user pressing a button again while the first request hasn't finished yet with that approach?

I guess it would still be worth doing since it is pretty annoying to use this state every time you want to add a favorite icon on map.

client/maps/map-image.jsx
2Pacalypse-: `display: block` is needed in lobbies actually, but I get your point. Would you say that the `img` element should have no styling in general use case?

client/maps/map-image.jsx
2Pacalypse-: Yeah, unstyleability of the NoImage case was annoying me as well :d
Do you think this component should still have some kind of a default way to display a NoImage case?

client/lobbies/loading.jsx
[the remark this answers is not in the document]
tec27: Ah, forgot about rounded corners, yeah that's fine 👍

client/lobbies/lobby-reducer.js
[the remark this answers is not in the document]
tec27: It definitely feels to me like this data *should* be in one place, indexable by an ID or whatever. Otherwise every place that involves a map has to have its own way of getting that data from the server, its own way of updating that data (favoriting, editing, etc), has to handle all of the various map events to ensure the data stays up-to-date, etc.

I think the idea of handling a single event in many different ways throughout our reducers is sound, but when that handling boils down to the exact same desired outcome on nearly the exact same data structure, it's far better to put all that logic/data in a single spot.

I'm fine with this being this way for this PR, since it already exists this way, but it's something we should be looking to move away from imo

client/lobbies/view.jsx
tec27: I don't think you really *need* to disable that. If you submit two requests in a row they're pretty likely to be delivered in that order and succeed in that order. The main complexity is what happens if many fail in succession, but I think you can solve that by keeping track of the "last known state" which is either what you last retrieved from the server directly, or what you know succeeded. If you retrieve from the server you update that value, if a request promise succeeds you update that value. If the last outstanding promise fails (and *only* if it's the last outstanding one), you reset the current value to whatever that last known state is.

imo this current state looks very much like internal state that things generally shouldn't be accessing, which is why I'm wary of needing to pass it to UI elements, especially frequently.

client/maps/map-image.jsx
tec27: I think pulling out the current one into a component and making that the default if it's not specified would be fine

client/maps/map-image.jsx
tec27: That particular style was the least concerning of them, I was more talking about the following two. I think generally I'd try to avoid specifying layout-type things in the styles for base UI components, as it makes them harder/weirder to use in specific UIs. For example, it'd be weird to set the width of a button to 100%, as it'd then require you to put it in a container with a particular size to make it smaller (or override the style with styled-components, I suppose, but same idea). Not applicable *all* the time, but I think you at least need a really good reason to include stuff like that at the root level of a UI component.

client/maps/map-image.jsx
2Pacalypse-: Hmm, I tried removing this and it made images draggable (by displaying a ghost image while dragging). I think we only set CSS to disable user _selection_ of elements, which apparently is a different thing to dragging?

PR #512, opened 2020-09-22
Allow user to change their password while logged in.

### LINKED ISSUE #507: Add a way to change your password while logged in
We can currently work around not having this feature by using the Forgot Password thing, but it'd be nice to just let logged in users change their passwords. Probably needs to include notifying the user via email that their password has changed, and probably giving them a link or something to undo that change for a small amount of time.

2Pacalypse-:
Fixes #507

### DIFF
--- a/client/actions.js
+++ b/client/actions.js
@@ -34,6 +34,7 @@ export const ADMIN_SET_PERMISSIONS = 'ADMIN_SET_PERMISSIONS'
 // The AudioManager has finished initializing and is ready to play sounds
 export const AUDIO_MANAGER_INITIALIZED = 'AUDIO_MANAGER_INITIALIZED'
 
+export const AUTH_ACCOUNT_UPDATE = 'AUTH_ACCOUNT_UPDATE'
 export const AUTH_CHANGE_BEGIN = 'AUTH_CHANGE_BEGIN'
 export const AUTH_LOG_IN = 'AUTH_LOG_IN'
 export const AUTH_LOG_OUT = 'AUTH_LOG_OUT'
--- a/client/auth/edit-account.jsx
+++ b/client/auth/edit-account.jsx
@@ -1,0 +1,194 @@
+import React from 'react'
+import { connect } from 'react-redux'
+import styled from 'styled-components'
+
+import Avatar from '../avatars/avatar.jsx'
+import Dialog from '../material/dialog.jsx'
+import FlatButton from '../material/flat-button.jsx'
+import form from '../forms/form.jsx'
+import LoadingIndicator from '../progress/dots.jsx'
+import PasswordTextField from '../material/password-text-field.jsx'
+import SubmitOnEnter from '../forms/submit-on-enter.jsx'
+
+import { closeDialog } from '../dialogs/action-creators'
+import { updateAccount } from '../auth/auther'
+import { composeValidators, matchesOther, minLength, required } from '../forms/validators'
+import { PASSWORD_MINLENGTH } from '../../common/constants'
+
+import { colorError } from '../styles/colors'
+import { Subheading } from '../styles/typography'
+
+const passwordValidator = composeValidators(
+  required('Enter a password'),
+  minLength(PASSWORD_MINLENGTH, `Enter at least ${PASSWORD_MINLENGTH} characters`),
+)
+const newPasswordValidator = composeValidators(
+  required('Enter a new password'),
+  minLength(PASSWORD_MINLENGTH, `Enter at least ${PASSWORD_MINLENGTH} characters`),
+)
+const confirmNewPasswordValidator = composeValidators(
+  required('Confirm your new password'),
+  matchesOther('newPassword', 'Enter a matching password'),
+)
+
+@form({
+  currentPassword: passwordValidator,
+  newPassword: newPasswordValidator,
+  confirmNewPassword: confirmNewPasswordValidator,
+})
+class AccountForm extends React.Component {
+  render() {
+    const { bindInput, onSubmit } = this.props
+    const textInputProps = {
+      autoCapitalize: 'off',
+      autoCorrect: 'off',
+      spellCheck: false,
+      tabIndex: 1,
+    }
+
+    return (
+      <form noValidate={true} onSubmit={onSubmit}>
+        <SubmitOnEnter />
+        <PasswordTextField
+          {...bindInput('currentPassword')}
+          label='Current password'
+          floatingLabel={true}
+          inputProps={textInputProps}
+        />
+        <PasswordTextField
+          {...bindInput('newPassword')}
+          label='New password'
+          floatingLabel={true}
+          inputProps={textInputProps}
+        />
+        <PasswordTextField
+          {...bindInput('confirmNewPassword')}
+          label='Confirm new password'
+          floatingLabel={true}
+          inputProps={textInputProps}
+        />
+      </form>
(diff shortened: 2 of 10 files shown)


client/auth/edit-account.jsx
```diff
@@ -0,0 +1,194 @@
...
+    return (
+      <form noValidate={true} onSubmit={onSubmit}>
+        <SubmitOnEnter />
+        <PasswordTextField
+          {...bindInput('currentPassword')}
+          label='Current password'
+          floatingLabel={true}
+          inputProps={textInputProps}
+        />
+        <PasswordTextField
```
tec27, line 58: It doesn't really make sense atm, but will once we add email changing as well: I sort of like Discord's pattern here. Editing any of this information requires the current password, but the new password textboxes (new + confirm) aren't visible until users click a 'Change password?' link underneath the current password input. Would probably be good to do something similar here.

(And also, once email is added, the new password would only be required if you were trying to change it. It seems like Discord only validates if you've actually entered something into the new password box)

client/auth/edit-account.jsx
```diff
@@ -0,0 +1,194 @@
...
+        <LoadingArea>
+          <LoadingIndicator />
+        </LoadingArea>
+      )
+    }
+
+    if (reqId && auth.lastFailure && auth.lastFailure.reqId === reqId) {
+      errorElem = (
+        <ErrorText>
+          There was an issue with updating your account. Please check your data and try again.
```
tec27, line 154: Not a huge fan of this error message :)

It's safe here to tell someone if they've entered a wrong password (because they already know this user exists, they're logged into it). We probably want to show something specific for that, otherwise something like:

There was an issue updating your account. Please try again later.

server/lib/mail/templates/password-change.html
```diff
@@ -0,0 +1,63 @@
+<!DOCTYPE html>
```
tec27, line 1: I don't see the equivalent of this in `mail-templates/`, did you forget to add it to this commit?

server/lib/mail/templates/password-change.txt
```diff
@@ -0,0 +1,15 @@
+******************************
+ShieldBattery Password Changed
+******************************
+
+You are receiving this email because
```
tec27, line 5: This sounds like our system is inherently insecure, would probably phrase differently. Maybe something like:
`You are receiving this email because the password for {{username}} on ShieldBattery has been changed. If you did not change it, please contact the ShieldBattery team.`

We may also want to direct them to the password reset page rather than having them contact us, I feel like that's probably where we'd end up sending them anyway.

tec27 (changes requested):

client/actions.js
```diff
@@ -34,6 +34,7 @@ export const ADMIN_SET_PERMISSIONS = 'ADMIN_SET_PERMISSIONS'
 // The AudioManager has finished initializing and is ready to play sounds
 export const AUDIO_MANAGER_INITIALIZED = 'AUDIO_MANAGER_INITIALIZED'
 
+export const AUTH_ACCOUNT_UPDATE = 'AUTH_ACCOUNT_UPDATE'
```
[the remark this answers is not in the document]
2Pacalypse-: Yeah, since we're using ID-ed requests for auth actions, we have a single action `AUTH_CHANGE_BEGIN` that is used there:
https://github.com/ShieldBattery/ShieldBattery/blob/master/client/auth/auther.js#L16

However, I'm not entirely sure if that should be applicable here?

client/auth/edit-account.jsx
2Pacalypse-: Yeah, I've thought a bit about the UI/UX of this form once we allow email changes (and even profile image changes), but decided to leave it for now and implement it once we allow email changes. Which should be in the next PR :)

And yeah, I agree with your (and Discords) suggestion here. Although I have a couple of questions about that since we're on the subject:
- Should we display the current email in the text field that allows email changes?
- And if so, what's the best way to get the email of a user? Stick it in the session info that user gets initialized with?
- Should changing profile image require current password to be entered as well?

client/auth/edit-account.jsx
2Pacalypse-: Oh, I wasn't sure about the security implications of that :)
I'll update it so it's clear when the entered current password is wrong. It would be nice if it were possible to even manually set the error of that text field in form (like what Discord does).

server/lib/mail/templates/password-change.html
2Pacalypse-: I did indeed :)
It's been a while since I composed an email for something; seems like there's quite a bit of repetition ^^.

client/actions.js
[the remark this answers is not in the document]
tec27: Feels a bit odd to me, but it's probably fine. We do use it in password reset too, so I guess using it here makes sense.

client/auth/edit-account.jsx
tec27: Showing the current email seems fine, it's already available in the session and sent in the AUTH_UPDATE action, the store just doesn't do anything with it.

Changing avatar without entering the current password seems fine.

client/auth/edit-account.jsx
tec27: I don't think a minlength validator on the current password makes much sense anyway, tbh, although adding custom errors when we have validations feels sorta weird. Might be better to try and build more support for that into the form stuff rather than just avoiding doing it on something that has validation functions.

2Pacalypse-: Actually, I was playing a bit with that too afterwards, but not really sure what's the best way to deal with it. Seems like the form would need a way to run some validations _after_ a submit was done? Which doesn't really make sense. Or it would need to expose a method which can set the internal error state for that field manually, similar to how `setInputValue` can be used to set its value.

Also, not having minLength validator does seem a bit weird though now that I've tried it, since the server will just return BadRequest in that case, which will display a generic "something went wrong message". I guess we can just not check that on the server either and instead just return "Incorrect password" instead.

tec27: Perhaps some system where the submission function could return a list of validation failures, or notify the form of validation failures from the submission in some way?

As far as the minlength stuff, I mostly meant that it doesn't make sense *provided* that we're checking that the field has content at all. No content should still be checked (and/or block submission entirely), but I guess that's still a validator that you're going to have trouble with. Returning invalid password for that case on the server seems fine too (and might be good to do regardless of this stuff), but I'd still like to fix the field-specific error message stuff in a cleaner way

2Pacalypse-: I haven't tried it yet, but I feel like having the submission function return _anything_ to the form might feel a bit awkward, especially since all it does now is just dispatches an action. But I also feel like our current representation of a form is only applicable until it gets submitted; after that, it doesn't seem like it should be form's responsibility anymore. Though I might be wrong about that.

But yeah, in any case I'm a bit stumped here and I don't see a clean way in which we can have validations _after_ the form gets submitted.

tec27: I think if you can't figure out a clean-ish way to do it, then it's better to use the less optimal UX pattern (an error that isn't directly on the field) and leave a TODO to find that clean way. I don't think the UX difference is so great that it's going to be worth the problems it will cause if/when we make changes to the form stuff.

tec27 (approved):

PR #515, opened 2020-09-29
SCR team game support

neivv:
Would be nice if someone verified that it works fine for them as well, as there already were more issues than the single one I expected there to be.

### DIFF
--- a/game/Cargo.lock
+++ b/game/Cargo.lock
@@ -629,7 +629,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 [[package]]
 name = "samase_scarf"
 version = "0.1.0"
-source = "git+https://github.com/neivv/samase_scarf/?rev=c4eff01bb4cdd88f773e96cd8c1e6e4d56d3f363#c4eff01bb4cdd88f773e96cd8c1e6e4d56d3f363"
+source = "git+https://github.com/neivv/samase_scarf/?rev=8a35243db57e7719b14798e8fee83fe81d6fb750#8a35243db57e7719b14798e8fee83fe81d6fb750"
 dependencies = [
  "arrayvec 0.5.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "byteorder 1.3.4 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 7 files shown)


2Pacalypse-:
I'm not entirely sure about the rules of the team melee games, but the following lobby configuration ends up with Team 2 being eliminated as soon as the game starts. Is that expected? 😃

image

neivv:
Nope, guess something is still broken

neivv:
Wasn't able to reproduce that though - are you sure the game dll was rebuilt?

2Pacalypse-:
Interestingly enough, the following configuration works without a problem:

image

And so does this one:

image

2Pacalypse-:
> Wasn't able to reproduce that though - are you sure the game dll was rebuilt?

Yeah, since it works in other cases.

neivv:
Then it seems to be something that's happening randomly :l

2Pacalypse-:
But it consistently fails with that configuration and works with others I posted.

Also, I went to test this race configuration on battle.net as well, but there when I chose a race for a computer, it selects that race for _all_ computer slots. So it seems impossible to select different races for a computer on battle.net. I assume that's a bug on their system?

And finally, the first race configuration I posted works with 2 human players. It seems to only fail if the second team is made of computers.

neivv:
Huh, that's really weird. Maybe the first computer spawning as random is causing issues then? And maybe it's related to the map as i didn't have issues with FS or regular hunters.

Can you check with 2 humans on the first team and computers on the other?

And 1161 allows selecting different races for the computer team (even if it is pretty pointless as the ai won't be able to take advantage of that), guess blizzard broke it with remastered.

neivv:
Nevermind, 1161 doesn't allow separate races for computers either. And turns out that setting one computer in a team to random but not all causes out-of-bounds memory reads.. Unless the lowest possible slot in ai team is random, which seems to be consistent with what you are seeing.

I probably happened to have the out-of-bounds data that still resulted in valid races or something while doing the first tests.

2Pacalypse-:
Huh, that's pretty weird. Should we update the UI as well to disallow setting different races to computers then, similar to what SC:R does? Even though _some_ race combinations are valid for a computer, I don't think anyone will really mind not being able to choose them :d

2Pacalypse- (approved):

PR #521, opened 2020-10-03
Fix window being hidden after starting in windowed fullscreen and switching to windowed in game.

neivv:


2Pacalypse- (approved):
Works great for me :)

PR #524, opened 2020-10-24
Add SC:R settings.

2Pacalypse-:
The SC:R settings are dealt with in a similar manner to our own local
settings. We watch the SC:R settings file and merge settings directly to
it. Also, we don't try to be smart and detecting which exact settings
have changed, and instead just simply save both the local and SC:R
settings every time.


2Pacalypse-:
A few things that could still use some work (either in this PR or open issues for them):
- add/remove/rename settings if needed
- improve spacing/ordering/design of the form elements
- number version of the text-field component (needed for APM Alert settings, which are left out for now)
- tooltip component (SC:R uses tooltips for various settings; unclear whether we need to do the same or can make things clear enough without them)
- radio button component (SC:R uses radio button component for some of its settings; we're using a Select component as a replacement for now)
- slider/tabs disabled state has been added as a part of this PR, but it could use a visual inspection (slider especially)
- settings dialog has variable height when switching tabs, which can have a sort of annoying jumpy effect

2Pacalypse-:
Also, I forgot to mention in the description, but this PR is renaming some of our local settings. So be mindful of that when trying out this branch as we don't have "undo" migration for local settings ^^.

client/actions.js
```diff
@@ -228,11 +228,11 @@ export const LOBBY_PREFERENCES_UPDATE_BEGIN = 'LOBBY_PREFERENCES_UPDATE_BEGIN'
 // The server has responded with success/failure to us updating the lobby preferences
 export const LOBBY_PREFERENCES_UPDATE = 'LOBBY_PREFERENCES_UPDATE'
 
-// We are starting the process of saving the settings
+// We are starting the process of saving the local settings
 export const LOCAL_SETTINGS_SET_BEGIN = 'LOCAL_SETTINGS_SET_BEGIN'
-// The server has responded with success/failure to our settings saving
+// The server has responded with success/failure to our local settings saving
```
tec27, line 233: Not really a server :)

client/actions.js
```diff
@@ -327,6 +327,13 @@ export const NETWORK_SITE_DISCONNECTED = 'NETWORK_SITE_DISCONNECTED'
 // We have started the process of launching the game with the selected replay in it
 export const REPLAYS_START_REPLAY = 'REPLAYS_START_REPLAY'
 
+// We are starting the process of saving the scr settings
```
tec27, line 330: nit: `SC:R` or `SCR` in these, don't think lowercase is clear enough

client/settings/settings-records.js
```diff
@@ -0,0 +1,129 @@
...
+  ['apmDisplayOn', 'apm_Showing'],
+  ['apmAlertOn', 'apm_AlertUser'],
+  ['apmAlertValue', 'apm_AlertValue'],
+  ['apmAlertColorOn', 'apm_AlertUseColor'],
+  ['apmAlertSoundOn', 'apm_AlertUseSound'],
+])
+
+export function fromScrToSb(scrSettings) {
+  const mappedScrSettings = Object.entries(scrSettings).reduce((acc, [name, value]) => {
+    const sbKeyName = scrSettingMappings.findKey(scrKeyName => scrKeyName === name)
```
tec27, line 111: I would *really* recommend just building the inverse map once at launch, otherwise this conversion is N^2. Should be simple, just:
```js
const sbToScrMapping = scrSettingMappings.entrySeq().map(([key, value]) => [value, key]).toMap()
```

client/settings/settings.jsx
```diff
@@ -88,16 +112,52 @@ export default class Settings extends React.Component {
           <AppSettings
             localSettings={local}
             formRef={this._form}
+            onChange={this.onSettingsChange}
             onSubmit={this.onSettingsSubmit}
           />
         )
-      case TAB_GAME:
+      case isRemastered ? SCR_TAB_INPUT : V1161_TAB_INPUT:
```
tec27, line 119: I honestly did not know you could do this kind of thing in a case statement and I am mildly scared :P

tec27: Might be better to move away from this idea that each of our settings tabs would support both remastered and 1.16, and instead write dedicated components for each (then you'd just have separate case statements here instead of this complexity)

tec27: Thinking some more, is it even really necessary to have separate identifiers for these things? We only display one or the other anyway

tec27 (changes requested):

app/settings.js
```diff
@@ -78,6 +59,58 @@ export default class LocalSettings extends EventEmitter {
...
+  }
+
+  _emitChange() {
+    this.emit('change', this._settings)
+  }
+}
+
+export class LocalSettings extends Settings {
+  constructor(filepath) {
+    const initializeFunc = async function () {
```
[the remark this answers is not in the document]
2Pacalypse-: Hmm, I'm not sure if I'm doing it wrong, but I get this error when I try that:

image

client/settings/settings.jsx
2Pacalypse-: Yeah, this all started by trying to hide Sound and Gameplay tabs from the 1.16.1 version since we don't have any of those settings for that version yet. So the the 1.16.1 version only had 3 tabs (App, Input, and Video) and the SC:R had all 5 (App, Input, Sound, Video, and Gameplay).

The first issue I noticed was due to the way I was rendering tab items, which was (I had to use the fragment instead of `null` when not rendering anything, since tab component is cloning the children to attach some props to them):
```js
<Tabs activeTab={activeTab} onChange={this.onTabChange}>
  <TabItem key='app' text='App' />
  <TabItem key='input' text='Input' disabled={isTabDisabled} />
  {isRemastered ? <TabItem key='sound' text='Sound' disabled={isTabDisabled} /> : <></>}
  <TabItem key='video' text='Video' disabled={isTabDisabled} />
  {isRemastered ? <TabItem key='gameplay' text='Gameplay' disabled={isTabDisabled} /> : <></>}
</Tabs>
```
The problem with this approach was that the tab component would incorrectly calculate the tab indicator size, since it calculates that based on the children count (which counts fragments as well). So the indicator size would be wrong in the 1.16.1 version since it only displayed 3 tabs, but counted 5.

So the next thing I did was I dynamically created the array of tabs based on the version and that fixed the indicator size problem. However, now the switch/case statement which used the tab value (which is just an index in an array), would incorrectly display the contents since the 1.16.1 version had fewer tabs. E.g. Video is the 4th tab in the SC:R version and 3rd tab in the 1.16.1 version. That's what cause the switch/case statement above to become weirdly complicated.

The _proper_ fix for this would be upgrading the Tab (and TabItem) components and make each TabItem have a defined value, so that we don't rely on array indices as its implicit value which we use to detect an active item here:
https://github.com/ShieldBattery/ShieldBattery/blob/master/client/material/tabs.jsx#L122

At the time I wrote the tab component I thought the usage of tabs will remain simple enough to not warrant doing that (as we already do in the Select component afaik). And I really didn't feel like doing that in this PR because this "bug" was the last thing I was fixing in it and I was already too tired at this point :d

If you think we should do it anyway in this PR, then I can muster up some mental fortitude and get it over with :p

tec27: Not sure I quite understand how this affects the switch statement, but I assume you know better than I do :) I'm fine with this for this PR, would probably just file an issue to make that change to the tabs component later

app/settings.js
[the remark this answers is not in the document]
tec27: Will play around with this to figure out why that's happening

tec27:
> add/remove/rename settings if needed

All labels should follow our standard style for capitalization, which means only the first letter should be capitalized. I think they're pretty much all not following that at the moment :P (Keyboard Scroll Speed -> Keyboard scroll speed, for example)

Ones I would probably change:
*Input*
Use Mouse Sensitivity -> Custom mouse sensitivity
Use Mouse Confine -> Lock cursor to window

*Sound*
Turn Music On -> Music
Turn Digital Sounds On -> Game sounds
Digital Volume -> Sound volume
Original 1998 Campaign -> Original campaign sounds? (I dunno we should probably get rid of this setting)

*Video*
Enable FPS Cap -> Enable FPS limit
Pillarbox -> Keep aspect ratio? Unsure really, but I basically guarantee most people don't know what "Pillarbox" means :)

*Gameplay*
Mini-Map Position -> Minimap position
Display APM In Game -> APM display

> improve spacing/ordering/design of the form elements

Yeah we really need the 2 column layout for most of these tabs, I think that would help a lot.

> number version of the text-field component (needed for APM Alert settings, which are left out for now)

Is this a new component we need to build? I would figure you could just pass a few attributes to the `input` and it should work okay today

> tooltip component (SC:R uses tooltips for various settings; unclear whether we need to do the same or can make things clear enough without them)

While this would certainly be nice to have, I think you could get away with some basic explanatory text around the settings that need it for now.

> radio button component (SC:R uses radio button component for some of its settings; we're using a Select component as a replacement for now)

Dropdowns seem fine to me, we don't really need to copy their choices there.

> slider/tabs disabled state has been added as a part of this PR, but it could use a visual inspection (slider especially)

Slider feels like it needs something more to differentiate it when disabled, especially when it's in the leftmost position. The color difference just isn't significant enough. I think we should probably bring our slider more in-line with the current guidelines, which means:

- Track color in the non-disabled state is a transparent version of the accent color
- The "uncolored when at the minimum setting" behavior goes away
- The filled part of the track is thicker than the unfilled part

That should make the disabled ones stand out more, I think.

As for tabs, I'm not sure I see a case where they are disabled now? So I can't really comment on that.

> settings dialog has variable height when switching tabs, which can have a sort of annoying jumpy effect

Yeah, ideally we would smoothly transition that or pick a minimum height that works for all the tabs. (The latter is definitely easier right now :) ). I don't think this will be as much of an issue when we move to a 2 column layout.

tec27:
I'm okay with you merging this without the 2 column layout and cleaning it up after (or I can try to clean it up after). I think I'd just try to organize things such that related settings are next to each other (Video is kind of a mixed bag on this right now, even in a single column layout, although I know that's how they have it ingame too :) )

app/settings.js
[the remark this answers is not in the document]
tec27: You know, I think we used to have a similar issue related to the code that gets generated by babel for async stuff when it's not natively supported. I imagine the simpler approach will probably work fine on electron 10, but you can leave this for now.

2Pacalypse-:
> Is this a new component we need to build? I would figure you could just pass a few attributes to the `input` and it should work okay today

Yeah, I was thinking of making a new component, similar to `PasswordTextField` that wraps the `TextField` component and deals with some specifics when handling numbers (like casting the value into a number, instead of using the text). Also, would be nice if we could style those up/down arrows ^^. But anyway, I just passed a few attributes as you said for now, which should be good enough I guess.

> While this would certainly be nice to have, I think you could get away with some basic explanatory text around the settings that need it for now.

Do you have any suggestions for any of these? Also, the styling of this might be a bit annoying, with the label elements and what not :/ (I left it out for now)

client/active-game/active-game-manager.js
```diff
@@ -274,6 +275,15 @@ async function doLaunch(gameId, serverPort, settings) {
     throw new Error(`StarCraft path ${starcraftPath} not valid: ` + JSON.stringify(checkResult))
   }
   const isRemastered = checkResult.remastered
+  if (isRemastered) {
+    try {
+      // Blizzard reinitializes their settings file everytime the SC:R is opened through their
+      // launcher. So we must do the same thing with our own version of settings before each game.
+      await ipcRenderer.invoke(SCR_SETTINGS_OVERWRITE)
+    } catch (err) {
+      throw new Error('Error overwriting SC:R settings: ' + err)
```
tec27, line 284: Feels like this should probably just throw the original error instead of discarding it in place of a toString'd version of it? (e.g. just remove the `try/catch` here and do the `await` right inside the if)

tec27 (approved):
Looks good to merge with these minor things fixed

client/settings/gameplay-settings.jsx
```diff
@@ -0,0 +1,123 @@
...
+        <SubmitOnEnter />
+        <FormContainer>
+          <div>
+            <Select {...bindCustom('unitPortraits')} label='Portraits' tabIndex={0}>
+              <Option value={2} text='Animated' />
+              <Option value={1} text='Still' />
+              <Option value={0} text='Disabled' />
+            </Select>
+            <Select {...bindCustom('minimapPosition')} label='Minimap position' tabIndex={0}>
+              <Option value={true} text='Left Corner' />
```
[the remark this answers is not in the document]
2Pacalypse-: Yeah, I copied this from the game. Changed it to Bottom-left corner.

PR #527, opened 2020-10-31
Add meta tags to HTML (Fixes #477)

### LINKED ISSUE #477: Add meta tags to HTML so that a proper thumbnail/description is given when the site is linked somewhere
There are various things that Discord/Twitter/Facebook look at, we should add some to the page so that it shows something reasonable (logo, description?) when a page on the site is linked. At some point in the future we could consider doing more for various types of linkable content (e.g. maps, replays).

GrantJenk:
I also added an image that is the recommended 1200x630 image size


tec27:
Thank you!

PR #529, opened 2020-11-03
Fix textfield highlighting in Firefox

GrantJenk:
I noticed it was possible to highlight text inside of text-fields in Chrome, but not in Firefox.
I used inherit for input-base so that it did not allow highlighting text in select components.


2Pacalypse-:
Hey, thanks for this!

PR #530, opened 2020-11-10
Add version number to app bar

### LINKED ISSUE #475: Add a version number visible somewhere in the client
Ideally all the time, so that we can easily identify versions in screenshots/on streams

GrantJenk:
Fixes #475 

shieldbattery

### DIFF
--- a/client/app-bar/lockup.jsx
+++ b/client/app-bar/lockup.jsx
@@ -3,8 +3,6 @@ import styled from 'styled-components'
 
 import Logo from '../logos/logo-no-bg.svg'
 import LogoText from '../logos/logotext-white-154x56.svg'
-import { Caption } from '../styles/typography.js'
-import { version as Version } from '../../package.json'
 
 const Container = styled.div`
   height: 100%;
@@ -12,12 +10,6 @@ const Container = styled.div`
   align-items: center;
 
   padding: 8px 0px 8px 8px;
-`
-
-const TitleContainer = styled.div`
-  margin-top: -10px;
-  display: grid;
-  justify-items: center;
 `
 
 const SizedLogo = styled(Logo)`
@@ -31,19 +23,12 @@ const SizedLogoText = styled(LogoText)`
   margin-left: 8px;
 `
 
-const VersionText = styled(Caption)`
-  margin: -15px 0px;
-`
-
 class Lockup extends React.Component {
   render() {
     return (
       <Container aria-label='ShieldBattery'>
         <SizedLogo />
-        <TitleContainer>
-          <SizedLogoText />
-          <VersionText>ver {Version}</VersionText>
-        </TitleContainer>
+        <SizedLogoText />
       </Container>
     )
   }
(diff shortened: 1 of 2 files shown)


client/app-bar/lockup.jsx
```diff
@@ -12,6 +14,12 @@ const Container = styled.div`
   padding: 8px 0px 8px 8px;
 `
 
+const TitleContainer = styled.div`
+  margin-top: -10px;
+  display: grid;
+  justify-items: center;
```
2Pacalypse-, line 20: This would probably look better if it's left-aligned, though not sure.

client/app-bar/lockup.jsx
```diff
@@ -23,15 +31,22 @@ const SizedLogoText = styled(LogoText)`
   margin-left: 8px;
 `
 
+const VersionText = styled(Caption)`
+  margin: -15px 0px;
```
2Pacalypse-, line 35: Per material design spec, try to stick to using a number divisible by 4 when using margins/paddings, unless there's a good reason to do otherwise. `-16px` would probably be a better choice here.

2Pacalypse- (changes requested):
I'll let tec make the final decision, but I'm not particularly sure about this placement. As far as I know, _Lockup_ is a particular entity in a design (Logo + LogoText), so I'm not sure what sticking a version in there does to that :d

Also, here's how the version looks like with some tweaks (mostly the stuff I wrote in the comments):
image

I don't really have a strong preference either way, just making sure we have some options.

P.S. The build is failing (probably due to some linter errors)

tec27:
I think this is definitely more prominent than I had intended. I'll try to mock something up today to give you a better idea of what I had in mind, so you don't have to guess at it.

tec27:
Alright, I think this would be a bit more appropriate:

image

image

It'll be a tad different in the actual version right now since we don't have those smaller actions implemented yet, but the basic idea is to just put it as the bottom item, `8px` from the action above it. The text is the Caption style (Roboto, 12px, 20px line height) at 70% `#ffffff`, centered.

GrantJenk:
I made a new commit that should better reflect tec's mockup.

The activity-bar had a padding of 8px at the bottom that I wasn't sure if you wanted left in or not. I decided to leave it in for now.

Preview:
version

client/main-layout.jsx
```diff
@@ -82,6 +85,11 @@ const StyledMapsIcon = styled(MapsIcon)`
   height: 36px;
 `
 
+const VersionText = styled(Caption)`
+  margin: 8px 0px 0px 0px;
+  opacity: 0.7;
```
tec27, line 90: Instead of using opacity directly, change the color. We have an exported variable called `colorTextSecondary` in `styles/colors.js` for this particular color

tec27 (changes requested):
The bottom padding looks fine for no. If it looks weird when we build the updated bottom actions, we can deal with it then. Thanks for making these changes!

PR #534, opened 2020-11-20
Add user popover menu in userlist

### LINKED ISSUE #269: Add popover menu to users in the chat list
For now this can just contain a whisper option? Eventually it would also expand to things like "View profile"

GrantJenk:
Fixes #269

I copied self-profile-overlay.jsx and only made slight changes for user-overlay.jsx
I also added a hover effect to the users in the userlist to make it more clear that it can be clicked. I think it looks a bit odd, not entirely sure how it should be styled.

Preview:
user-popover

### DIFF
--- a/client/chat/channel.jsx
+++ b/client/chat/channel.jsx
@@ -145,7 +145,7 @@ class UserList extends React.Component {
         <UserListSubheader as='p'>{title}</UserListSubheader>
         <UserSublist>
           {users.map(u => (
-            <UserListEntry user={u} key={u} onWhisperClick={this.props.onWhisperClick}/>
+            <UserListEntry user={u} key={u} onWhisperClick={this.props.onWhisperClick} />
           ))}
         </UserSublist>
       </UserListSection>
--- a/client/chat/channel.jsx
+++ b/client/chat/channel.jsx
@@ -103,7 +103,6 @@ class UserListEntry extends React.Component {
   }
 
   render() {
-    console.log(this.props)
     return (
       <>
         <UserListEntryItem ref={this._userEntryRef} onClick={this.onOpenUserOverlay}>
@@ -141,6 +140,7 @@ class UserList extends React.Component {
       return null
     }
 
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
     const { users, ...rest } = this.props
     return (
       <UserListSection>
@@ -227,6 +227,7 @@ class Channel extends React.Component {
   }
 
   render() {
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
     const { channel, onSendChatMessage, onRequestMoreHistory, ...rest } = this.props
     return (
       <Container>
--- a/client/chat/channel.jsx
+++ b/client/chat/channel.jsx
@@ -77,10 +77,10 @@ const UserListEntryItem = styled.li`
   }
 `
 
-@connect()
 class UserListEntry extends React.Component {
   static propTypes = {
     user: PropTypes.string.isRequired,
+    onWhisperClick: PropTypes.func,
   }
 
   state = {
@@ -97,8 +97,8 @@ class UserListEntry extends React.Component {
     this.setState({ userOverlayOpen: false })
   }
 
-  onWhisper = () => {
-    this.props.dispatch(navigateToWhisper(this.props.user))
+  onWhisperClick = () => {
+    this.props.onWhisperClick(this.props.user)
   }
 
   renderUserOverlay() {
@@ -108,7 +108,7 @@ class UserListEntry extends React.Component {
         onDismiss={this.onCloseUserOverlay}
         anchor={this._userEntryRef.current}
         user={this.props.user}>
-        <MenuItem text='Whisper' onClick={this.onWhisper} />
+        <MenuItem text='Whisper' onClick={this.onWhisperClick} />
       </UserOverlay>
     )
   }
@@ -128,6 +128,7 @@ class UserListEntry extends React.Component {
 class UserList extends React.Component {
   static propTypes = {
     users: PropTypes.object.isRequired,
+    onWhisperClick: PropTypes.func,
   }
 
   shouldComponentUpdate(nextProps) {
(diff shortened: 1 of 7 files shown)


client/chat/channel.jsx
```diff
@@ -68,15 +71,57 @@ const UserSublist = styled.ul`
...
+  }
+
   render() {
-    return <UserListEntryItem>{this.props.user}</UserListEntryItem>
+    return (
+      <>
+        <UserListEntryItem ref={this._userEntryRef} onClick={this.onOpenUserOverlay}>
+          {this.props.user}
+        </UserListEntryItem>
+        {this.renderUserOverlay()}
```
2Pacalypse-, line 122: I'm pretty sure we'll want to render the user overlay from the other parts of the application, and not just the chat channel. For example, from the whispers and lobbies. Assuming we'll want to render the _same_ user overlay in all of these places (and I can't see why not at the moment), then I think the logic that deals with rendering the user overlay should be moved up in hierarchy until it encompasses all of those places.

In our application, that's probably the `MainLayout` component. Keep in mind though, since the `MainLayout` is using `Route`s to render various stuff, that means you'll probably have to send certain props to the component that the `Route` is rendering. See `render` property in the react router documentation on how to do that easily: https://reactrouter.com/web/api/Route/render-func

client/chat/user-overlay.jsx
```diff
@@ -0,0 +1,161 @@
...
+    transition-timing-function: linear;
+  }
+
+  .exitActive & {
+    opacity: 0;
+    pointer-events: none;
+  }
+`
+
+export default class UserOverlay extends React.Component {
```
2Pacalypse-, line 102: Depending on the final design of the user overlay component, I feel like there's a lot of code here that could be reused from the self profile overlay component ^^.

Either by using a single component for both overlays (if the designs end up being _very_ similar), or pulling common parts in some generic component and using that component in both overlays.

2Pacalypse- (changes requested):
I'm pretty sure tec will have some feedback on the actual design of the user overlay (and probably even provide you with a mockup :p), but I left some general comments on code organization. Let me know if there's anything unclear or you need any help with.

client/chat/channel.jsx
2Pacalypse-: Actually, disregard this comment, it wasn't well thought through ^^.

client/material/overlay.jsx
```diff
@@ -0,0 +1,112 @@
+import React from 'react'
```
2Pacalypse-, line 1: Similar to above, I would move this file to the `profile` folder as well and perhaps rename it to `profile-overlay` or something.

Perhaps I misled you a bit when I said that this should be a general component. I meant that it should be a general component for the two profile overlays that we have now, not general for the whole application ^^. The popover component that we have in the material folder is already the general one that allows configuring transition timings and whatnot.

If we decide in the future that the transition timings that we're using here for the profile overlays want to also use in other places, unrelated to the profile overlays, then we can think about moving it to a more general place like material folder (which is what we're doing for the `Menu` component for example).

client/material/overlay.jsx
```diff
@@ -0,0 +1,112 @@
...
+      onDismiss,
+      anchor,
+      anchorOriginVertical,
+      anchorOriginHorizontal,
+      popoverOriginVertical,
+      popoverOriginHorizontal,
+    } = this.props
+
+    return (
+      <Popover
```
2Pacalypse-, line 69: When you have a component like this one that is only passing properties to the one that it is using, then you can just use the spread operator instead of specifying each prop individually, e.g.
```jsx
<Popover {...this.props} />
```
If you don't want to pass _all_ of the props to the underlying one, then you can do something like this:
```jsx
const { prop1, prop2, ...rest } = this.props

<Popover {...rest} />
```
So if a certain component doesn't use certain props, but just passes them down the chain, then using the spread operator makes it easier to update stuff later if a prop needs to be added/removed/renamed. Also, a component doesn't need to validate the prop types if it's not using those props, so you can remove that as well here.

Btw, this could also be applied in the above code too, like with passing the `onWhisperClick` handler down the chain. But I'm guilty of not doing this always myself :d, and I guess an argument could be made that it obfuscates things a bit.

2Pacalypse- (changes requested):
Besides the specific comments I left, I wanted to give you a slight update on why there's no feedback yet on the general design of the user profile. Tec is currently finishing the work on pieces needed to collect the game results, after which he'll design a mockup that shows user's win/loss statistic and/or matchmaking rank in the profile too, not sure yet :d.

client/material/overlay.jsx
2Pacalypse-: Now that I've actually seen this implemented, allow me to backtrack on this suggestion please ^^.  This is not the pattern that we should be using since it could very easily bite us in the ass (for example when new props are added, they could accidentally and unintentionally be passed down).

So there are two things I would change back here:
- revert the spreading of `onWhisperClick`. Just pass it down directly as it was before.
- make the `ProfileOverlay` component receive a prop called `popoverProps` explicitly (which is an object of all the Popover-specific properties) and then spread that instead, e.g in `profile-overlay.jsx`.
```jsx
const { children, popoverProps } = this.props

return (<Popover {...popoverProps}>...</Popover>)
```
Sorry about that :/

client/chat/channel.jsx
```diff
@@ -313,6 +362,10 @@ export default class ChatChannelView extends React.Component {
     this.props.dispatch(retrieveNextMessageHistory(this.props.match.params.channel))
   }
 
+  onWhisperClick(user) {
```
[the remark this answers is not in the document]
GrantJenk: Now that the bindings have been removed, can the constructor be deleted entirely now? Or should it be left there?

client/material/overlay.jsx
GrantJenk: No problem! I was actually wondering about that myself :P

client/chat/channel.jsx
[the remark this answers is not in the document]
2Pacalypse-: Yes, the constructor can be removed now.

tec27:
Sorry for taking so long to get to this! I think I'd like to see a sort of different approach where we split the informational popover from the action one (so the informational one displays on left click, actions like whisper appear on right click), *but* most of the information we want to display isn't in a state where you're actually going to be able to build that right now.

So in the interest of moving this PR forward, and since we have a redesign of the user list anyway, I'm happy to merge it as-is and I can clean up the design things. The only thing I'd like to see changed before doing that is moving from the `{...rest}` approach of prop forwarding to one where we explicitly pass a forwarded prop in (e.g. `{...popoverProps}` as 2Pac discussed above).

tec27:
Thank you!

PR #539, opened 2020-12-17
Send the game results to the server on game end.

tec27:
This sets things up so we can make this request from the Rust code, as the game finishes. We make 3 attempts, if all 3 fail, the expectation is that the app will retry at some point in the future (but this is NOT implemented yet as of this PR). The app can know to do this by the fact that the game exits without calling `/game/resultSent` (and my plan is to always submit some sort of results if that occurs, to try to give us fewer result timeouts in the case of weird issues).

APM calculation is also not included in this PR, for now everybody is registered as having 0 APM.

Note that there's a number of random Rust reformatting changes in here, as VSCode really wanted to do them. I adjusted our `.editorconfig` to preserve most of the existing formatting.

### DIFF
--- a/game/src/game_state.rs
+++ b/game/src/game_state.rs
@@ -785,15 +785,11 @@ impl InitInProgress {
                         GamePlayerResult {
                             result: results[player.storm_id.0 as usize] as u8,
                             race: match game_results.race[player_id as usize] {
-                                Some(bw::RACE_ZERG) => Race::Zerg,
-                                Some(bw::RACE_TERRAN) => Race::Terran,
-                                Some(bw::RACE_PROTOSS) => Race::Protoss,
+                                bw::RACE_ZERG => Race::Zerg,
+                                bw::RACE_TERRAN => Race::Terran,
+                                bw::RACE_PROTOSS => Race::Protoss,
                                 r => {
-                                    warn!(
-                                        "Invalid race ({}) for player {}",
-                                        r.map_or("None".to_string(), |v| v.to_string()),
-                                        player_id
-                                    );
+                                    warn!("Invalid race ({}) for player {}", r, player_id);
                                     Race::Zerg
                                 }
                             },
--- a/game/src/game_state.rs
+++ b/game/src/game_state.rs
@@ -1,14 +1,18 @@
+use crate::app_messages::GamePlayerResult;
+use crate::app_messages::GameResultsReport;
+use reqwest::header::HeaderMap;
+use reqwest::header::ORIGIN;
 use std::ffi::CStr;
 use std::mem;
 use std::net::Ipv4Addr;
 use std::path::{Path, PathBuf};
 use std::pin::Pin;
 use std::sync::Arc;
-use std::time::{Duration};
+use std::time::Duration;
 
 use bytes::Bytes;
-use futures::{pin_mut, select};
 use futures::prelude::*;
+use futures::{pin_mut, select};
 use quick_error::quick_error;
 use tokio::sync::{mpsc, oneshot};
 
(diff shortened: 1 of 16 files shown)


game/src/game_thread.rs
```diff
@@ -46,7 +46,7 @@ pub enum GameThreadMessage {
     Snp(snp::SnpMessage),
     /// Storm player id (which stays stable) -> game player id mapping.
     /// Once this message is sent, any game player ids used so far should be
-    /// considered invalid and updated to match this mapping.
+    /// considered invalid and updated to match this mappin
```
2Pacalypse-, line 49: This seems like an odd change :p

neivv, line 49: oops?

game/src/game_state.rs
```diff
@@ -336,17 +346,61 @@ impl GameState {
...
             game_done.await;
             let results = results.await?;
-            app_socket::send_message(&mut ws_send, "/game/end", results).await
+            app_socket::send_message(&mut ws_send, "/game/end", &results)
+                .await
                 .map_err(|_| GameInitError::Closed)?;
+
+            // Attempt to send results to the server, if this fails, we expect
+            // the app to retry in the future
+            let client = reqwest::Client::new();
```
neivv, line 363: Oh one more thing: reqwest seems to default for no timeout, maybe we should specify something here? Though I'd expect there's still some timeout value at TCP level?

game/src/game_state.rs
```diff
@@ -695,6 +748,17 @@ impl InitInProgress {
                         }
                     }
                 };
+                unsafe {
+                    with_bw(|bw| {
```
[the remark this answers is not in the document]
tec27: That was my desire as well and why this felt pretty gross to me, but for some reason when I looked at things last night I didn't find my way to the `game_results` function which seems like the right place for this. New implementation should be better (I *think*).

(The semantics of player IDs, or at least the values BW stores in that field is something I can't totally recall, but I think my code is handling that correctly now).

game/src/game_state.rs
```diff
@@ -336,17 +346,61 @@ impl GameState {
...
+            for _ in 0..3 {
+                let result = client
+                    .post(&result_url)
+                    .headers(result_headers.clone())
+                    .json(&result_body)
+                    .send()
+                    .await;
+
+                if result.is_ok() {
+                    if result.unwrap().status().is_success() {
```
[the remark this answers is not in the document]
tec27: Went with a `match` on `error_for_status()` and that seems a lot nicer :) Logs errors now as well

game/src/game_state.rs
```diff
@@ -1,14 +1,18 @@
+use crate::app_messages::GamePlayerResult;
```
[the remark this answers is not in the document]
tec27: This is what I get for letting VSCode import stuff as it wanted to 😠 Should be fixed now

game/src/game_state.rs
tec27: This is a great point, went with what seems to be typical for request timeouts in browsers (30 seconds)

game/src/game_thread.rs
```diff
@@ -106,6 +106,7 @@ pub enum PlayerLoseType {
 pub struct GameThreadResults {
     // Index by ingame player id
     pub victory_state: [u8; 8],
+    pub race: [Option<u8>; 8],
```
neivv, line 109: This can be just u8; and the `player.player_id` indirection below isn't needed as that field itself is redundant - `players[i].player_id == i` always.

(Or well, the field is only useful if code would take a pointer to a single struct in the array)

neivv (approved):
The player race collecting can be made simpler but otherwise looks good

game/src/game_thread.rs
tec27: Okay that's what I thought but I couldn't recall 100%. Adjusted to the simpler approach

PR #545, opened 2020-12-22
Implement a service which tracks matchmaking status.

### LINKED ISSUE #336: Implement ability to turn matchmaking on/off dynamically
When we initially release matchmaking, we're going to want the ability to limit it to specific times. This has a number of benefits for testing:

- If there are problems, their scope/effects are limited to a shorter time period
- Concentrates the population of people that want to try it into a smaller time period, increasing the population available during those times and making the matching experience faster
- Ensures that we can likely have better admin/help coverage during those periods to answer questions/assist with technical issues/diagnose bugs

## Proposed system

- Create a new permission `manage_matchmaking_times`
- Create a new table `matchmaking_times`:

```
id int (PK, autoincrement)
matchmaking_type enum (same as in other matchmaking tables)
start_date datetime
enabled bool
```

- API would prevent inserting start_dates < the current date
- Current state is retrieved by getting the latest start_date earlier than the current date for a particular matchmaking time.
- Find match UI would retrieve this state before allowing people to queue, and if not currently enabled, retrieve the *next* time matchmaking will be enabled to display to the user (in both absolute and relative form, ideally: `May 12th [1 week, 3 days, 2 hours, 41 minutes]`
- Ideally we'd display that state to users in some way on the activity button itself, so they don't get "teased" when it's not on, but this is a lower priority thing. This would probably requre that we update this status periodically, since admins can change these values. That or we subscribe all users to changes here.
- We may want to also think about notifying users when matchmaking turns on in some way, and perhaps also displaying how long they have until matchmaking turns off. Again though, lower priority than implementing the base system.
- Admin UI could display next N future changes + M previous changes

- It would seem nice to be able to specify recurring ranges (e.g. always enabled between the hours of 7PM Friday to 7AM Monday), but I think this is a lot more complex to query and deal with. Potential system could use something like the much-loved (ha) cron syntax. Generally I think this is not worthwhile, as this system is intended to be simple, and added even a full year's worth of weekends is completely doable admin busy work. Feels like we're better off optimizing for simple querying/client code than reducing work on the temporary system. It also sets us up better for utilizing this for a "maintenance" system that turns matchmaking off around times we want to perform maintenance/upgrades (where the status changes would be much less frequent)

2Pacalypse-:
And display an overlay if a user tries to use matchmaking while it's
disabled.

Fixes #336


server/lib/models/matchmaking-times.ts
```diff
@@ -154,6 +154,51 @@ export async function getPastMatchmakingTimes(
   }
 }
 
+/**
+ * Get matchmaking schedule. The schedule represents future N (where N defaults to 2) matchmaking
```
tec27, line 158: Nit: ideally the first sentence of a method comment wouldn't be exactly the same as the method name :)

`Retrieve the schedule of currently applicable matchmaking status changes.` would be a slight improvement

server/lib/models/matchmaking-times.ts
```diff
@@ -154,6 +154,51 @@ export async function getPastMatchmakingTimes(
...
+ * times, starting with a first S (where S defaults to enabled) status, after date X (where X
+ * defaults to current date) for a particular matchmaking type. Additionally, all subsequent
+ * matchmaking times with the same enabled status as the one before them are filtered out.
+ */
+export async function getMatchmakingSchedule(
+  matchmakingType: MatchmakingType,
+  date = new Date(),
+  firstStatus = true,
+  depth = 2,
```
tec27, line 167: Do we ever really need more than 2 here? Feel like making this configurable adds a ton of complexity to this query

server/websockets.js
```diff
@@ -46,6 +47,11 @@ class WebsocketServer {
 
     this.nydus.on('connection', socket => {
       this.nydus.subscribeClient(socket, '/status', { users: this.connectedUsers })
+
+      // TODO(2Pac): Only do this for Electron clients
+      if (matchmakingStatusInstance) {
```
tec27, line 52: The middleware for this exports a function to check this, so you should be able to do this in this PR. The request that started the connection is stored as `socket.conn.request`

tec27 (changes requested):

server/lib/api/matchmakingTimes.ts
```diff
@@ -165,6 +166,10 @@ async function addNew(ctx: Koa.Context, next: Koa.Next) {
   const { startDate, enabled } = ctx.request.body as AddMatchmakingTimeBody
 
   ctx.body = await addMatchmakingTime(matchmakingType, new Date(startDate), !!enabled)
+
+  if (matchmakingStatusInstance) {
+    matchmakingStatusInstance.maybePublish(matchmakingType)
```
tec27, line 171: Async function results need to be handled so that we don't end up with unhandled exceptions. Alternatively for here, since `maybePublish` is surrounded completely by a try/catch, you could expose a non-async version of that that calls through to a private `async` version but handles/logs errors from the `Promise` (you could also then remove the wrapping `try/catch` in that case since it would be redundant).

server/lib/models/matchmaking-times.ts
```diff
@@ -154,6 +154,51 @@ export async function getPastMatchmakingTimes(
...
+): Promise<Array<MatchmakingTime>> {
+  const query = sql`
+    WITH RECURSIVE schedule AS (
+      (SELECT *, 0 AS depth
+      FROM matchmaking_times AS mt
+      WHERE mt.matchmaking_type = ${matchmakingType} AND mt.enabled = ${firstStatus}
+        AND mt.start_date > ${date}
+      ORDER BY mt.start_date
```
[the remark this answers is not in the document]
2Pacalypse-: I'm not an expert in reading `EXPLAIN` output by any means, but from what I can tell, both using and not using `start_date DESC` results in the same query (which doesn't seem to use an index??):

image

image

server/lib/models/matchmaking-times.ts
2Pacalypse-: I don't think it adds that much complexity; it only adds an additional column that gets updated in the recursion. Unfortunately we can't really eliminate the recursion since each row needs to be compared with the one before it.

Plus, I can totally see the value in potentially displaying the longer matchmaking schedule to users in some way.

server/websockets.js
2Pacalypse-: Yes, but that middleware checks for `X-Shield-Battery-Client` header from what I can tell. And that header is not being set correctly for a while now on websocket connections ^^.

It's a whole thing, but basically we need to upgrade `nydus` to make it work. Either by waiting for `engine.io` to release a new version with my fix in (which has not been merged in yet), or by using the `attach()` method to construct a new server instance instead of using the constructor directly (which doesn't pass options correctly).

I _think_ I got this, but maybe I'm missing something?

server/lib/models/matchmaking-times.ts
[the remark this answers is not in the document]
tec27: yeah, `seq scan` = sequential scan = no index usage, but I think you would need a larger data set to actually check. 11 rows is basically always going to be a sequential scan for anything. But it's fine, I won't make you generate more data for this, I guess this query is going to be cached and not run most of the time anyway.

server/lib/matchmaking/matchmaking-status.ts
```diff
@@ -45,49 +45,54 @@ export default class MatchmakingStatus {
     }
   }
 
+  async _getStatus(type: MatchmakingType) {
```
tec27, line 48: You don't need to name these things this way any more, just make them actually `private`
```suggestion
  private async getStatus(type: MatchmakingType) {
```

client/matchmaking/find-match.jsx
```diff
@@ -186,7 +189,7 @@ class Find1vs1MatchForm extends React.Component {
             <BrowseButton onClick={onBrowsePreferred}>
               <RandomContainer>
                 <RandomIcon />
-                <SubheadingOld>Random map</SubheadingOld>
+                <Subtitle1>Random map</Subtitle1>
```
tec27, line 192: This should be `Subtitle2`. General rule is, things that are "ShieldBattery content" get put in Roboto Condensed:
users, maps, etc.

Refer to the change I made to matchmaking map pool screen and make this match

client/matchmaking/find-match.jsx
```diff
@@ -80,7 +80,8 @@ const Actions = styled.div`
   margin: 16px 24px;
 `
 
-const Overline = styled(SubheadingOld)`
+const Overline = styled.span`
```
tec27, line 83: We already have an Overline style, so this shouldn't inherit from `subtitle1`. If it doesn't match the other Overline style then it shouldn't be called that either

tec27 (approved):

client/matchmaking/find-match.jsx
```diff
@@ -81,7 +88,7 @@ const Actions = styled.div`
   margin: 16px 24px;
 `
 
-const Overline = styled.span`
+const SectionTtitle = styled.div`
```
tec27, line 91: typo?

PR #548, opened 2020-12-23
Remove GA.

### LINKED ISSUE #387: Make GA work
Ever since the move to react-router v4, GA doesn't work. Make it work again.

See this: https://github.com/react-ga/react-ga/wiki/React-Router-v4-withTracker

2Pacalypse-:
This commit removes the Google Analytics integration we had (though it
didn't even work since last react-router upgrade anyway). It's
questionable how useful GA would be to us, and not integrating it has a
nice bonus that we now don't need to add a cookie consent form.

Closes #387

### DIFF
--- a/client/app.jsx
+++ b/client/app.jsx
@@ -1,10 +1,7 @@
 import React from 'react'
 import { Route, Switch } from 'react-router-dom'
-import ga from 'react-ga'
 import { StyleSheetManager } from 'styled-components'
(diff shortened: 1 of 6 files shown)


2Pacalypse-:
Honestly I have no idea why the tests are failing on GitHub when I remove these two dependencies, they're passing fine locally. It might be due to some caching we're doing in GitHub actions, but not sure.

PR #550, opened 2020-12-23
Optimize map downloads in the matchmaking.

### LINKED ISSUE #483: Optimize map downloads in matchmaking
Currently we download a map on a client's computer during the game loading process, which can be further optimized. Since we won't know which exact map will be used in a match until after all the players accept the match, the easiest way to optimize this process would be to download _all_ of the maps as soon as a player enters the queue (in vast majority of cases, the map(s) will already exist on a player's computer, so this shouldn't be an expensive operation). During the game loading process, we'll still need to check if a map exists before starting a game, but that's a matter of just leaving the call to `downloadMap` function there, which does all of that, afaik.

2Pacalypse-:
Now the the whole map pool is downloaded as soon as the player enters
the queue.

Fixes #483

### DIFF
--- a/client/matchmaking/socket-handlers.js
+++ b/client/matchmaking/socket-handlers.js
@@ -2,6 +2,7 @@ import activeGameManager from '../active-game/active-game-manager-instance'
 import rallyPointManager from '../network/rally-point-manager-instance'
 import mapStore from '../maps/map-store-instance'
 import audioManager, { SOUNDS } from '../audio/audio-manager-instance'
+import log from '../logging/logger'
 import {
   ACTIVE_GAME_LAUNCH,
   MATCHMAKING_FIND,
@@ -164,7 +165,9 @@ const eventToAction = {
     } = event.chosenMap
     // Even though we're downloading the whole map pool as soon as the player enters the queue,
     // we're still leaving this as a check to make sure the map exists before starting a game.
-    mapStore.downloadMap(hash, format, mapUrl)
+    mapStore
+      .downloadMap(hash, format, mapUrl)
+      .catch(err => log.error('Error while downloading map: ' + err))
 
     const config = {
       localUser: user,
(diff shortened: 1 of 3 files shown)


client/matchmaking/socket-handlers.js
```diff
@@ -162,8 +162,8 @@ const eventToAction = {
       mapData: { format },
       mapUrl,
     } = event.chosenMap
-    // NOTE(2Pac): We can't download map any sooner since we don't know which map we'll play until
-    // all players accept the game
+    // Even though we're downloading the whole map pool as soon as the player enters the queue,
+    // we're still leaving this as a check to make sure the map exists before starting a game.
     mapStore.downloadMap(hash, format, mapUrl)
```
neivv, line 167: Half-unrelated to this PR, but shouldn't this have been made to await on the function to finish before continuing the launch?

2Pacalypse-: Yeah, I was thinking about that today myself.

We're not waiting for it in the lobbies and I think I just copy-pasted it from there when I initially added it here. But it's much more likely to cause issues here than downloading a map when someone joins a lobby ^^.

Not sure if this is the best place to await stuff for though. Maybe the active-game-manager should act as a fail-safe in case the map was not downloaded yet from these other places and wait for it to finish before launching the game?

tec27: Yeah, this should ideally get passed into that workflow in some way so that the game doesn't get told to set up until the map download is finished. That can go into a separate PR though, but while you're here can you attach an error handler to this Promise and log failures?

tec27:
You can merge once you add the error logging

client/matchmaking/socket-handlers.js
2Pacalypse-: Should we do this in all the places where we call `downloadMap`˙?

tec27: Ideally, yeah

PR #556, opened 2020-12-24
Update the design of the searching match overlay to match the latest

2Pacalypse-:
mockup.

### DIFF
--- a/client/actions.ts
+++ b/client/actions.ts
@@ -321,6 +321,8 @@ export const MATCHMAKING_UPDATE_MATCH_ACCEPTED = 'MATCHMAKING_UPDATE_MATCH_ACCEP
 export const MATCHMAKING_UPDATE_MATCH_FOUND = 'MATCHMAKING_UPDATE_MATCH_FOUND'
 // The server has responded that all players have accepted the match and game is ready to start
 export const MATCHMAKING_UPDATE_MATCH_READY = 'MATCHMAKING_UPDATE_MATCH_READY'
+// A second has elapsed while searching for a match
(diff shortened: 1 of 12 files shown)


client/actions.ts
```diff
@@ -321,6 +321,8 @@ export const MATCHMAKING_UPDATE_MATCH_ACCEPTED = 'MATCHMAKING_UPDATE_MATCH_ACCEP
 export const MATCHMAKING_UPDATE_MATCH_FOUND = 'MATCHMAKING_UPDATE_MATCH_FOUND'
 // The server has responded that all players have accepted the match and game is ready to start
 export const MATCHMAKING_UPDATE_MATCH_READY = 'MATCHMAKING_UPDATE_MATCH_READY'
+// A second has elapsed while searching for a match
+export const MATCHMAKING_UPDATE_SEARCHING_TIME = 'MATCHMAKING_UPDATE_SEARCHING_TIME'
```
tec27, line 325: This feels *extremely* heavy... Do we really want to run every reducer and potentially re-render the entire app every second? We managed to do this without actions for the left-nav version, I'm not sure why the addition of another display of this info requires a change to that approach

tec27 (changes requested):

client/actions.ts
tec27: My recommendation would be to store the current value of `window.performance.now()` in the store when the search starts, and set an interval in each display component to update based on the delta each second. There are some other things you could do to mitigate them not matching 100% of the time but I mean, it's not really a big deal.

PR #559, opened 2020-12-24
Move chat componentWillReceiveProps/componentWillUpdate -> componentD…

2Pacalypse-:
…idUpdate.

### DIFF
--- a/client/chat/channel.jsx
+++ b/client/chat/channel.jsx
@@ -209,10 +209,10 @@ const MessagesAndInput = styled.div`
 `
 
 const Messages = styled.div`
-  height: calc(100% - 56px - 16px); /* chat input height + margin */h
+  height: calc(100% - 56px - 16px); /* chat input height + margin */
   margin: 0 0 -1px 0;
-  border-bottom: 1px solid ${props =>
-    props.isScrolledUp ? colorDividers : 'rgba(255, 255, 255, 0)'};
+  border-bottom: 1px solid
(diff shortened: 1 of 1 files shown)


client/chat/channel.jsx
```diff
@@ -237,12 +237,12 @@ class Channel extends React.Component {
...
     const insertingAtTop =
-      nextProps.channel !== this.props.channel &&
-      this.props.channel.messages.size > 0 &&
-      nextProps.channel.messages.size > this.props.channel.messages.size &&
-      nextProps.channel.messages.get(0) !== this.props.channel.messages.get(0)
+      prevProps.channel !== this.props.channel &&
+      prevProps.channel.messages.size > 0 &&
+      this.props.channel.messages.size > prevProps.channel.messages.size &&
+      this.props.channel.messages.get(0) !== prevProps.channel.messages.get(0)
     this.messageList.setInsertingAtTop(insertingAtTop)
```
tec27, line 246: I thnk `componentDidUpdate` is too late to do this. In any case, this PR should contain both ScrollableContent changes *and* these, because the right place to do this depends on both changes.

2Pacalypse-: I tried using `getSnapshotBeforeUpdate()` instead and I couldn't make it work. It could probably be done, but I tested it as much as I could with `componentDidUpdate()` and everything works as far as I can see.

2Pacalypse-:
Closed in favour of #568

PR #560, opened 2020-12-24
Move whispers componentWillReceiveProps/componentWillUpdate -> compon…

2Pacalypse-:
…entDidUpdate.


tec27:
Should also be merged into the chat one (and whatever the right fix is for that should be applied here too)

2Pacalypse-:
Closed in favour of #568.

PR #566, opened 2020-12-24
Add user highlight when overlay is open

GrantJenk:
No idea if this is actually desired, no biggie if not.


client/chat/channel.jsx
```diff
@@ -81,6 +81,13 @@ const UserListEntryItem = styled.li`
     cursor: pointer;
     background-color: rgba(255, 255, 255, 0.08);
   }
+
+  ${props => {
+    if (props.isOverlayOpen) {
+      return `background-color: ${grey800};`
```
tec27, line 87: I think ideally this would be the same color as the hover state, instead of a unique value

tec27 (changes requested):

PR #568, opened 2020-12-25
Update scroll-bar (and things using it) to not use deprecated APIs.

2Pacalypse-:


client/chat/channel.jsx
```diff
@@ -247,12 +247,12 @@ class Channel extends React.Component {
...
     const insertingAtTop =
-      nextProps.channel !== this.props.channel &&
-      this.props.channel.messages.size > 0 &&
-      nextProps.channel.messages.size > this.props.channel.messages.size &&
-      nextProps.channel.messages.get(0) !== this.props.channel.messages.get(0)
+      prevProps.channel !== this.props.channel &&
+      prevProps.channel.messages.size > 0 &&
+      this.props.channel.messages.size > prevProps.channel.messages.size &&
+      this.props.channel.messages.get(0) !== prevProps.channel.messages.get(0)
     this.messageList.setInsertingAtTop(insertingAtTop)
```
tec27, line 256: Since this is used in `getSnapshotBeforeUpdate` in `ScrollableContent`, doesn't it need to be there in this component as well?

(Same for all the other places that are doing this)

2Pacalypse-: As I said in the other PR, I tried putting this in `getSnapshotBeforeUpdate()` and couldn't make it work. It could probably be done with some more tinkering, but I tested it as much as I could with `componentDidUpdate()` and everything seems to work as far as I can see.

tec27: Oh sorry, I didn't look through the PRs that got closed.

The thing this would have an effect on is when you're scrolling up and new messages are being loaded in. If it was broken I think the scroll position wouldn't get adjusted back down to the place you were at prior to loading. If that seems to work okay then you can go ahead and merge this

2Pacalypse-: You're right, that doesn't work :/
I'm not even sure how to fix it to be honest. Not that familiar with `getSnapshotBeforeUpdate()` to see why it didn't work when I tried it with that.

2Pacalypse-: Actually, to be more specific, when I put this in `getSnapshotBeforeUpdate()` that use-case of scrolling up and loading new messages works. But the auto-scrolling to the bottom when entering new message stops working.

PR #581, opened 2021-01-05
Extend new replays with extra data

neivv:
This adds support for expanding SC:R replays with custom data.

The SC:R replay format already has support for custom sections of data that can be skipped over by users that don't understand the section, so this will just append one more section at the end when the replay is being saved.
Also, as such, 1.16.1 replays don't change at all, maybe it can be patched to support SC:R replays in general if needed? Eventually.

The currently saved data & its format is very simple, uncompressed binary format, contains SC:R patch, shieldbattery version(*), and few bytes that fix team melee replays.
And a version id for future changes. Maybe there would be value in having some "better" format that is trivial to extend without breaking backwards compatibility for future before merging this PR? Though I'm personally not too concerned about it, we probably don't need to change the format too often in the end?

(*) Shieldbattery version is read from package.json when building the dll. Does the new version publishing procedure guarantee that package.json is updated before the game dll is built?

Team melee replays seemed to be pretty simple to fix, and at least a short test played back correctly, but I wouldn't be suprised if there were more bugs with them I missed.

### DIFF
--- a/client/active-game/game-config.ts
+++ b/client/active-game/game-config.ts
@@ -27,6 +27,8 @@ export interface PlayerInfo {
     | 'closed'
   /** The BW id of the type of this slot. */
   typeId: number
+  /** Shieldbattery user ID of the player. Only set for 'human' and 'observer' */
+  userId?: number
 }
 
 export interface ReplayMapInfo {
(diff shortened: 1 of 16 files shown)


neivv:
Hmm okay team melee replays don't quite work yet. Don't merge before I get to fix them :p

neivv:
Okay now the replays actually work

PR #584, opened 2021-01-12
Add a manual integration of redux devtools for Electron clients.

2Pacalypse-:
Usage:
  - ctrl+h to show/hide the tools
  - ctrl+q to change position of the tools

These were the default hotkeys as specified in the walkthrough guide.
Not sure if we want to change them though?

Currently we're using the official <DockMonitor /> compnent to position
the tools, which overlays the tools over the application on the same
page. Eventually we might want to invest some time to make it display
similarly to the chrome extensions, e.g. in the same window, but not
overlaying the application, or in a popup window.

Another thing that might be useful to implement at some point is to add
a button or something which opens/closes the redux devtools, instead of
relying on hotkeys alone.


client/index.jsx
```diff
@@ -16,6 +16,13 @@ import fetch from './network/fetch'
 import audioManager from './audio/audio-manager-instance'
 import { AUDIO_MANAGER_INITIALIZED } from './actions'
 
+const isDev = (__WEBPACK_ENV.NODE_ENV || 'development') === 'development'
```
tec27, line 19: the proper check here is really that it `!== 'production'`

tec27 (approved):

PR #586, opened 2021-01-20
Don't open a new page when resending a verification email.

### LINKED ISSUE #585: Resend verification email link should be an action, rather than an about:blank link
It being `about:blank` means it gets opened in the user's browser if they click it in the electron app, which may or may not be logged in. Pretty non-ideal experience.

Instead we should handle the click and `fetch` the right URL, probably show the result in a snackbar.

2Pacalypse-:
Instead, show the success/failure of the action in a snackbar.

Fixes #585

### DIFF
--- a/client/auth/action-creators.ts
+++ b/client/auth/action-creators.ts
@@ -133,7 +133,7 @@ export function verifyEmail(token: string) {
   return idRequest('@auth/verifyEmail', () => fetch<void>(url, { method: 'post' }))
 }
 
-export function sendVerificationEmail() {
+export function sendVerificationEmail(): ThunkAction {
   return dispatch =>
     fetch<void>('/api/1/users/sendVerification', { method: 'post' }).then(
       () =>
--- a/client/auth/action-creators.ts
+++ b/client/auth/action-creators.ts
@@ -18,7 +18,7 @@ function idRequest<
   ActionType extends Extract<IdRequestable, { type: ActionTypeName }>
 >(
   type: ActionTypeName,
-  fetcher: (dispatch: any) => Promise<ActionType['payload']>,
+  fetcher: () => Promise<ActionType['payload']>,
 ): {
   id: string
   action: ThunkAction<ActionType | AuthChangeBegin>
@@ -35,7 +35,7 @@ function idRequest<
         },
       })
 
-      const payload = fetcher(dispatch)
+      const payload = fetcher()
       const promisified: PromisifiedAction<ActionType> = ({
         type,
         payload,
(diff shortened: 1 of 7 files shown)


client/auth/action-creators.ts
```diff
@@ -17,7 +18,7 @@ function idRequest<
   ActionType extends Extract<IdRequestable, { type: ActionTypeName }>
 >(
   type: ActionTypeName,
-  fetcher: () => Promise<ActionType['payload']>,
+  fetcher: (dispatch: any) => Promise<ActionType['payload']>,
```
2Pacalypse-, line 21: Could use some help with typing the `dispatch` function here.

tec27: sendVerificationEmail isn't an ID'd request (it's not associated with a form, we'd never show the error from the reducer, so why do this?), so just don't add this and don't use this function for it and you won't have this problem.

For future reference though, `dispatch-registry` exports a typing for the dispatch function.

client/auth/action-creators.ts
```diff
@@ -135,7 +136,26 @@ export function verifyEmail(token: string) {
 export function sendVerificationEmail() {
   const url = '/api/1/users/sendVerification'
 
-  return idRequest('@auth/sendVerificationEmail', () => fetch<void>(url, { method: 'post' }))
+  return idRequest('@auth/sendVerificationEmail', dispatch =>
+    fetch<void>(url, { method: 'post' }).then(
+      () =>
+        dispatch(
+          openSnackbar({
+            message: 'Verification email has successfully been sent. Check your email.',
```
tec27, line 144: Remove the second sentence here

client/auth/action-creators.ts
```diff
@@ -135,7 +136,26 @@ export function verifyEmail(token: string) {
...
+            time: TIMING_LONG,
+          }),
+        ),
+      err => {
+        dispatch(
+          openSnackbar({
+            message: 'Something went wrong while trying to send a verification email.',
```
tec27, line 151: Replace with `Something went wrong while sending a verification email, please try again later.`

client/auth/action-creators.ts
tec27: Probably also `Verification email has been sent successfully.`

client/auth/action-creators.ts
```diff
@@ -132,10 +133,27 @@ export function verifyEmail(token: string) {
...
+        ),
+      err => {
+        dispatch(
+          openSnackbar({
+            message:
+              'Something went wrong while sending a verification email, please try again later.',
+            time: TIMING_LONG,
+          }),
+        )
+        throw err
```
tec27, line 154: Probably don't wanna throw this, it'll count as an uncaught error (and you *did* handle it here)

tec27 (approved):

PR #587, opened 2021-01-25
Change opacity of offline users

GrantJenk:


client/chat/channel.jsx
```diff
@@ -71,6 +71,18 @@ const UserSublist = styled.ul`
   list-style-type: none;
   margin: 0;
   padding: 0;
+
+  ${props => {
+    if (props.sectionTitle === 'Offline') {
```
tec27, line 76: This *works* but it's pretty hacky imo. At the very least I'd move the string comparison to the JS function and pass in a bool for `faded` or something here

client/chat/channel.jsx
```diff
@@ -71,6 +71,18 @@ const UserSublist = styled.ul`
...
+      return `
+        color: ${colorTextFaint};
```
tec27, line 78: `colorTextFaint` should be reserved for things that don't need to be readable as it doesn't generally meet contrast guidelines (basically, only the text of disabled things). The lowest you should go for stuff that does need to be readable is `colorTextSecondary`

client/chat/channel.jsx
```diff
@@ -71,6 +71,18 @@ const UserSublist = styled.ul`
...
+        svg {
```
tec27, line 79: This selector is definitely more general than it needs to be and might cause problems in the future. Prefer a more direct component reference, e.g.
```
${StyledAvatar} {
  opacity: ...
}
```

tec27 (changes requested):

client/chat/channel.jsx
GrantJenk: Would you prefer the opacity here to be changed to `alphaSecondary` so that it matches the color change to `colorTextSecondary`?

tec27: I don't have too much of a preference, generally icons can withstand a lower opacity and still be visible, so you just want it to look visually "the same" (which, counterintuitively, often means icons end up with a lower opacity than text). I haven't looked to see what feels like it matches in this particular case, but I feel like the lower opacity here is probably okay.

tec27:
Thank you!

PR #588, opened 2021-02-07
Add the beginnings of a party system.

2Pacalypse-:
This PR adds the server-side party service (and API handlers) for our
party system. Currently, only the following three actions are
implemented:
  - Send an invite to someone
  - Decline an invite (or cancel an invite you sent)
  - Accept an invite

Not sure how many more actions will be added to this PR, before we
decide it's ok to merge it. But I'm guessing at least the "Leave the
party" action as well.

NOTE: Some things will most likely change since we still haven't defined
a clear flowchart for all the possible scenarios, nor did we settle on
a UI/UX yet.


server/lib/api/parties.ts
```diff
@@ -0,0 +1,163 @@
...
+        },
+      ),
+    )
+
+    const leader = new PartyUser({
+      id: ctx.session!.userId,
+      name: ctx.session!.userName,
+    })
+
+    partyService.invite(leader, clientId, invites as PartyUser[])
```
tec27, line 104: if you add a generic parameter to `Promise.all` this cast will be unnecessary

server/lib/api/parties.ts
```diff
@@ -0,0 +1,163 @@
...
+
+    ctx.status = 204
+  } catch (err) {
+    if (isPartyServiceError(err)) {
+      throw new httpErrors[err.code](err.message)
```
tec27, line 109: So if you're going this route, you may as well just throw http-errors from the PartyService, that's not really what I was suggesting. What I meant is that the PartyService could throw error codes specific to it (its own enum of possible error conditions) and you'd have a function here that would convert a PartyServiceError code into the relevant http one

This would allow the PartyService to exist separate from whatever transport it ends up being on, and the tests for it to make sense without any understanding of HTTP

server/lib/parties/party-service.ts
```diff
@@ -0,0 +1,217 @@
...
+      user: client.userId,
+    })
+
+    // TODO(2Pac): Handle party leader leaving
+
+    // TODO(2Pac): Handle last person in a party leaving
+  }
+
+  private publishToParty(partyId: string, data: any) {
+    this.nydus.publish(getPartyPath(partyId as string), data)
```
tec27, line 192: Not really sure what this cast is here for, this is already the type of this parameter

server/lib/parties/party-service.ts
```diff
@@ -0,0 +1,217 @@
...
+}) {}
+
+export class PartyRecord extends Record({
+  id: null as string | null,
+  invites: Map<number | null, PartyUser>(),
+  members: Map<number | null, PartyUser>(),
+  leader: null as number | null,
+}) {}
+
+enum ErrorEnum {
```
tec27, line 26: I'd avoid naming enums with `Enum`. `PartyServiceErrorCode` or `PartyServiceErrorType` is probably better

server/lib/parties/party-service.ts
```diff
@@ -0,0 +1,217 @@
...
+  Unauthorized = 'Unauthorized',
```
tec27, line 27: If/when you move away from directly mirroring http-error codes, I'd probably avoid a String enum (just let typescript assign number values)

tec27 (commented):
Would definitely recommend trying to write a test sooner rather than later

server/lib/parties/party-service.ts
```diff
@@ -0,0 +1,217 @@
...
+    // TODO(2Pac): Handle party leader leaving
+
+    // TODO(2Pac): Handle last person in a party leaving
+  }
+
+  private publishToParty(partyId: string, data: any) {
+    this.nydus.publish(getPartyPath(partyId), data)
+  }
+
+  private getUser(username: string): UserSocketsGroup {
```
tec27, line 195: nit: I'd probably move to `getUserGroup` and `getClientGroup` (and use those names for variables as well) for this stuff, `user` is especially confusing since now it can mean:

 - a `User` object
 - a username
 - a user ID
 - a PartyUser
 - a UserSocketsGroup
 
If we can get that down to just one option (which in this context would be a PartyUser?) I think that'd be ideal :)

2Pacalypse-: What do you think about `getUserSockets` and `getClientSockets`? "Group" sounds ambiguous too imo, especially in a _party_ context :d.

But on the bright side, we now have TypeScript which makes stuff like this much less of a problem!

tec27: That was my initial suggestion, actually, but it overlaps with what you're injecting the `UserSocketsManager` and `ClientSocketsManager` as :)

If you want to change that, that's fine.

And yeah, I agree it's nice that TS will prevent actual misuses, but it does make the code a bit harder to review at a glance, which is my main reason for wanting the rename

test/setup.ts
```diff
@@ -0,0 +1,50 @@
+import 'core-js/proposals/reflect-metadata'
```
tec27, line 1: So I would probably recommend now taking this approach for unit tests, I think it makes you have to setup a sort of a parallel universe to all the app things eventually, which isn't ideal for small tests. One of the nice things about our injection setup is that we have regular constructor params to use, so you could actually just call them directly.

I'd probably pull these fakes into another file somewhere (maybe a `test/fakes/` folder or something?) and not setup the injection at all, instead constructing them manually.

The goal here is to fake as few things as possible. I definitely think you'll need to fake out `NydusServer` since it has so many external deps, but I think `UserSocketsManager`/`ClientSocketsManager` are probably fine with no faking, provided you integrate a fake session lookup.

test/setup.ts
```diff
@@ -0,0 +1,50 @@
...
+import { UpdateOrInsertUserIp } from '../server/lib/network/user-ips-type'
+import { RequestSessionLookup, SessionInfo } from '../server/lib/websockets/session-lookup'
+import {
+  ClientSocketsGroup,
+  ClientSocketsManager,
+  UserSocketsGroup,
+  UserSocketsManager,
+} from '../server/lib/websockets/socket-groups'
+
+export class FakeNydusServer extends NydusServer {
```
tec27, line 13: I think I'd probably fake this whole thing, stubbing out particular methods only seems very dangerous for updates. I *think* you can do something like:

```ts
class FakeNydusServer extends EventEmitter implements NydusServer {
  // ...
}
```

to get the EventEmitter methods implemented for you (those are fine to leave unfaked imo) and then either include empty methods for the rest of the API, or implement things to help with testing (logging events to lists to verify later, etc.)

test/setup.ts
```diff
@@ -0,0 +1,50 @@
...
+  publish() {
+    console.log('FakeNydusServer - publish')
+  }
+}
+
+export class FakeUserSocketsManager extends UserSocketsManager {
```
tec27, line 19: Ideally these don't need to be faked, but the goal here should probably be that the state of all these things (NydusServer, UserSocketsManager, ClientSocketsManager) aligns to the same thing without manual effort. That's why I'd recommend faking the session lookup instead. May help to make a utility method somewhere to create/connect fake clients with a particular username/client ID.

server/lib/parties/party-service.test.ts
```diff
@@ -0,0 +1,264 @@
...
+
+  beforeEach(() => {
+    nydus = createFakeNydusServer()
+    const sessionLookup = new RequestSessionLookup()
+    const userSocketsManager = new UserSocketsManager(nydus, sessionLookup, async () => {})
+    const clientSocketsManager = new ClientSocketsManager(nydus, sessionLookup)
+    partyService = new PartyService(nydus, clientSocketsManager, userSocketsManager)
+    connector = new NydusConnector(nydus, sessionLookup)
+
+    client1 = connector.connectClient(user1, USER1_CLIENT_ID)
```
tec27, line 55: Wonder if it might be cleaner to just make an array of these things and reference them that way :)

server/lib/parties/party-service.test.ts
```diff
@@ -0,0 +1,264 @@
...
+      beforeEach(() => {
+        leader = user1
+        party = partyService.invite(leader, USER1_CLIENT_ID, [user2])
+        partyService.acceptInvite(party.id, user2, USER2_CLIENT_ID)
+      })
+
+      test('should throw if invited by non-leader', () => {
+        const result = () => partyService.invite(user2, USER2_CLIENT_ID, [user3])
+
+        expect(result).toThrow()
```
tec27, line 88: Probably want to be more specific about what is being thrown here, since we actually have specific errors for it? (Can't recall what matching capabilities Jest has but I *think* you should be able to look for an error with a particular code here)

server/lib/parties/party-service.test.ts
```diff
@@ -0,0 +1,264 @@
...
+
+    describe('when party exists', () => {
+      beforeEach(() => {
+        leader = user1
+        party = partyService.invite(leader, USER1_CLIENT_ID, [user2])
+        partyService.acceptInvite(party.id, user2, USER2_CLIENT_ID)
+      })
+
+      test('should throw if invited by non-leader', () => {
+        const result = () => partyService.invite(user2, USER2_CLIENT_ID, [user3])
```
tec27, line 86: nit: I'd probably just inline this, but also this variable name isn't really accurate to what this is

server/lib/parties/party-service.test.ts
```diff
@@ -0,0 +1,264 @@
...
+
+        expect(nydus.publish).toHaveBeenCalledWith(getPartyPath(party.id), {
+          type: 'invite',
+          invites: [user3],
+        })
+      })
+    })
+
+    describe("when party doesn't exist", () => {
+      beforeEach(() => {
```
tec27, line 108: Not a huge fan of putting these things in a `beforeEach` here. I get that the act of creating a party is common among these tests, but this is actually the code under test for these cases and putting it here obscures that fact. Better to put this in each test case

server/lib/parties/party-service.test.ts
```diff
@@ -0,0 +1,264 @@
...
+    let leader: PartyUser
+    let party: PartyRecord
+
+    beforeEach(() => {
+      leader = user1
+      party = partyService.invite(leader, USER1_CLIENT_ID, [user2, user3])
+    })
+
+    test('should throw if the party is not found', () => {
+      const result = () => partyService.removeInvite('INVALID_PARTY_ID', user2)
```
tec27, line 158: Same as above, naming isn't really accurate and I'd probably just inline. Apply throughout these tests

server/lib/parties/party-service.test.ts
```diff
@@ -0,0 +1,264 @@
...
+
+      expect(result).toThrow()
```
tec27, line 160: Also same as above, if we have a particular type of error we expect based on the API, try to match on that when using this

server/lib/parties/party-service.ts
```diff
@@ -0,0 +1,228 @@
...
+    }
+
+    if (leader && leader.id !== party.leaderId) {
+      throw new PartyServiceError(
+        PartyServiceErrorCode.InsufficientPermissions,
+        'Only party leaders can remove invites to other people',
+      )
+    }
+
+    // TODO(2Pac): Check if the target is invited first before trying to do any of the below?
```
tec27, line 125: `delete` returns true/false to say whether anything was actually removed, so you can just use that + throw if it's false?

server/lib/parties/party-service.ts
```diff
@@ -0,0 +1,228 @@
...
+
+    party.invites.delete(target.id)
+    this.publishToParty(partyId, {
+      type: 'decline',
+      target,
+    })
+
+    // TODO(2Pac): Do we want to remove the party record if there's only one member left in a party
```
tec27, line 133: I'd say probably, yes, although it depends on how much the UI changes between being in a party/not being in a party

server/lib/parties/party-service.ts
```diff
@@ -0,0 +1,228 @@
...
+    const party = this.parties.get(partyId)
+    if (!party) {
+      throw new PartyServiceError(PartyServiceErrorCode.PartyNotFound, 'Party not found')
+    }
+
+    if (party.members.size >= MAX_PARTY_SIZE) {
+      throw new PartyServiceError(PartyServiceErrorCode.PartyFull, 'Party is full')
+    }
+
+    // TODO(2Pac): Check if the target is invited first before trying to do any of the below?
```
tec27, line 152: yes, obv :) don't want to let people join themselves into random other parties they weren't invited to

server/lib/parties/party-service.ts
```diff
@@ -0,0 +1,228 @@
...
+
+    const oldParty = this.getClientParty(clientSockets)
```
tec27, line 154: probably need to update this name now that this isn't immutable

server/lib/parties/party-service.test.ts
2Pacalypse-: Eh, I thought it would be cleaner too, but after having done it I realized it wasn't really :d. Immediate problem I noticed was that now clients are referenced from 0, instead of 1, due to zero-based indexing nature of arrays. This makes it confusing to read when using, for example, `user1` and `client0` in the same test. Then I tried renaming all users to be 0-based as well, but that makes it more annoying when you have to test how many users there are in a party since you have to add 1 to the user index to know how many of them are there. And _then_ I tried adding a dummy client to the array at the index 0, so all other clients could be easily linked to their 1-based users counterpart, and as I was writing a comment explaining why this dummy user is in the array, I realized that this isn't really that cleaner than what it was before :d

server/lib/parties/party-service.ts
2Pacalypse-: Yeah, initially I did write the code to remove the party record, but then I thought if that results in party chat being destroyed, then the person who invited someone won't even see the message that someone declined. Gonna leave this as TODO for now, until we settle on a UI.

server/lib/api/parties.ts
```diff
@@ -0,0 +1,182 @@
...
+
+  switch (err.code) {
+    case PartyServiceErrorCode.PartyNotFound:
+      throw new httpErrors.NotFound(err.message)
+    case PartyServiceErrorCode.InsufficientPermissions:
+      throw new httpErrors.Forbidden(err.message)
+    case PartyServiceErrorCode.PartyFull:
+      throw new httpErrors.Conflict(err.message)
+    case PartyServiceErrorCode.UserOffline:
+      throw new httpErrors.Unauthorized(err.message)
```
tec27, line 102: Mentioned this in the service file as well, but `Unauthorized` is really about the *current* user, not *other* users.

`Not Found`, `Conflict`, or `Gone` are probably more applicable status codes

server/lib/api/parties.ts
```diff
@@ -0,0 +1,182 @@
...
+        async (target: string): Promise<PartyUser> => {
+          const foundTarget = await users.find(target)
+          if (!foundTarget) {
+            throw new httpErrors.NotFound('Target user not found')
+          }
+
+          // TODO(2Pac): Check if the target user has blocked invitations from the user issuing
+          // the request. Or potentially use friends list when implemented.
+
+          return { id: foundTarget.id, name: foundTarget.name } as PartyUser
```
tec27, line 125: This cast is unnecessary

server/lib/api/parties.ts
```diff
@@ -0,0 +1,182 @@
...
+    case PartyServiceErrorCode.InsufficientPermissions:
+      throw new httpErrors.Forbidden(err.message)
+    case PartyServiceErrorCode.PartyFull:
+      throw new httpErrors.Conflict(err.message)
+    case PartyServiceErrorCode.UserOffline:
+      throw new httpErrors.Unauthorized(err.message)
+    case PartyServiceErrorCode.InvalidAction:
+      throw new httpErrors.BadRequest(err.message)
+    default:
+      assertUnreachable(err.code as never)
```
tec27, line 106: Don't do this cast. At best this is casting something that already has the type `never`, at worst it makes this assertion not work.

`assertUnreachable` works because TypeScript is smart enough to look at the cases of the switch and realize that when all possible values are exhausted by case statements, the type of the switch'd value is non-existent (e.g. `never`). If you call `assertUnreachable` with a value when that value *hasn't* been auto-narrowed to `never`, the compilation will fail. But that is *the entire point*: it's meant to tell you if someone adds a value to the enum without also updating this switch.

So, never cast the value you pass into it

server/lib/api/parties.ts
```diff
@@ -0,0 +1,182 @@
...
+    // If the `targetId` is provided, that means the user issuing the request wants to cancel an
+    // invite to someone else.
+    if (targetId) {
+      const foundTarget = await users.find(targetId)
+      if (!foundTarget) {
+        throw new httpErrors.NotFound('Target user not found')
+      }
+
+      leader = target
+      target = { id: foundTarget.id, name: foundTarget.name } as PartyUser
```
tec27, line 159: Unnecessary cast

server/lib/api/parties.ts
```diff
@@ -0,0 +1,182 @@
...
+      throw new httpErrors.Unauthorized(err.message)
+    case PartyServiceErrorCode.InvalidAction:
+      throw new httpErrors.BadRequest(err.message)
+    default:
+      assertUnreachable(err.code as never)
+  }
+}
+
+async function invite(ctx: RouterContext) {
+  const { clientId, targets } = ctx.request.body
```
tec27, line 111: If you're using `ctx.request.body` you should probably be casting it to some interface that matches what you've validated. Otherwise, all the things you pull out of it have type `any` which defeats a lot of the usefulness of a type system

server/lib/api/parties.ts
```diff
@@ -0,0 +1,182 @@
...
+  const { partyId, targetId } = ctx.params
+
+  try {
+    let leader: PartyUser | undefined
+    let target: PartyUser = { id: ctx.session!.userId, name: ctx.session!.userName }
+
+    // If the `targetId` is provided, that means the user issuing the request wants to cancel an
+    // invite to someone else.
+    if (targetId) {
+      const foundTarget = await users.find(targetId)
```
tec27, line 153: `ctx.params` is an object of `string`s, so I don't this actually works as you'd expect. You need to parse the number out instead, otherwise `users.find` is going to search it as a username (which is in hindsight, a bad API :) )

tec27 (changes requested):

package.json
```diff
@@ -25,6 +25,7 @@
   "jest": {
     "setupFilesAfterEnv": [
       "./babel-register.js",
+      "core-js/proposals/reflect-metadata",
```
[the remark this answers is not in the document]
2Pacalypse-: This seems to be necessary as soon as you import something that _does_ use injection, which I'm not sure we can avoid.

server/lib/api/parties.ts
2Pacalypse-: Hmm, I haven't tested this fully yet, but I thought `targetId` would be converted to a number by `Joi` since it was declared as a number in the schema.

tec27: Nope, we don't anything with the output of Joi so no conversions occur.

I've been thinking that we should move to a model where we call a function to validate params inside the handlers (instead of doing it as part of a Koa middleware chain) but I haven't written the code for that yet. This would fix the need to do conversions by hand and also not be as hard to deal with in TypeScript (probably).

tec27 (approved):

PR #594, opened 2021-02-23
Add documentation on how to obtain stardat.mpq and broodat.mpq

nickpainter:
Add documentation on how to obtain `stardat.mpq` and `broodat.mpq` now that they are no longer included with SCRM.


tec27:
Thank you!

PR #597, opened 2021-02-27
Rectify run server commands. Clarify how to get MQP files.

coin-au-carre:
There is no `package.json` file in the `server/` folder, `yarn start` cannot be available, I suppose it is legacy. 
Both proposed commands worked for me from project directory:

```
yarn run start-server
```

```
node ./server/index.js
```

### DIFF
--- a/docs/GETTING_STARTED.md
+++ b/docs/GETTING_STARTED.md
@@ -182,7 +182,7 @@ structure), and set `SB_SPRITE_DATA` in the `.env` file to that directory.
 
 ### Run the server
 
-The standard way to run the server is (assuming you are in the projet root directory):
+The standard way to run the server is (assuming you are in the project root directory):
 
 ```
 yarn run start-server
--- a/docs/GETTING_STARTED.md
+++ b/docs/GETTING_STARTED.md
@@ -174,8 +174,9 @@ The server needs access to some of BW's data files in order to generate map imag
 editor, such as [this one](http://www.zezula.net/en/mpq/download.html) and make sure to download
 "listfiles" from that website as well, which you'll need to use in the mpq editor. Use the mpq
 editor to extract BW's data files from `stardat.mpq`, `broodat.mpq`, in that order, having
-`broodat.mpq` overwrite any conflicting files from `stardat.mpq`. You can obtain `stardat.mpq` and `broodat.mpq`
-directly from Blizzard by downloading the final release of [StarEdit](http://download.blizzard.com/pub/starcraft/StarEdit/StarEdit.zip).
+`broodat.mpq` overwrite any conflicting files from `stardat.mpq`. You can obtain `stardat.mpq`
+and `broodat.mpq` directly from Blizzard by downloading the final release of
+[StarEdit](http://download.blizzard.com/pub/starcraft/StarEdit/StarEdit.zip).
 The necessary directories in .mpq files are `unit/` and `tileset/`. Extract those files to a directory (keeping the directory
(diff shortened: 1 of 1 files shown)


docs/GETTING_STARTED.md
```diff
@@ -174,24 +174,27 @@ The server needs access to some of BW's data files in order to generate map imag
 editor, such as [this one](http://www.zezula.net/en/mpq/download.html) and make sure to download
 "listfiles" from that website as well, which you'll need to use in the mpq editor. Use the mpq
 editor to extract BW's data files from `stardat.mpq`, `broodat.mpq`, in that order, having
-`broodat.mpq` overwrite any conflicting files from `stardat.mpq`. The necessary directories in
+`broodat.mpq` overwrite any conflicting files from `stardat.mpq`. 
+You can obtain `stardat.mpq` and `broodat.mpq` directly from Blizzard by downloading the final release of
+[StarEdit](http://download.blizzard.com/pub/starcraft/StarEdit/StarEdit.zip). You can also obtain those files from
+an old Starcraft Broodwar version (such as 1.16) installation folder. Note that you cannot find those files from SC:Remastered.
```
2Pacalypse-, line 180: _"Note that you cannot find those files from SC:Remastered."_ - This is not technically true ^^
Those files are still available in SC:R, but a bit harder to get. They now need to be extracted from a CASC container file using some program that can read those, e.g. [CascView](http://www.zezula.net/en/casc/main.html).

Honestly, I would just leave that line out, and perhaps even the one about using 1.16 installation folder. The method of using the StarEdit seems to be version-agnostic, so I would just mention that method to make things simpler.

Also, there might be some importance of casing for the name of those files that we might want to mention, but I'm not sure atm, so I'll leave someone else to comment on that ^^.

docs/GETTING_STARTED.md
```diff
@@ -174,24 +174,27 @@ The server needs access to some of BW's data files in order to generate map imag
...
+The necessary directories in
```
2Pacalypse-, line 181: Not sure if it affects the final output, but there's an odd line break here.

docs/GETTING_STARTED.md
coin-au-carre: Thanks for the comment I will remove the line I agree with you :)

docs/GETTING_STARTED.md
coin-au-carre: This does not affect the view since you need two blank spaces to represents a line break in Markdown.
I agree it is odd I will fix this.

docs/GETTING_STARTED.md
```diff
@@ -174,24 +174,25 @@ The server needs access to some of BW's data files in order to generate map imag
...
+`broodat.mpq` overwrite any conflicting files from `stardat.mpq`. You can obtain `stardat.mpq`
+and `broodat.mpq` directly from Blizzard by downloading the final release of
+[StarEdit](http://download.blizzard.com/pub/starcraft/StarEdit/StarEdit.zip).
+The necessary directories in .mpq files are `unit/` and `tileset/`. Extract those files to a directory (keeping the directory
+structure), and set `SB_SPRITE_DATA` in the `.env` file to that directory.
 
 ### Run the server
 
-The standard way to run the server is (from the `server` directory):
+The standard way to run the server is (assuming you are in the projet root directory):
```
2Pacalypse-, line 185: There is a typo here. It should be `project`.

coin-au-carre: Thx

PR #598, opened 2021-02-27
Add Cargo config rule to allow Cargo to fetch repositiories with git

coin-au-carre:
Using `game/build.bat` throws the following error

```
cargo build -v
    Updating crates.io index
error: failed to get `arrayvec` as a dependency of package `shieldbattery v0.1.0 (E:\Prog\ShieldBattery\game)`

Caused by:
  failed to fetch `https://github.com/rust-lang/crates.io-index`

Caused by:
  failed to authenticate when downloading repository: git@github.com:rust-lang/crates.io-index

  * attempted ssh-agent authentication, but no usernames succeeded: `git`

  if the git CLI succeeds then `net.git-fetch-with-cli` may help here
  https://doc.rust-lang.org/cargo/reference/config.html#netgit-fetch-with-cli

Caused by:
  error authenticating: failed connecting agent; class=Ssh (23)
```

After investigating the batch script I found that I had a `cargo build` authentification problem.
Following this https://github.com/rust-lang/cargo/issues/2078#issuecomment-434388584 helped

### DIFF
--- a/game/.cargo/config
+++ b/game/.cargo/config
@@ -3,3 +3,6 @@ target = "i686-pc-windows-msvc"
 
 [target.i686-pc-windows-msvc]
 rustflags = ["-Ctarget-feature=+crt-static"]
+
+[net]
+git-fetch-with-cli = true


neivv:
Hmm, so, looking at the linked issue and further on, it seems that your system has configured global .gitconfig to override HTTPS-based github fetches with SSH accesses instead? Or something to similar effect.
Seems that some other software does that to access github repos that require authentication, but Cargo isn't trying to handle SSH by itself like git CLI can and fails with that error.

That config change making Cargo use system git fixes it, but you should be editing system-global `~/.cargo/config` (Or well, in Windows `%USERPROFILE%/.cargo/config`) instead so that you don't have to add a config edit for any other Rust projects you may want to compile later.
And inversely changing that intended-to-be-system-wide config setting specifically for ShieldBattery building ends up being rude towards other people who prefer to keep that setting off.

Though I do think that giving instructions and even recommending to globally edit this config setting in GETTING_STARTED.md would be good. It also fixes issues with Cargo downloading its package index really slowly, which seems to have become pretty common problem recently.

coin-au-carre:
Thanks @neivv . I understand this change is not ideal, so maybe we could close this PR and add an issue to document a troubleshooting subsection in https://github.com/ShieldBattery/ShieldBattery/blob/master/docs/GETTING_STARTED.md#rust ? 
I can assign myself to it.

neivv:
Right, let's do that

PR #631, opened 2021-03-14
Matchmaking for random race

### LINKED ISSUE #630: Matchmaking for Random Race improvements
So I was trying the beta and launched  3 games as 'R' race with  Use alternate race 'T' to avoid mirror matchups. 

I was a little disappointed to have encountered 3 TvT so I digged a little in the code and I understood that was not that simple

The current code is straightforward in `server\lib\wsapi\matchmaking.ts`:

```ts
if (playersHaveSameRace && players.every(p => p.useAlternateRace === true)) {
  // All players have the same race and want to use an alternate race; select randomly one
  // player to use the alternate race and everyone else their main race. This is only done for
  // the first player, as the whole concept of the alternate race doesn't make much sense when
  // there are more than two players.
  const randomPlayerIndex = getRandomInt(players.size)
  slots = players.map((p, i) =>
    createHuman(p.name, p.id, i === randomPlayerIndex ? p.alternateRace : p.race),
  )
} else if (playersHaveSameRace && players.some(p => p.useAlternateRace === true)) {
  // All players have the same race, but only some of them are choosing to use an alternate
  // race; find the first player who wants to use the alternate race and have everyone else
  // use their main. Again, this is only done for the first player.
  const useAlternateRacePlayer = players.find(p => p.useAlternateRace === true)!
  slots = players.map(p =>
    createHuman(p.name, p.id, p.id === useAlternateRacePlayer.id ? p.alternateRace : p.race),
  )
} else {
  // All players have different race or don't want to use alternate race; nothing special to
  // do here.
  slots = players.map(p => createHuman(p.name, p.id, p.race))
}
```

In my case, there are 3 possibilities (one for each condition) which explain why I had several TvTs: 

1. My opponent has picked 'R' and has chosen an alternate race. We are in the `if (playersHaveSameRace && players.every(p => p.useAlternateRace === true))` case. One of us was chosen to use the alternate race by SB (me or the opponent if he has picked 'T' as an alternate race) and unfortunately the other one was chosen randomly as T by BW.
2. My opponent has picked 'R' but has no alternate race. We are in the `else if (playersHaveSameRace && players.some(p => p.useAlternateRace === true))` case. I was then chosen to play as 'T' by SB and my opponent was chosen randomly as T by BW. 
3. My opponent has picked 'T' so we were in the `else` case since 'R' and 'T' are considered different race, So BW launched as 'TvR'. I was unlucky and T was chosen randomly for me by BW (and not SB)

Note that in the 1st and 2nd cases, **one of the player is at disadvantage since one can be seen as 'R' in the game whereas the other has his/her race picked by SB.  I will call this advantage to be seen as 'Random' in BW game: the 'In game Random advantage'.**

Therefore, I have several propositions to improve the Matchmaking for random people (who hate mirror MUs also) and be "fair" as much as possible:

- If the two players are random and have chosen different alternate races to avoid mirrors, we should favor non mirror matchup. Hence SB should handle the randomness and we should show MU races over 'In game Random Advantage'. 
- If the two players are random and have chosen the same alternate races then both of them should be launched as 'R' by SB and both can have the 'In game Random advantage' (we should not favor one player over another).
- If the two players are random and only one has chosen a different alternate race then both of them should be launched as 'R' by SB. this one can be discussed but I think the one who assume every MUs possible should be favored more  than the other. 
- If the two players are random and have no alternate races so both of them should be launched as 'R' by SB. We should fall in the `else`  condition **which is not currently the case** since 2 Randoms are considered the same race according to SB.
- If one player has picked a specific race and the other one has picked 'R' with the option an alternate race.  then SB should handle the randomness to avoid mirror MUs for the 'R' player. We can still display 'R' in SB but in BW the race will be picked. 

For later, a nice to have option for random people who would favor  'In game Random advantage' would be the ability to choose:

  - 1st option: "I prefer avoiding mirror MUs over  In Game Random advantage". SB will handle the randomness, we can still display Random in SB but in BW the race will be picked. IMO this should be the default behavior. 
  - 2nd option: "I prefer to have the In Game Random advantage" over mirror MUs. 

I think both options could be nice to have if we explain well for random players the pros and cons for the 2 options. Personally  as a random player, I do not like mirror MUs! so I think we should implement the propositions mentioned above and see this nice to have later.

If that's ok for you I would be glad to be assigned and work on that, since I would like to participate in the 2v2 and 3v3 matchmakings that would be a good training.

coin-au-carre:
Fixes #630  

- [x] Update the find match form such that if random is selected, the alternate race checkbox is disabled and unchecked
- [x] Update the matchmaking preferences API to disallow alternate race selection with random
- [x] Update the matchmaking race resolution logic such that playersHaveSameRace is only true if the players are not random

### DIFF
--- a/client/matchmaking/find-match.jsx
+++ b/client/matchmaking/find-match.jsx
@@ -186,10 +186,10 @@ class Find1vs1MatchForm extends React.Component {
       getInputValue,
       onSubmit,
     } = this.props
+    const race = getInputValue('race')
     const useAlternateRace = getInputValue('useAlternateRace')
     const preferredMapsItems = Range(0, 2).map(index => {
       const map = preferredMaps.get(index)
-
       return (
         <PreferredMap key={index}>
           {map ? (
@@ -210,10 +210,12 @@ class Find1vs1MatchForm extends React.Component {
       <form noValidate={true} onSubmit={onSubmit}>
         <SectionTitle>Race</SectionTitle>
         <RaceSelect {...bindCustom('race')} size={RACE_PICKER_SIZE_LARGE} />
-        <CheckBox
-          {...bindCheckable('useAlternateRace')}
-          label='Use alternate race to avoid mirror matchups'
-        />
+        {race !== 'r' ? (
+          <CheckBox
+            {...bindCheckable('useAlternateRace')}
+            label='Use alternate race to avoid mirror matchups'
+          />
+        ) : null}
         {useAlternateRace ? (
           <>
             <SectionTitle>Alternate race</SectionTitle>
--- a/client/matchmaking/find-match.jsx
+++ b/client/matchmaking/find-match.jsx
@@ -187,7 +187,7 @@ class Find1vs1MatchForm extends React.Component {
       onSubmit,
     } = this.props
     const race = getInputValue('race')
-    const useAlternateRace = getInputValue('useAlternateRace')
+    const useAlternateRace = race !== 'r' ? getInputValue('useAlternateRace') : false
     const preferredMapsItems = Range(0, 2).map(index => {
       const map = preferredMaps.get(index)
       return (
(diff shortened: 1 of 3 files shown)


coin-au-carre:
All tasks should have been done

tec27:
Thank you!

PR #636, opened 2021-03-14
Don't use -1 as a default for SC:R settings

neivv:
Those defaults get passed to SC:R if there isn't anything else
specified, and at least sdGraphicsFilter being -1 straight up crashes.


neivv:
Could probably be worth it to have some sanity check for valid SC:R setting values right before we copy our settings there, just to avoid work from having to work backwards from a crash, but not going to investigate what values are acceptable right now.

PR #649, opened 2021-03-19
Hide alternate race icon that matches selected main race

GrantJenk:
I figured it looked a bit odd that you could select your alternate race to be the same as your main race.

Preview:
hide-race

### DIFF
--- a/client/lobbies/race-picker.jsx
+++ b/client/lobbies/race-picker.jsx
@@ -4,9 +4,10 @@ import styled from 'styled-components'
 
 import Button from '../material/button'
 import RaceIcon from './race-icon'
+import { Deselected } from './selected-race'
 
 import { fastOutSlowIn } from '../material/curve-constants'
-import { colorTextFaint, colorDividers } from '../styles/colors'
+import { colorTextFaint } from '../styles/colors'
 
 export const RACE_PICKER_SIZE_MEDIUM = 'MEDIUM'
 export const RACE_PICKER_SIZE_LARGE = 'LARGE'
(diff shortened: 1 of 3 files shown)


client/lobbies/selected-race.jsx
```diff
@@ -6,7 +6,7 @@ import { RaceButton, StyledRaceIcon } from './race-picker'
 
 import { colorDividers } from '../styles/colors'
 
-const Deselected = styled.span`
+export const Deselected = styled.span`
```
tec27, line 9: nit: if we're gonna be using this for other purposes, might be good to rename it, 'Deselected' doesn't make much sense for the new `disabledRace` case

client/lobbies/race-picker.jsx
```diff
@@ -87,27 +88,31 @@ export const StyledRaceIcon = styled(RaceIcon)`
 export default class RacePicker extends React.Component {
   static propTypes = {
     race: PropTypes.oneOf(['r', 'p', 't', 'z']).isRequired,
+    disableRace: PropTypes.oneOf(['r', 'p', 't', 'z']),
```
tec27, line 91: If you made this take a list instead of just 1 race, I think you could have `SelectedRace` forward completely to this component as well, might be a nice way to clean up the various implementations of this

(might be better to name it `hiddenRaces` or something in that case, since everything in `SelectedRace` is non-interactable)

client/lobbies/race-picker.jsx
```diff
@@ -87,27 +88,31 @@ export const StyledRaceIcon = styled(RaceIcon)`
...
     size: PropTypes.oneOf([RACE_PICKER_SIZE_MEDIUM, RACE_PICKER_SIZE_LARGE]),
     allowRandom: PropTypes.bool,
     onSetRace: PropTypes.func,
   }
 
   static defaultProps = {
+    disableRace: null,
```
tec27, line 98: Whenever we get around to moving more of these components to TS, `undefined` tends to function more nicely than `null`, I'd probably just leave this out of the defaultProps if the intention is that it has no value by default

tec27 (changes requested):

PR #652, opened 2021-03-23
Changes to left-nav children

### LINKED ISSUE #640: Left nav footer items should always be visible
In the case that the window is sized too small to display all the items in the left nav *and* keep the footer items showing, the left-nav items should scroll (in the area above the footer items)

Right now, the footer items just get cut off, which is not ideal since the user avatar is useful for various actions you can't accomplish other ways.

adaugherty42:
Footer will now be sticky while other left-nav content will scroll, if the window is small. This fixes #640

### DIFF
--- a/client/material/left-nav/left-nav.jsx
+++ b/client/material/left-nav/left-nav.jsx
@@ -4,7 +4,11 @@ import styled from 'styled-components'
 
 import { colorBackground } from '../../styles/colors'
 
-const Footer = styled.div``
+const Footer = styled.div`
+  position: sticky;
+  align-self: flex-end;
+  bottom: 0;
+`
 
 const Container = styled.nav`
   width: 240px;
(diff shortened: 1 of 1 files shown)


client/material/left-nav/left-nav.jsx
```diff
@@ -4,7 +4,11 @@ import styled from 'styled-components'
 
 import { colorBackground } from '../../styles/colors'
 
-const Footer = styled.div``
+const Footer = styled.div`
+  position: sticky;
```
tec27, line 8: I don't think any of these changes are are actually necessary, I think the overflow-y accomplishes everything we need (since the `Footer` is not contained within the `Sections`)

tec27:
Went ahead and made a change for just the overflow part so we can get this in for then next version, thank you for your help!

PR #664, opened 2021-04-06
Refactor the handling of chat message types.

2Pacalypse-:
This is the first (out of many I'm sure) steps that refactors the chat
stuff to make it more organized and with less repetitions. This PR
organizes and refactors a bunch of stuff related to message types.

The idea was simple:
  - Everything that is common to all messaging-related services (chat,
    whispers, lobbies, parties) should go in the "messaging" folder
  - Everything else that's specific to a particular message service goes
    in their respective folders

So related to the message types, in the messaging folder there's only
basic text message type defined that's used in all the services, while
chat and lobbies (whispers don't have any yet) specific message types
are defined in their own folder.

Besides definition of message types, this PR applies the same principle
for messages layout as well. Now each messaging-related service is in
charge of rendering its own messages. `MessageList` component has been
extended with an additional property which allows its users to define
how to render their message, with `MessageList` only rendering the
common message types (e.g. text message).

Lobby has been refactored to take advantage of this new setup.

### DIFF
--- a/client/chat/channel.tsx
+++ b/client/chat/channel.tsx
@@ -1,31 +1,41 @@
-import React, { useCallback, useRef, useState } from 'react'
 import PropTypes from 'prop-types'
+import React, { useCallback, useRef, useState } from 'react'
 import { connect } from 'react-redux'
-import { push } from '../navigation/routing'
-import styled, { css } from 'styled-components'
 import { List as VirtualizedList } from 'react-virtualized'
+import styled, { css } from 'styled-components'
+import { MULTI_CHANNEL } from '../../common/flags'
+import { User } from '../auth/auth-records'
+import Avatar from '../avatars/avatar'
+import { DispatchFunction } from '../dispatch-registry'
+import WindowListener from '../dom/window-listener'
+import MenuItem from '../material/menu/item'
+import MessageInput from '../messaging/message-input'
+import MessageList from '../messaging/message-list'
+import { Message } from '../messaging/message-records'
+import { push } from '../navigation/routing'
+import UserProfileOverlay from '../profile/user-profile-overlay'
+import LoadingIndicator from '../progress/dots'
+import { RootState } from '../root-reducer'
+import { alphaDisabled, colorDividers, colorTextSecondary } from '../styles/colors'
+import { body2, overline, singleLine } from '../styles/typography'
+import { navigateToWhisper } from '../whispers/action-creators'
 import {
-  sendMessage,
-  retrieveInitialMessageHistory,
-  retrieveNextMessageHistory,
-  retrieveUserList,
   activateChannel,
   deactivateChannel,
   joinChannel,
+  retrieveInitialMessageHistory,
+  retrieveNextMessageHistory,
+  retrieveUserList,
+  sendMessage,
 } from './action-creators'
-import { navigateToWhisper } from '../whispers/action-creators'
-
-import Avatar from '../avatars/avatar'
-import WindowListener from '../dom/window-listener'
-import MessageInput from '../messaging/message-input'
-import LoadingIndicator from '../progress/dots'
-import MessageList from '../messaging/message-list'
-import MenuItem from '../material/menu/item'
-import UserProfileOverlay from '../profile/user-profile-overlay'
-import { colorDividers, colorTextSecondary, alphaDisabled } from '../styles/colors'
-import { body2, overline, singleLine } from '../styles/typography'
-
-import { MULTI_CHANNEL } from '../../common/flags'
+import {
+  JoinChannelMessage,
+  LeaveChannelMessage,
+  NewChannelOwnerMessage,
+  SelfJoinChannelMessage,
+} from './chat-message-layout'
+import { ChatMessageType } from './chat-message-records'
+import { Channel as ChannelRecord, ChatState, Users as UsersRecord } from './chat-reducer'
 
 // Height to the bottom of the loading area (the top of the messages)
 const LOADING_AREA_BOTTOM = 32 + 8
@@ -78,7 +88,12 @@ const fadedCss = css`
 
 const USER_ENTRY_HEIGHT = 44
 
-const UserListEntryItem = styled.div`
+interface UserListEntryItemProps {
+  isOverlayOpen?: boolean
+  faded?: boolean
+}
+
+const UserListEntryItem = styled.div<UserListEntryItemProps>`
   ${body2};
   ${userListRow};
   height: ${USER_ENTRY_HEIGHT}px;
@@ -118,7 +133,14 @@ const UserListName = styled.span`
   display: inline-block;
 `
 
-const UserListEntry = React.memo(props => {
+interface UserListEntryProps {
+  user: string
+  faded?: boolean
+  style?: any
+  onWhisperClick: (user: string) => void
+}
+
+const UserListEntry = React.memo<UserListEntryProps>(props => {
   const [overlayOpen, setOverlayOpen] = useState(false)
   const userEntryRef = useRef(null)
 
@@ -140,7 +162,7 @@ const UserListEntry = React.memo(props => {
         onDismiss={onCloseOverlay}
         anchor={userEntryRef.current}
         user={props.user}>
-        <MenuItem text='Whisper' onClick={onWhisperClick} />
+        <MenuItem key='whisper' text='Whisper' onClick={onWhisperClick} />
       </UserProfileOverlay>
 
       <UserListEntryItem
@@ -156,13 +178,6 @@ const UserListEntry = React.memo(props => {
   )
 })
 
-UserListEntry.propTypes = {
-  user: PropTypes.string.isRequired,
-  onWhisperClick: PropTypes.func.isRequired,
-  faded: PropTypes.bool,
-  style: PropTypes.any,
-}
-
 const UsersVirtualizedList = styled(VirtualizedList)`
   &:focus,
   & > div:focus {
@@ -170,7 +185,12 @@ const UsersVirtualizedList = styled(VirtualizedList)`
   }
 `
 
-class UserList extends React.Component {
+interface UserListProps {
+  users: typeof UsersRecord
+  onWhisperClick: (user: string) => void
+}
+
+class UserList extends React.Component<UserListProps> {
   static propTypes = {
     users: PropTypes.object.isRequired,
     onWhisperClick: PropTypes.func.isRequired,
@@ -180,20 +200,20 @@ class UserList extends React.Component {
     width: 0,
     height: 0,
   }
-  _contentRef = React.createRef()
-  _listRef = React.createRef()
+  _contentRef = React.createRef<HTMLDivElement>()
+  _listRef = React.createRef<VirtualizedList>()
 
   componentDidMount() {
     this.updateDimensions()
   }
 
-  componentDidUpdate(prevProps) {
+  componentDidUpdate(prevProps: UserListProps) {
     if (prevProps.users !== this.props.users) {
       this._listRef.current?.recomputeRowHeights()
     }
   }
 
-  getRowHeight = ({ index }) => {
+  getRowHeight = ({ index }: { index: number }) => {
     const { active, idle, offline } = this.props.users
     if (index === 0) {
       return FIRST_OVERLINE_HEIGHT
@@ -229,7 +249,7 @@ class UserList extends React.Component {
     throw new Error('Asked to size nonexistent user: ' + index)
   }
 
-  renderRow = ({ index, style }) => {
+  renderRow = ({ index, style }: { index: number; style: any }) => {
     const { active, idle, offline } = this.props.users
     if (index === 0) {
       // NOTE(tec27): We know the active header is always visible because this user is online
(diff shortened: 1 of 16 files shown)


client/messaging/message-list.tsx
```diff
@@ -41,34 +35,44 @@ const Messages = styled.div`
 `
 
 interface PureMessageListProps {
-  messages: List<ChatMessage>
+  messages: List<Message>
+  renderMessages?: (msg: Message) => ReactNode
```
tec27, line 39: nit: `renderMessage` (non-plural) is more suitable imo

client/messaging/message-list.tsx
```diff
@@ -41,34 +35,44 @@ const Messages = styled.div`
...
-const PureMessageList = React.memo<PureMessageListProps>(({ messages }) => {
-  return <Messages>{messages.map(m => messageToElem(m))}</Messages>
+const PureMessageList = React.memo<PureMessageListProps>(({ messages, renderMessages }) => {
+  return (
+    <Messages>
+      {messages.map(m => {
+        // NOTE(2Pac): We only handle common messages here, e.g. text message. All other types of
+        // messages are handled by calling the `renderMessages` function below which should be
+        // supplied by each service if they have any special messages to handle.
+        if (Object.values(CommonMessageType).includes(m.type as CommonMessageType)) {
```
tec27, line 60: collecting the values and doing an includes on each message seems kind of expensive. At the very least I would collect the values into a readonly array outside this function (or a Set for a probably-faster inclusion check), but honestly I think a simpler approach is just:

```ts
return messageToElem(m) ?? (renderMessage ?? DEFAULT_RENDER_MESSAGE)(m)
```

client/messaging/message-records.ts
```diff
@@ -1,75 +1,37 @@
 import { Record } from 'immutable'
+import { ChatMessage } from '../chat/chat-message-records'
+import { LobbyMessage } from '../lobbies/lobby-message-records'
 
-export enum ChatMessageType {
+/**
+ * NOTE(2Pac): A common message type that's used in all messaging-related services (e.g. chat,
```
tec27, line 6: Not sure why this enum documentation is in the form of a `NOTE(...):`, you can just document it normally :)

client/messaging/message-records.ts
```diff
@@ -1,75 +1,37 @@
...
+ * whispers, lobbies, parties, etc.). For now that's only a default text message, but there might be
```
tec27, line 7: I'd leave out this 'For now...' sentence, it's not particularly useful compared to the source itself and is likely to be left around long past being true

client/messaging/message-records.ts
```diff
@@ -1,75 +1,37 @@
...
 }
 
 // TODO(tec27): Write a function or something to declare just the extra parts + do the correct
 // typing of the type field automatically.
-export class TextMessage
+export class TextMessageRecord
   extends Record({
     id: '',
-    type: ChatMessageType.TextMessage as typeof ChatMessageType.TextMessage,
+    type: CommonMessageType.TextMessage,
```
tec27, line 29: Does doing it this way break type discrimination? i.e. if you check that `msg.type === CommonMessageType.TextMessage`, does TS then know that `msg` is a `TextMessageRecord`?

If that still works (and works for the Chat/Lobby/Whisper message types too), cool. If not, you may need tot add a generic parameter to this class to specifies which range of enum values are valid (so for the Chat reducer this might be `BaseMessage<CommonMessageType | ChatMessageType>` 

Aside from that I'll also note that the cast here made it so the type was explicitly equal to the enum value specified, so you couldn't e.g. initialize a TextMessageRecord with `type: 'whateverFakeType'`, it *had* to have the right type if you were using that Record. Is there some compilation problem you're trying to avoid by removing the type, or did you just think it was unnecessary?

2Pacalypse-: I had no idea what I was doing so I removed them since TS compiler didn't complain :p

client/chat/channel.tsx
```diff
@@ -180,20 +200,20 @@ class UserList extends React.Component {
...
 
-  componentDidUpdate(prevProps) {
+  componentDidUpdate(prevProps: UserListProps) {
     if (prevProps.users !== this.props.users) {
       this._listRef.current?.recomputeRowHeights()
     }
   }
 
-  getRowHeight = ({ index }) => {
+  getRowHeight = ({ index }: { index: number }) => {
```
tec27, line 216: I'm pretty sure react-virtualized has types for these functions (this one and the one below), might be better to use those types, e.g.

```ts
getRowHeight: VirtualizedRowHeightFunction = ({ index }) => // ...
```

(some of them are definitely just done inline though, in which case you can ignore this advice)

client/chat/channel.tsx
```diff
@@ -427,23 +475,29 @@ class Channel extends React.Component {
...
     user: state.auth.user,
     chat: state.chat,
   }
 }
 
-function isLeavingChannel(oldProps, newProps) {
+interface ChatChannelViewProps {
+  user: SelfUserRecord
+  chat: typeof ChatState
+  dispatch: DispatchFunction<any> // TODO(2Pac): Type this better
```
tec27, line 488: Yes, do this in this PR :) (and the one below it)

I assume you didn't type this how we have it in DispatchRegistry because we're dispatching old-style actions in here, but I'd prefer that you type this correctly and cast those to `any` when you dispatch them instead. That way we get some type safety now, and as we migrate those things we get ever-more safety.

As for params, it's just what comes in from wouter, so it's:
`{ channel: string }`

client/lobbies/lobby.tsx
```diff
@@ -413,9 +222,10 @@ export default class Lobby extends React.Component {
     onRemoveObserver: PropTypes.func,
     onMapPreview: PropTypes.func,
     onToggleFavoriteMap: PropTypes.func,
+    onStartGame: PropTypes.func,
   }
 
-  getTeamSlots(team, isObserver, isLobbyUms) {
+  getTeamSlots(team: typeof Team, isObserver: boolean, isLobbyUms: boolean) {
```
tec27, line 228: I think the places you're using `typeof` here are wrong (in a similar way to how the `typeof [WhateverMessage]` stuff was in the prior version of this PR). You probably intend `ReturnType<typeof Team>` or something (but ideally we would migrate these records to be class-extends form as well so this stuff is unnecessary, don't expect you to do that in this PR though)

client/lobbies/lobby.tsx
```diff
@@ -448,7 +258,7 @@ export default class Lobby extends React.Component {
                 isObserver={isObserver}
                 canMakeObserver={!isObserver && canAddObsSlots && team.slots.size > 1}
                 canRemoveObserver={isObserver && canRemoveObsSlots}
-                onAddComputer={onAddComputer && !isLobbyUms ? () => onAddComputer(id) : undefined}
+                onAddComputer={!isLobbyUms && onAddComputer ? () => onAddComputer(id) : undefined}
```
tec27, line 261: You're treating onAddComputer (and all the following callbacks) as optional here but the typings on the props says it's required. If it really *is* optional you should add a `?` to the types

client/lobbies/lobby.tsx
```diff
@@ -394,7 +155,55 @@ const Countdown = styled.div`
   margin: 16px 0;
 `
 
-export default class Lobby extends React.Component {
+function renderChatMessages(msg: Message) {
```
tec27, line 158: similar to the comment on the last iteration of this, this probably makes more sense as `renderChatMessage`

client/messaging/message-list.tsx
```diff
@@ -41,34 +35,45 @@ const Messages = styled.div`
...
-      return <NewChannelOwnerMessage key={msg.id} record={msg} />
-    case ChatMessageType.SelfJoinChannel:
-      return <SelfJoinChannelMessage key={msg.id} record={msg} />
     default:
       return null
   }
 }
 
+const handleUnknown = (msg: Message) => {
+  throw new Error('Trying to render unknown message type: ' + msg.type)
```
tec27, line 52: I'm not sure it really makes sense to do this, we didn't enforce that all message types got rendered before. This also doesn't really offer any protection from anything, as it only ever gets called if `renderMessage` is not passed. Converting a situation that doesn't show a message into one that breaks the applications seems like a not great idea to me. If we could check that all cases were handled at compile-time, fine, but we can't really do that with how these messages are split up now.

So, tl;dr: just make this an empty function that returns null

tec27 (approved):
can merge when this stuff is fixed

client/chat/channel.tsx
2Pacalypse-: I'm not sure if I'm doing something wrong, but I can't find the types for those functions anywhere :/

client/chat/channel.tsx
```diff
@@ -463,109 +461,66 @@ class Channel extends React.Component<ChannelProps> {
...
-      !prevProps.chat.byName.has(routeChannel) &&
-      prevChannel.toLowerCase() !== routeChannel.toLowerCase()
-    ) {
+  // TODO(2Pac): Make this less spammy?
+  useEffect(() => {
+    dispatch(activateChannel(channelName) as any)
+  }, [channel])
+
+  useEffect(() => {
+    if (channel) {
```
tec27, line 486: There's a lint rule that checks these but imo it has a lot of false positives, but generally:

Don't use values from outside `useEffect` that you haven't declared as part of the dependencies. I can see why you'd think this would be safe, but it actually isn't. If you navigate to a channel you haven't yet joined, this will run once and join the channel. Then it will never run again, because joining the channel doesn't change the channel name. So the following code that we expect to run (retrieving the user list, retrieving the message history) will never occur

I would probably add `channel` to the deps, add additional checks inside the effect to ensure these only run on actual channel changes. Probably you should combine all these disparate effects back into a single one that checks all these things together, more similarly to the `componentDidUpdate`. I understand the appeal of splitting things up, but these things likely are actually dependent on each other and should be together.

tec27 (changes requested):

client/chat/channel.tsx
```diff
@@ -463,109 +461,66 @@ class Channel extends React.Component<ChannelProps> {
...
-  }
 
-  componentWillUnmount() {
-    this.props.dispatch(deactivateChannel(decodeURIComponent(this.props.params.channel)) as any)
-  }
-
-  render() {
-    const routeChannel = decodeURIComponent(this.props.params.channel)
-    const channel = this.props.chat.byName.get(routeChannel.toLowerCase())
+    return () => dispatch(deactivateChannel(channelName) as any)
```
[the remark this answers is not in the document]
2Pacalypse-: Am I missing something in how `useEffect` works, but I thought it would call the cleanup function each time after it executes the effect? And if activating the channel depends on the changes to `channel`, putting a cleanup function which deactivates the channel each time it's run into the same effect will just cause an infinite loop.

Seems to me like conceptually we're deactivating channel at the right times, e.g. when navigating away from a channel (either when a channel name in route params changes, or the component unmounts), but for channel activation we're not doing the same (well, symmetrical) thing. I think this is one of those cases where react's hook usage actually brings some underlying issues up to the surface. So, yay hooks? :d

I feel like we _should_ be able to write this hook:
```js
useEffect(() => {
  dispatch(activateChannel(channelName))

  return () => dispatch(deactivateChannel(channelName))
}, [channelName])
```
But currently we're activating the channel each time its record in the store changes, and I'm not sure why.

client/chat/channel.tsx
2Pacalypse-: After thinking a bit more about this, I realized my comment above is wrong on when should we be activating the channel. It totally makes sense to activate the channel only after we retrieve the initial message history as we've been doing. However, the reason I couldn't make that work is because we were depending on _any_ change to the `channel` record in the store, which makes things more complicated than they should be.

For example, when joining the channel, the `channel` record gets updated four distinct times and each time we execute code to retrieve message history, user list, and activate the channel. Doing that was mostly fine because those functions have failsafes on when to actually preform the action (except the channel activation one, which does it each time, thus making the action history a bit spammy). But since now with hooks we were also deactivating channel after each render, depending on _any_ change to the `channel` record doesn't seem to make much more sense anymore.

I feel like we could easily deal with this by simply calculating values that we're interested in outside the effect function, and then depend on only those values, instead of the whole `channel` record. Something like:
```js
const isInChannel = !!channel
const isLeavingChannel = !isInChannel && prevChannel && prevChannelName === channelName

useEffect(() => {
  if (isLeavingChannel) {
    push('/')
  }
}, [isLeavingChannel])

useEffect(() => {
  if (isInChannel) {
    dispatch(retrieveUserList(channelName))
    dispatch(retrieveInitialMessageHistory(channelName))
    dispatch(activateChannel(channelName) as any)
  } else if (!isLeavingChannel) {
    if (MULTI_CHANNEL) {
      dispatch(joinChannel(channelName))
    } else {
      push('/')
    }
  }

  return () => dispatch(deactivateChannel(channelName) as any)
}, [isInChannel, isLeavingChannel, channelName])
```
I tested this and it seems to be doing exactly what we need, but I'm not sure if I'm missing something.

tec27: Yeah, that looks much more correct to me 👍 

In general this is also why it's better to grab granular things out of the state rather than the entire record (also helps optimize re-renders from `useAppSelector`)

PR #667, opened 2021-04-16
Add a generic chat component that's used in all other messaging-related

2Pacalypse-:
services.

This PR pulls the common code related to the chat into a couple of
generic components which can now be easily used wherever we need chat
functionality.

Also refactors the code a bit of the components that are using these new
components (mainly chat channels, whispers, and lobbies).


client/messaging/message-input.tsx
```diff
@@ -0,0 +1,100 @@
...
+}
+
+export default function MessageInput(props: MessageInputProps) {
+  const [message, setMessage] = useState<string>('')
+  const inputRef = useRef<TextField | null>(null)
+
+  const onChange = useCallback(
+    (event: React.ChangeEvent<HTMLInputElement>) => {
+      const { value } = event.target
+      if (value !== message) {
```
tec27, line 39: useState hooks already do this check, no need to do it here (and it will make it so this callback doesn't get recreated on nearly every render anyway, which is also nice :) )

client/messaging/message-input.tsx
```diff
@@ -0,0 +1,100 @@
...
+`
+
+export interface MessageInputProps {
+  className?: string
+  showDivider?: boolean
+  onSendChatMessage: (msg: string) => void
+}
+
+export default function MessageInput(props: MessageInputProps) {
+  const [message, setMessage] = useState<string>('')
```
tec27, line 33: nit: shouldn't need the generic parameter here, it'll be inferred from the default you pass in

client/messaging/message-input.tsx
```diff
@@ -0,0 +1,100 @@
...
+  const inputRef = useRef<TextField | null>(null)
```
tec27, line 34: I *think* the `useRef` will already include `null` in the possible types if you pass null as the parameter, so you could just do:

```ts
const inputRef = useRef<TextField>(null)
```

(in VSCode, at least, you can mouse over these vars and see what type has been assigned to them)

client/messaging/message-input.tsx
```diff
@@ -0,0 +1,100 @@
...
+        }
+
+        inputRef.current?.focus()
+        setMessage(message + key)
+        return true
+      }
+
+      return false
+    },
+    [message, inputRef.current],
```
tec27, line 79: If you have a ref, you shouldn't be including `ref.current` in the dependencies list. Generally the point of refs is that they give you a box that you can access the current value of at any point in time (e.g. it will not be closed over by any closures, unlike a 1-level variable like `message` here)

client/messaging/message-input.tsx
```diff
@@ -0,0 +1,100 @@
...
+
+      const key = event.key ? event.key : String.fromCharCode(event.charCode)
+      if (key && key.length === 1) {
+        if (key === ' ' && target.tagName === 'BUTTON') {
+          // Space bar should click the button, rather than doing any of this
+          return false
+        }
+
+        inputRef.current?.focus()
+        setMessage(message + key)
```
tec27, line 73: If you use the function form, you can avoid needing to add `message` to the dependencies list of this (and make it so this function never needs to be recreated).

i.e:
```ts
setMessage(message => message + key)
```

tec27 (approved):

PR #668, opened 2021-04-18
Use the infinite-scroll-list component in chat channels and whispers.

2Pacalypse-:
Infinite-scroll-list component is now extended to support inserting data
both at the beginning and the ending of the list. At the same time even!
For now we're just inserting at one end of the list, but in future, we
might use it to jump into a middle of some data, that can dynamically
load both the previous data and the next data.

Also refactors a bunch of stuff with regards to fetching messages in
chat channels and whispers.


client/lists/infinite-scroll-list.tsx
```diff
@@ -0,0 +1,173 @@
...
+  /** Whether the functionality of loading more data at the beginning of the list is enabled. */
+  prevLoadingEnabled?: boolean
+  /** Whether the functionality of loading more data at the ending of the list is enabled. */
+  nextLoadingEnabled?: boolean
+  /** Whether we are currently loading more data at the beginning of the list. */
+  isLoadingPrev?: boolean
+  /** Whether we are currently loading more data at the ending of the list. */
+  isLoadingNext?: boolean
+  /** Whether this list has more data that could be loaded at the beginning of the list. */
+  hasMorePrevData?: boolean
```
tec27, line 24: nit: maybe `hasPrevData`/`loadPrevData` and the same for `Next`? The 'more' makes these names pretty weird and I don't think adds much to the meaning

client/messaging/chat.tsx
```diff
@@ -38,12 +38,8 @@ export default function Chat(props: ChatProps) {
       if (newIsScrolledUp !== isScrolledUp) {
         setIsScrolledUp(newIsScrolledUp)
       }
-
-      if (props.listProps.onScrollUpdate) {
-        props.listProps.onScrollUpdate(target)
-      }
     },
-    [isScrolledUp, props.listProps.onScrollUpdate],
+    [isScrolledUp],
```
tec27, line 42: Doesn't need to be done in this PR, but I think we could use the same IntersectionObserver to update this state (have an Observer at the bottom, when it's not intersecting `isScrolledUp = true`)

Seems like a potentially common case for users of infinite-scroll-list

tec27 (approved):

client/lists/infinite-scroll-list.tsx
```diff
@@ -0,0 +1,157 @@
...
+  /** Whether this list has data that could be loaded at the beginning of the list. */
+  hasPrevData?: boolean
+  /** Whether this list has data that could be loaded at the ending of the list. */
+  hasNextData?: boolean
+  /**
+   * A unique string value which will restart the intersection observer, i.e. disconnect and start
+   * observing again, when it changes. This is useful when the same instance of this component is
+   * used for lists with different content, e.g. in tabs, chat channels, and whisper sessions.
+   */
+  refreshToken?: string
```
tec27, line 34: You're not passing this as a string in some places already (e.g. carousel), so I don't think this type is right? There's no need for it to be a string either, anything that can be compared is fine, so I'd just make this `unknown`

client/messaging/message-list.tsx
```diff
@@ -79,16 +71,24 @@ export interface MessageListProps {
   loading?: boolean
   /** Whether this message list has more history available that could be requested. */
   hasMoreHistory?: boolean
+  /**
+   * A unique string value that's used by the infinite scroll list to decide when to refresh the
+   * list.
+   */
+  refreshToken?: string
```
tec27, line 78: Same as the other one, I don't think this type (or the comment) is correct. I also don't think the comment really describes its purpose or how to generate it, "unique" implies it is unique to something (e.g. throughout the app). A better way to describe this is "a value that changes when the values the list is displaying change, e.g. if the list is now displaying a different chat channel"

client/state-hooks.ts
```diff
@@ -19,3 +19,29 @@ export function usePrevious<T>(value: T): T | undefined {
   })
   return ref.current
 }
+
+/**
+ * A hook to move a prop to a ref, so it can be used in `useCallback` and `useEffect` hooks, without
+ * needing to specify it in the dependency arrays of those hooks, thus reducing the amount of times
```
tec27, line 25: This is a pretty roundabout way of saying what the purpose *actually* is, imo.

The real purpose is to make it so callbacks + effects can access the current value of the prop without needing to be re-run/re-created. The "dependencies array" part is incidental, that's just the mechanic React uses to make these decisions

client/state-hooks.ts
```diff
@@ -19,3 +19,29 @@ export function usePrevious<T>(value: T): T | undefined {
...
+ *
+ * export const CounterComponent = (props) => {
+ *   const countRef = usePropAsRef(props.count)
+ *   useEffect(() => {
+ *     if (countRef.current === 5) {
+ *       console.log('Count is 5!!!')
+ *     } else {
+ *       console.log('Count is not 5 :(')
+ *     }
+ *   }, [props.something])
```
tec27, line 38: I'm not sure what this value is doing in this array but this would also be a lint warning since the closure would close over the entire `props` value. If you wrote this as actual code, the linter would ask to you to assign `something` to a variable rather than dereference it off of `props`. (But, since it is not used and is meaningless to the example, my preference is that you just delete it and pass an empty array)

Side note: I also don't think it's a particularly good example of the usefulness of this, a better example would be e.g. accessing a value in a click handler

client/state-hooks.ts
```diff
@@ -19,3 +19,29 @@ export function usePrevious<T>(value: T): T | undefined {
...
+ *   return <div>Prop: {props.count}, ref: {countRef.current}</div>
+ * }
+ */
+export function usePropAsRef<T>(value: T): React.MutableRefObject<T> {
+  const ref = useRef<T>(value)
```
tec27, line 43: nit: generic parameter is unnecessary here as it can be inferred

client/lists/infinite-scroll-list.tsx
```diff
@@ -0,0 +1,157 @@
...
+    }
+
+    const nextTarget = nextTargetRef.current
+    if (nextLoadingEnabledRef.current && nextTarget) {
+      observer.current.observe(nextTarget)
+    }
+  }, []) // eslint-disable-line react-hooks/exhaustive-deps
+
+  const onIntersection = useCallback((entries: IntersectionObserverEntry[]) => {
+    entries.forEach(entry => {
```
tec27, line 105: nit: prefer `for...of` to `forEach` calls

client/maps/browse-server-maps.jsx
```diff
@@ -285,12 +282,10 @@ export default class Maps extends React.Component {
...
-    // after resetting the `currentPage` to 0, since the infinite list will start to load maps
-    // with whatever is currently saved in state.
-    this.setState(
-      () => ({ currentPage: 0 }),
-      () => this.infiniteList.reset(),
-    )
+    // before resetting the `currentPage` to 0, since the refresh token is a regular class field
+    // which doesn't cause a re-render.
+    this._refreshToken++
+    this.setState(() => ({ currentPage: 0 }))
```
tec27, line 288: nit: this can just be the passed state form of `setState` instead of the callback form now

tec27 (approved):

PR #669, opened 2021-04-25
Add a message to the chat when a day changes between any two messages.

2Pacalypse-:
Also reorganizes the message layout stuff a bit. Now a
common-message-layout file contains the implementation of the messages
that are common to all chat services, and message-layout file is a mixed
bag of primitives which can be used to style different kinds of
messages. The organization of this file might need some further
organization at some point, but should be fine for now.

### DIFF
--- a/client/chat/chat-reducer.js
+++ b/client/chat/chat-reducer.js
@@ -119,6 +119,7 @@ export default keyedReducer(new ChatState(), {
       return m.push(
         new SelfJoinChannelMessageRecord({
           id: cuid(),
+          time: Date.now(),
           channel,
         }),
       )
--- a/client/messaging/message-list.tsx
+++ b/client/messaging/message-list.tsx
@@ -1,4 +1,3 @@
-import cuid from 'cuid'
 import { List } from 'immutable'
 import PropTypes from 'prop-types'
 import React, { ReactNode } from 'react'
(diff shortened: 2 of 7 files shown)


client/messaging/message-list.tsx
```diff
@@ -49,11 +60,25 @@ const handleUnknown = (msg: Message) => {
...
-        return renderCommonMessage(m) ?? (renderMessage ?? handleUnknown)(m)
+        const messageLayout = renderCommonMessage(m) ?? (renderMessage ?? handleUnknown)(m)
+
+        const nextMessage = array.get(index + 1)
+        if (!nextMessage) {
+          return messageLayout
+        }
+
+        const newDayMessageRecord = new NewDayMessageRecord({
+          id: cuid(),
```
tec27, line 75: Doing this will mean these messages will never be the same on each render, you should prefer something deterministic instead (e.g. the timestamp concatenated with some string unique to new day messages)

client/messaging/message-list.tsx
```diff
@@ -49,11 +60,25 @@ const handleUnknown = (msg: Message) => {
...
+          time: nextMessage.time,
+        })
+
+        return !isSameDay(new Date(m.time), new Date(nextMessage.time))
```
tec27, line 79: Pretty weird to do this check *after* constructing the record for it, fairly wasteful. I know ternaries are cool but this can easily be a much less wasteful `else if`/`else` on the `if` above

client/messaging/message-list.tsx
```diff
@@ -49,11 +60,25 @@ const handleUnknown = (msg: Message) => {
...
+          ? [messageLayout, renderCommonMessage(newDayMessageRecord)]
+          : [messageLayout]
```
tec27, line 81: In the early return above, you just return a single item (instead of an array of a single item), so it feels a bit weird that you wrap it here

client/messaging/message-list.tsx
```diff
@@ -49,11 +60,25 @@ const handleUnknown = (msg: Message) => {
...
     <Messages>
-      {messages.map(m => {
+      {messages.map((m, index, array) => {
         // NOTE(2Pac): We only handle common messages here, e.g. text message. All other types of
         // messages are handled by calling the `renderMessage` function which should be supplied by
         // each service if they have any special messages to handle.
-        return renderCommonMessage(m) ?? (renderMessage ?? handleUnknown)(m)
+        const messageLayout = renderCommonMessage(m) ?? (renderMessage ?? handleUnknown)(m)
+
+        const nextMessage = array.get(index + 1)
```
tec27, line 69: It's a small thing, but generally I think you should prefer looking *backwards* rather than *forwards* to do things like this. This message only makes sense to show if the "real" message after it is rendered, and if we're looking backwards, that message is *this* one.

client/messaging/message-list.tsx
```diff
@@ -49,11 +60,25 @@ const handleUnknown = (msg: Message) => {
 const PureMessageList = React.memo<PureMessageListProps>(({ messages, renderMessage }) => {
   return (
     <Messages>
-      {messages.map(m => {
+      {messages.map((m, index, array) => {
```
tec27, line 63: `array` is not a good name for something that is not an array ;)

tec27 (changes requested):

client/chat/chat-reducer.js
```diff
@@ -119,6 +119,7 @@ export default keyedReducer(new ChatState(), {
       return m.push(
         new SelfJoinChannelMessageRecord({
           id: cuid(),
+          time: Date.now(),
```
tec27, line 122: In general this not a thing you want to do, as reducers are meant to be pure and deterministic. This time needs to be derived from something in the action (whether that's a time value passed by the action creator, or from some other part of the init message that already exists). I don't really think this is safe to do with a client-side `Date.now()` value either way, as all the other times in messages are server-provided. Ideally the server would provide a time value for this as well.

tec27 (approved):

client/chat/chat-reducer.js
2Pacalypse-: Fair enough. But I thought about fixing this in a separate PR, as a part of this issue: https://github.com/ShieldBattery/ShieldBattery/issues/139. Because as far as I can tell, we're doing this same thing for all the other non-user messages. I was thinking of moving the chat stuff to use HTTP pretty soon, and then do some server-side cleanup of chat stuff as a part of that.

PR #670, opened 2021-04-25
Combine win/loss columns into one and add a last played date column to

2Pacalypse-:
the ladder standings table.

### DIFF
--- a/client/ladder/ladder.tsx
+++ b/client/ladder/ladder.tsx
@@ -153,6 +153,7 @@ export interface LadderTableProps {
 
 export function LadderTable(props: LadderTableProps) {
   const [dimensionsRef, containerRect] = useDimensions()
+  const nowRef = useRef(Date.now())
   const containerRef = useRef<HTMLElement | null>(null)
   // multiplex the ref to the container to our own ref + the dimensions one
   const containerCallback = useCallback(
@@ -211,7 +212,7 @@ export function LadderTable(props: LadderTableProps) {
     )
   }, [])
   const renderLastPlayed = useCallback<TableCellRenderer>(props => {
-    return <NumberText>{timeAgo(props.cellData)}</NumberText>
+    return <NumberText>{timeAgo(Math.floor((nowRef.current - props.cellData) / 1000))}</NumberText>
   }, [])
 
   return (
--- a/client/ladder/ladder.tsx
+++ b/client/ladder/ladder.tsx
@@ -2,7 +2,6 @@ import { List } from 'immutable'
 import React, { useCallback, useEffect, useLayoutEffect, useRef, useState } from 'react'
 import { Column, Table, TableCellRenderer, TableHeaderProps } from 'react-virtualized'
 import styled from 'styled-components'
-import TimeAgo from 'timeago-react'
 import { LadderPlayer } from '../../common/ladder'
 import { MatchmakingType } from '../../common/matchmaking'
 import Avatar from '../avatars/avatar'
@@ -20,6 +19,7 @@ import {
   grey900,
 } from '../styles/colors'
 import { overline, subtitle1, subtitle2 } from '../styles/typography'
+import { timeAgo } from '../time/time-ago'
 import { getRankings } from './action-creators'
 
 const LadderPage = styled.div`
@@ -206,16 +206,12 @@ export function LadderTable(props: LadderTableProps) {
   const renderStats = useCallback<TableCellRenderer>(props => {
     return (
       <NumberText>
-        {props.cellData.wins} - {props.cellData.losses}
+        {props.cellData.wins} &ndash; {props.cellData.losses}
       </NumberText>
     )
   }, [])
   const renderLastPlayed = useCallback<TableCellRenderer>(props => {
-    return (
-      <NumberText>
-        <TimeAgo datetime={props.cellData} />
-      </NumberText>
-    )
+    return <NumberText>{timeAgo(props.cellData)}</NumberText>
   }, [])
 
   return (
@@ -268,7 +264,7 @@ export function LadderTable(props: LadderTableProps) {
         <Column
           label='Last played'
           dataKey='lastPlayedDate'
-          width={128}
+          width={96}
           columnData={{ rightAlignHeader: true }}
           cellRenderer={renderLastPlayed}
           headerRenderer={LadderTableHeader}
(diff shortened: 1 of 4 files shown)


yarn.lock
```diff
@@ -11325,6 +11325,18 @@ ticky@1.0.1:
...
   integrity sha1-t8+nHnaPHJAAxJe5FRswlHxQ5G0=
 
+timeago-react@^3.0.2:
+  version "3.0.2"
+  resolved "https://registry.yarnpkg.com/timeago-react/-/timeago-react-3.0.2.tgz#1bd8b86c37c93bcd6a4340fea340e6c9c2bdd4e0"
+  integrity sha512-K+NqoRcE13xv7lrmHtZmy7G+yXF8ZxmuwRK/EGKyT1chpI3zmV6YnB5o9+slnOCq1dXuQJknr/hWjINyP7C9LA==
+  dependencies:
+    timeago.js "^4.0.0"
+
+timeago.js@^4.0.0:
```
tec27, line 11335: These strings are much longer/more complex than what we have mocked, not sure I'm a huge fan of this library choice tbh. It also has a bunch of complexity around real-time updates of these strings that isn't very React-y and I'm not sure we really need

client/ladder/ladder.tsx
```diff
@@ -202,6 +203,20 @@ export function LadderTable(props: LadderTableProps) {
   const renderNumber = useCallback<TableCellRenderer>(props => {
     return <NumberText>{props.cellData}</NumberText>
   }, [])
+  const renderStats = useCallback<TableCellRenderer>(props => {
+    return (
+      <NumberText>
+        {props.cellData.wins} - {props.cellData.losses}
```
tec27, line 209: this should probably be an en dash

client/time/time-ago.ts
```diff
@@ -0,0 +1,51 @@
+const NOW = Date.now()
```
tec27, line 1: This doesn't actually work, this will be resolved to whatever time is "now" at the time this file is evaluated, not when the function is run.

I'd instead add another parameter to the `timeAgo` function that represents "now", and default it to `Date.now()` (defaults are executed when the function is called). That'd allow us to write tests for this easily but also make it simple to call if we don't care (note that it still might be good to call `Date.now()` once and pass that value for all further renders, just as a matter of efficiency

tec27: Alternatively you could just change the function to take the difference in milliseconds directly, and leave it up to the caller to figure out how they want to calculate that (might be preferable if we e.g. have APIs that return a difference already)

client/time/time-ago.ts
```diff
@@ -0,0 +1,48 @@
+const MINUTE = 60 // in seconds
+const HOUR = MINUTE * 60
+const DAY = HOUR * 24
+const WEEK = DAY * 7
+const MONTH = DAY * 30.42
+const YEAR = MONTH * 12
+
+/**
+ * A function which takes the number of seconds representing a difference between two dates, and
```
tec27, line 9: nit: avoid "A function that Xes" type comments in favor of "Xes ...."

e.g. this would be better written as:
"Returns a short string representing the difference between two dates."

I think most of the comment gets too much into the weeds of the actual implementation and can be left out, or replaced with a simple example with output commented

client/time/time-ago.ts
```diff
@@ -0,0 +1,48 @@
...
+ * returns a string representation of that difference, in "Xy ago" syntax. Where X equals amount of
+ * time, and y specifies the time unit.
+ *
+ * Note that this is not a *perfect* measurement of time. Some values are approximated, e.g. number
+ * of days in a month, and leap days/seconds/whatever are not taken into account.
+ */
+export function timeAgo(diff: number): string {
```
tec27, line 16: In general, try to name time values with a unit included, e.g. `diffMs` or `diffMillis`

client/time/time-ago.ts
```diff
@@ -0,0 +1,48 @@
...
+const YEAR = MONTH * 12
+
+/**
+ * A function which takes the number of seconds representing a difference between two dates, and
+ * returns a string representation of that difference, in "Xy ago" syntax. Where X equals amount of
+ * time, and y specifies the time unit.
+ *
+ * Note that this is not a *perfect* measurement of time. Some values are approximated, e.g. number
+ * of days in a month, and leap days/seconds/whatever are not taken into account.
+ */
```
tec27, line 15: For functions where parameters could use some explanation (e.g. you need to explain the format they're expected to be in, the units, etc.), it's good to include `@param` entries in the doc comment

client/ladder/ladder.tsx
```diff
@@ -152,6 +153,7 @@ export interface LadderTableProps {
 
 export function LadderTable(props: LadderTableProps) {
   const [dimensionsRef, containerRect] = useDimensions()
+  const nowRef = useRef(Date.now())
```
tec27, line 156: This value needs to be updated when the data is refreshed, might be easier to just use a memoized value with a dependency on the ladder data or something (we don't really care too much if it gets updated unnecessarily occasionally, just as long as it isn't getting updated every render)

client/ladder/ladder.tsx
```diff
@@ -202,6 +204,16 @@ export function LadderTable(props: LadderTableProps) {
...
   }, [])
+  const renderStats = useCallback<TableCellRenderer>(props => {
+    return (
+      <NumberText>
+        {props.cellData.wins} &ndash; {props.cellData.losses}
+      </NumberText>
+    )
+  }, [])
+  const renderLastPlayed = useCallback<TableCellRenderer>(props => {
+    return <NumberText>{timeAgo(Math.floor((nowRef.current - props.cellData) / 1000))}</NumberText>
```
tec27, line 215: Honestly I find any API that takes seconds to be pretty weird in JS, given that Dates operate on millis natively. I'd move this division back into the timeAgo where it was previously

tec27 (approved):

client/time/time-ago.ts
```diff
@@ -0,0 +1,53 @@
...
+const HOUR = MINUTE * 60
+const DAY = HOUR * 24
+const WEEK = DAY * 7
+const MONTH = DAY * 30.42
+const YEAR = MONTH * 12
+
+/**
+ * Returns a short string representing the difference between two dates.
+ *
+ * @param {number} diffMs Number representing a difference between two dates, in milliseconds.
```
tec27, line 11: type annotations are unnecessary, we're using typescript :)

PR #671, opened 2021-04-30
Implement the server side notifications and use it for party invites.

2Pacalypse-:


client/notifications/socket-handlers.ts
```diff
@@ -0,0 +1,29 @@
+import type { NydusClient, RouteInfo } from 'nydus-client'
+import { ReduxAction } from '../action-types'
+import { dispatch, Dispatchable } from '../dispatch-registry'
+import { addNotification } from './action-creators'
+
+const eventToAction: Record<string, (event: any) => Dispatchable<ReduxAction> | undefined> = {
```
tec27, line 6: You should really declare interfaces for these things like I did elsewhere, it is *very hard* to type these things after they're `any`, you have to do a lot of code-reading and debugging and looking at websocket messages, it sucks. So just do it up-front so we (I) don't have to spend hours guessing at this stuff please :)

server/lib/notifications/notification-model.ts
```diff
@@ -0,0 +1,110 @@
...
+  createdAt: Date
+  data: NotificationData
+}>
+
+function fromDbNotification(result: Readonly<DbNotification>): Notification {
+  return {
+    id: result.id,
+    type: result.data.type,
+    read: result.read,
+    createdAt: Number(result.created_at),
```
tec27, line 20: We have definitely talked about this before, you do *not* want to do this conversion at this level. Keep this a Date until the very level at which it is transmitted to the client. You will likely need a type separate from the type returned/used by these methods (and not shared with the clients as "this is the type we transmit to you") to do so, and that is fine and normal

client/material/popover.jsx
```diff
@@ -193,7 +198,8 @@ export default class Popover extends React.Component {
     const rect = {
       top: anchorElement.top + anchorOffsetVertical,
       right: anchorElement.right + anchorOffsetHorizontal,
-      bottom: anchorElement.bottom + anchorOffsetVertical,
+      // bottom: anchorElement.bottom + anchorOffsetVertical,
+      bottom: 600,
```
tec27, line 202: would *really* recommend using something that breaks the linter when you do stuff like this, e.g.
`bottom: 600, // FIXME`

client/material/popover.jsx
```diff
@@ -231,19 +237,27 @@ export default class Popover extends React.Component {
...
-        if (popoverPosition[pos] !== undefined) {
-          if (popoverPosition[pos] <= safeZoneVertical) {
-            popoverPosition[pos] = safeZoneVertical
-          } else if (popoverPosition[pos] + refRect.height >= clientHeight) {
-            popoverPosition[pos] = clientHeight - refRect.height - safeZoneVertical
-          }
+      if (popoverPosition.top !== undefined) {
+        if (popoverPosition.top <= safeZoneVerticalTop || refRect.height >= clientHeight) {
+          popoverPosition.top = safeZoneVerticalTop
+          popoverPosition.height = clientHeight - safeZoneVerticalTop - safeZoneVerticalBottom
```
tec27, line 243: I think the actual change you need to make is to give `Container` in here a max-height of `100%`, the lack of a height on there seems to be the real source of the issues here.

An easier way to accomplish the safe zones alongside that might be through padding on that element? Unsure, just a thought I had while looking at it.

It'd probably also be nice to allow people to pass a className that gets applied to `Container` (such that this is styleable and if we want a shorter max-height (like we might for notifications) that is possible in one-offs

2Pacalypse-: Yeah, you're right; `max-height` is enough to constrain the popover fit inside the window, instead of these shenanigans that I did ^^. It's not exactly the same, since we can't really specify safe zone on top and bottom separately, but it's good enough.

However, scrollbars don't work with `max-height` alone it seems, so `height` needs to be explicitly set as well. So I also had to set `height` in javascript once the popover's height is actually calculated based on its contents (and limited to `max-height`).

tec27: I think scrollbars will work fine without setting height in any pixel value, once you've set max-height on `Container`, you can set `height` on one of the children to 100% (and/or set `overflow`, one of those two things should make the layout constrain properly even with nested flexes). At no point should setting the height via JS be necessary (and doing so is quite weird since this component is also asking the component for its height post-layout)

common/notifications.ts
```diff
@@ -1,13 +1,36 @@
...
+}
+
+export type NotificationData = PartyInviteNotificationData
+
+export interface BaseNotification {
+  readonly id: string
+  readonly type: NotificationType
+  readonly read: boolean
+  readonly createdAt: number
+  readonly data?: NotificationData
```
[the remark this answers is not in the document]
tec27: Even if you remove the type from `NotificationData`, there is still zero guarantee that these things match. In fact, there's even less of a chance that you can guarantee they match at that point in the type system.

Like I said above, I think you're trying way too hard to share these type all the way through. We structure data in the database for reasons of the database: efficiency, query-ability, ability to update/migrate the schema, etc. There's no reason these same constraints apply to all the other ways we use these things, there's no reason that clients should need to deal with the data in a certain way just because it's structured that way in the database. Take the data out of the database, convert it to a form that is actually useful to clients and then send it to them, there's no real reason to share all these structures in this way.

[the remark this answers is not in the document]
2Pacalypse-: Yeah, that's what I did (I haven't pushed the changes yet). I let DB (and server side in general) have its own data structure for the notification, and the one here in the common folder is the one that's sent over the wire by the server and used on the client. I guess I was still sharing the `NotificationData` structure across the stack, which is probably why I had problems with typings when I tried to remove the `type` property in one place, but not the other :d. I'll just have a separate structure for the `NotificationData` that is saved in the DB, and the one defined here in the common folder will be the one with the optional `data` structure and without a `type` property in it.

Is that a good idea, or am I still misunderstanding things? ^^

client/material/popover.jsx
2Pacalypse-: As far as I know, setting `height` to 100% only works if some element above it also has a fixed `height` set; it doesn't work with `max-height` alone, even though I'm not really sure why. I've tried doing what you suggested, but I just couldn't get it to work. I'd be happy to be wrong and learn something new about CSS :d

common/notifications.ts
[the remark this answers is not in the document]
tec27: There's two options:
1) Have a container interface for all notifications that can contain some extra data. This container interface is never extended or discriminated on, it's always the same structure. Inside, the `data` field is a union type that can be discriminated
2) Have a base interface that has a `type` field that all notifications extend from. The notification type is then a union of all those notification types, and extra fields for a particular notification type are placed directly on this interface (*not* in a `data` field)

It sounds like you're doing some combination of these two approaches and I highly do not recommend that, if you do that you make it so that we have to cast `NotificationData` every single time because there's nothing to discriminate it with. We want a system here that does not require casting, that is the goal

2Pacalypse-:
Closed in favor of #673 and a new PR for party stuff that will come.

PR #673, opened 2021-05-09
Implement server-side notifications.

2Pacalypse-:


common/notifications.ts
```diff
@@ -2,12 +2,49 @@ export enum NotificationType {
...
+  /**
+   * A string that is unique if the notification is unique. For locally-generated notifications
+   * that should never show duplicates, this might be a constant string. For server-generated ones
+   * that have multiple of the same type (e.g. chat notifications, party invites, etc.) these are
+   * generated by the server.
+   */
+  readonly id: string
+  readonly type: NotificationType
+  readonly read: boolean
+  readonly createdAt: number
```
tec27, line 15: This could definitely use a comment (honestly probably just comment all the fields here)

server/lib/notifications/notification-model.ts
```diff
@@ -0,0 +1,146 @@
...
+ * created as unread, but visible by default.
+ */
+export async function addNotification({
+  userId,
+  data,
+  createdAt = new Date(),
+}: {
+  userId: number
+  data: NotificationData
+  createdAt?: Date
```
tec27, line 84: Not totally clear what happens here if this is undefined? Does postgres use the default value for the column? Might be more sensible to just add a default value for this param of `Date.now()`

server/lib/notifications/notification-service.ts
```diff
@@ -0,0 +1,71 @@
...
+} from './notification-model'
+
+export function getNotificationsPath(userId: number): string {
+  return `/notifications/${userId}`
+}
+
+@singleton()
+export default class NotificationService {
+  constructor(private nydus: NydusServer, private clientSocketsManager: ClientSocketsManager) {
+    this.clientSocketsManager.on('newClient', async c => {
```
tec27, line 23: Event handlers are not expected to be `async`, so this is adding a new spot that has potential rejected promises that get dropped on the floor. If you'd like to use an async function, you need to attach an error handler to it and do something with those errors. Likely in this case, it's a lot simpler to just write

```ts
  retrieveNotifications(...).then(notifications => {
  
  }, err => {

  })
```

tec27: I'll just note that this is also a race condition, if something sends this user a notification between the time this query hits the DB and comes back, they will not see that notification from this message *or* any message sent to the route (it won't have been subscribed yet).

Fixing this is probably a bit of a pain:

  - You could add promise locks around initialization that are awaited in mutations
  - You could keep track of when clients are being initialized and store info about the mutations that occurred during that, such that it could be appended to the `serverInit` message or just simply re-sent after (probably fine to re-send given IDs are unique)
  - Something else :)

(would be really nice if the nydus setup let us us subscribe and then send initial data in some async timeframe afterwards =/ )

server/lib/notifications/notification-model.ts
tec27: And actually looking at the migration, this doesn't even have a default value there, so definitely should have one here :)

client/notifications/notifications-popup.tsx
```diff
@@ -0,0 +1,171 @@
...
+  const dispatch = useAppDispatch()
+  const map = useAppSelector(s => s.notifications.map)
+  const ids = useAppSelector(s => s.notifications.ids)
+  const unreadIds = ids.filter(id => !map.get(id)?.read)
+  const prevIds = usePrevious(unreadIds.toSet()) || Set()
+  const newIds = unreadIds.toSet().subtract(prevIds)
+  const removedIds = prevIds.subtract(unreadIds.toSet())
+
+  const popupElems = useRef(new Map<string, HTMLDivElement>())
+  // TODO(2Pac): Replace this eslint rule with func-call-spacing as that one supports generics.
```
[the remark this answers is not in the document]
tec27: 665de0675bc67fdf3bbc5ed307d827648a39a7a7 fixes the lint rule stuff, so you can just remove the TODO here

client/notifications/activity-bar-entry.tsx
```diff
@@ -80,17 +85,26 @@ const PopoverContents = styled.div<PopoverContentsProps>`
...
 
   const [anchor, setAnchor] = useState<EventTarget | null>(null)
   const onClick = useCallback((event: React.MouseEvent) => {
     setAnchor(event.currentTarget)
   }, [])
   const onDismiss = useCallback(() => {
     setAnchor(null)
-    dispatch(markNotificationsRead())
-  }, [dispatch])
+    if (localNotifications?.count()) {
```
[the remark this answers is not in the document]
2Pacalypse-: Apparently, the `groupBy` function returns a [Keyed Collection](https://immutable-js.github.io/immutable-js/docs/#/Collection/Keyed) data structure which doesn't seem to have the `size` property as far as I can tell. Not sure if using the `count()` method is correct here or do we need to transform the Collection to something else first, that has a `size` property.

client/notifications/notification-reducer.ts
```diff
@@ -1,25 +1,23 @@
...
 
 export class EmailVerificationNotificationRecord
   extends Record({
-    type: NotificationType.EmailVerification as typeof NotificationType.EmailVerification,
     id: EMAIL_VERIFICATION_ID,
-    unread: true,
+    type: NotificationType.EmailVerification as typeof NotificationType.EmailVerification,
+    read: false,
+    createdAt: 0,
+    local: true,
```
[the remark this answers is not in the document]
2Pacalypse-: I'm not sure what I'm doing wrong, but I get this error when I try to do that:

image

client/notifications/notification-reducer.ts
```diff
@@ -35,26 +33,52 @@ function toNotificationRecord(notification: Readonly<Notification>): Notificatio
...
-  list: List<NotificationRecord>(),
-  ids: Set<string>(),
+  map: Map<string, NotificationRecord>(),
+  ids: List<string>(),
 }) {}
 
 export default keyedReducer(new NotificationState(), {
+  ['@notifications/serverInit'](state, { payload: { notifications } }) {
+    return state
+      .update('map', m => m.merge(notifications.map(n => [n.id, toNotificationRecord(n)])))
```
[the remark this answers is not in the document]
2Pacalypse-: I thought the `merge` method will already do that though? At least from what I can tell by reading the documentation, it should just update the existing keys to the new value. Unless you're talking about something else that I'm missing here.

client/notifications/notifications-popup.tsx
```diff
@@ -0,0 +1,171 @@
...
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  width: 48px;
+  height: 100%;
+  cursor: pointer;
+  background-color: ${blue900};
+
+  &:hover {
+    background-color: ${lighten(0.08, blue900)};
```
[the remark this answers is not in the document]
2Pacalypse-: Hmm, I'm not exactly sure how to make this work. When I use the suggested color for background, I get the grey color on hover, instead of just making the color of the button (which is blue) a bit lighter.

server/lib/notifications/notification-model.ts
2Pacalypse-: The `createdAt` has a default value in the argument list, so it _should_ use that if an `undefined` value is sent. I'm not sure if I'm missing something?

client/notifications/activity-bar-entry.tsx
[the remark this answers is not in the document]
tec27: That's fine then, sorry, I thought this was returning a `Map` or similar. `size` is preferred since it's always constant time, but if it requires a data structure conversion it's not worth avoiding `count()`.

client/notifications/notification-reducer.ts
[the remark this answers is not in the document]
tec27: Yeah, that's weird. Testing it out in TS Playground it seems it works as `true as true`. Odd that it is different from string literals but 🤷

client/notifications/notification-reducer.ts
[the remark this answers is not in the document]
tec27: This comment is on the line above the relevant one, sorry. Below this you're unshifting all of the IDs into a list, regardless of whether they are new to the map. So yes, the map will only have one version of each notification, but you might reference that notification 2+ times (and thus it will be displayed 2+ times).

client/notifications/notifications-popup.tsx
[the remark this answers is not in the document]
tec27: You need to add a separate element that sits on top of the content. That's why I said `::after`, e.g. a pseudo-element that you create in CSS. I'd really recommend just moving to IconButton for this particular usage so you don't need to worry about this, and we can fix this in just the Button components instead of having to fix it all over the place.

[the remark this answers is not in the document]
2Pacalypse-: I did move this to the `IconButton` component, but I still need to have these styles for hover/active states. At least until the `IconButton` is made to do that automatically, for any color of the background it might have.

client/notifications/notification-reducer.ts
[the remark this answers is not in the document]
2Pacalypse-: Using `true as true` gives me an eslint error though:

image

And if I change that to `true as const`, then I get another error I don't know how to fixx:

image

server/lib/notifications/notification-service.ts
2Pacalypse-: What about if we simply subscribe the user to the notifications path immediately when they connect, and then just publish them the server-init notifications once they're fetched from the DB?

tec27: Pretty wasteful, because you're publishing to every connected client for that user then. You could subscribe the user to two paths, e.g. `/notifications/<user>/` and `/notifications/<user>/<clientId>/` to make that work (still not perfectly, since they might have multiple connections for the same client at some points, but it's at least a bit better).

I think I may just try to make nydus allow for this though, doesn't seem like that huge of a change

client/notifications/notification-reducer.ts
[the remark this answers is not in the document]
tec27: huh, was not aware that was a thing, guess it got added after I last used TS significantly :) ( https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-4.html#const-assertions )

I think it may need a change to `NotificationRecordBase`, let me play around with it

client/notifications/notifications-popup.tsx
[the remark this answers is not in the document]
tec27: Yeah, I was looking at this last night, I think we can build better implementations of these button classes that share things more cleanly now that hooks exist. The way things are setup atm feels pretty backwards (with content components, etc.). If you get this PR ready before I make that change you can merge this as-is and I will clean it up then

client/notifications/notification-reducer.ts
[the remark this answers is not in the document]
tec27: Ah okay, actually straightforward:

In `common/notifications.ts`, you need to update `EmailVerificationNotification` to specify that `local` will always be `true` (instead of allowing it to be any `boolean` value). e.g.

```ts
export interface EmailVerificationNotification extends BaseNotification {
  type: typeof NotificationType.EmailVerification
  local: true
}
```

The error you were seeing above is because when TS has narrowed the type to just an `EmailVerificationNotification`, it still allowed for `local` to be `false` or `undefined` because of the base interface.

client/notifications/action-creators.ts
```diff
@@ -8,23 +12,29 @@ import { AddNotification, MarkNotificationsReadBegin } from './actions'
 export function clearNotifications(): ThunkAction {
   return (dispatch, getState) => {
     const { idToNotification, notificationIds } = getState().notifications
+    const newestNotification = idToNotification.get(notificationIds.last())
+    const timestamp = newestNotification?.createdAt ?? Date.now()
```
tec27, line 16: Passing `Date.now()` isn't safe because this is a local time. The time you've grabbed here also might be a time for a local notification (I think?).

Instead of the code you have here, I'd probably do:
```ts
const newestServerId = notificationIds.findLastEntry(id => {
  const notification = idToNotification.get(id)
  return notification && !notification.local
})
const timestamp = idToNotification.get(newestServerId)?.createdAt
```

(Note that `timestamp` can be `undefined`, but this will be fine with other changes I'll comment)

client/notifications/notification-reducer.ts
```diff
@@ -57,6 +57,11 @@ function updateReadStatus(
   )
 }
 
+// Cache the values of the cleared notifications so they can easily be restored if the server
+// returns an error.
+let clearedIdToNotification: Map<string, NotificationRecord>
```
tec27, line 62: This makes the reducer non-pure, bad idea

If you want to store these, they should be included in the reducer's state

client/notifications/notification-reducer.ts
```diff
@@ -93,17 +98,23 @@ export default keyedReducer(new NotificationState(), {
       .deleteIn(['notificationIds', notificationId])
   },
 
-  ['@notifications/clearBegin'](state) {
+  ['@notifications/clearBegin'](state, { payload: { timestamp } }) {
+    clearedIdToNotification = state.idToNotification.filter(n => n.createdAt <= timestamp)
```
tec27, line 102: Since local notifications don't have a server-relative timestamp (but also don't have the weirdness of time delay between sending/receiving), you should always clear all local notifications here and not try to compare times

tec27: You also need to be careful here, what if someone clicks the button 5 times before the server responds, and then the server responds with an error? A better solution here would be to track the request through a field on the action (e.g. give each request an ID, store these optimistically-removed items grouped under that ID, clear them out of this structure once you see a `@notifications/clear` with that id)

client/notifications/socket-handlers.ts
```diff
@@ -25,13 +25,29 @@ const eventToAction: EventToActionMap = {
     }
   },
 
+  clear: event => {
+    const { timestamp } = event
+    return {
+      type: '@notifications/clearBegin',
```
tec27, line 31: This is just a memory leak if used like this. You either need to:

- Add a separate action for this
- Dispatch both a `clearBegin` and a `clear`
or
- Make the `clear` case in the reducer attempt to remove items as well (if you do the grouping by request ID thing I suggested, you could do this only in the case that you didn't have a request ID or it wasn't in the map)

server/lib/notifications/notification-api.ts
```diff
@@ -25,8 +27,14 @@ export class NotificationApi extends HttpApi {
 }
 
 async function clearNotifications(ctx: RouterContext) {
-  // FIXME: this needs to use the service and needs to notify all connected clients
-  await clear(ctx.session!.userId)
+  const { body } = validateRequest(ctx, {
+    body: Joi.object<ClearNotificationsServerBody>({
+      timestamp: Joi.number().required(),
```
tec27, line 32: Timestamp should be optional because the client may not have any server notifications to pull a timestamp from. In the case that it isn't provided, the server should use its current time.

Since the client won't necessarily know what timestamp was used, the server should also return it in the response body (so this will be a `200` instead of a `204`).

server/lib/notifications/notification-model.ts
```diff
@@ -120,15 +120,16 @@ export async function markRead(userId: number, notificationIds: string[]): Promi
 }
 
 /**
- * Marks all notifications for a particular user as not visible, i.e. cleared.
+ * Marks all notifications before a given timestamp for a particular user as not visible, i.e.
+ * cleared.
  */
-export async function clear(userId: number): Promise<void> {
+export async function clear(userId: number, timestamp: number): Promise<void> {
```
tec27, line 126: nit: name this `clearBefore`

tec27: also make it take a `Date`, not a number, and fix the query to pass it directly

server/lib/notifications/notification-service.ts
```diff
@@ -45,8 +49,8 @@ export default class NotificationService {
   }
 
   /**
-   * Adds a new notification with the provided data to the DB, and notifies all clients to update
-   * their local lists.
+   * Creates a new notification for a particular user with the provided data to the DB and notifies
```
tec27, line 52: nit: this sentence doesn't really make sense. I think what you're trying to say is:

> Creates a new notification for a particular user, saves it to the database, and notifies all of the user's connected clients.

server/lib/notifications/notification-service.ts
```diff
@@ -67,6 +71,24 @@ export default class NotificationService {
     this.nydus.publish(getNotificationsPath(notificationProps.userId), addEventData)
   }
 
+  /**
+   * Clears, i.e. makes not visible, all notifications before a given timestamp for a particular
```
tec27, line 75: nit: this extra explanation isn't really necessary, just saying `Clears all notifications before a given timestamp for a particular user.` is fine

server/lib/notifications/notification-service.ts
```diff
@@ -67,6 +71,24 @@ export default class NotificationService {
...
+   * user and notifies all of user's connected clients.
+   */
+  async clear(userId: number, timestamp: number) {
```
tec27, line 78: nit: call this `clearBefore`, make it take a `Date`

server/lib/notifications/notification-service.ts
```diff
@@ -76,4 +98,18 @@ export default class NotificationService {
     }
     this.nydus.publish(getNotificationsPath(userId), clearByIdEventData)
   }
+
+  /**
+   * Markes all of the given notifications as "read" for a particular user and notifies all of
```
tec27, line 103: nit: Marks

tec27 (changes requested):

client/notifications/notification-reducer.ts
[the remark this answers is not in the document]
tec27: You marked this as resolved but the problem still exists?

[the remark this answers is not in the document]
tec27: Ah nevermind, I see you changed it to an OrderedSet, that's fine.

server/lib/notifications/notification-api.ts
2Pacalypse-: Is it really necessary to include it in the response body of the HTTP request if we're already including it in the websocket event that is published to all connected clients after the notifications are cleared?

client/notifications/socket-handlers.ts
2Pacalypse-: What about leaving this code here as it is, and instead have the `clearBegin` action in the reducer not attempt to cache any values if the request ID is not sent?

tec27: I think semantically that's a bit weird, since elsewhere in the app, `fooBegin` means `foo` is about to happen, and `foo` is the "commit" of that action. I'd prefer to keep those semantics here as well.

I get that you're trying to reuse code in the reducer, but I think consistent semantics are a better thing to optimize for here

server/lib/notifications/notification-api.ts
tec27: Not passing it back assumes a sort of implementation on the server that may not always be true, in general I'd prefer to think about these APIs in isolation from any other behaviors that may occur. It's also difficult to map a particular request to a particular websocket broadcast, those things are not connected like a request/response are.

client/notifications/notification-reducer.ts
```diff
@@ -1,60 +1,173 @@
...
+      n => timestamp && n.createdAt <= timestamp,
+    )
+    // Preserve the order of the cleared notification IDs set
+    const clearedNotificationIds = state.notificationIds.filter(id =>
+      clearedIdToNotification.has(id),
+    )
+
+    // Apply the clear changes optimistically
+    return state
+      .update('idToNotification', map =>
```
tec27: think we already went through a round of many places in this PR where `map` was a bad name choice, now it's made its way back in >_<

client/notifications/notifications-popup.tsx
```diff
@@ -0,0 +1,176 @@
...
+      enter: item => async (next, cancel) => {
+        cancelFuncs.current.set(item.id, cancel)
+        await next({ opacity: 1, height: popupElems.current?.get(item.id)?.offsetHeight })
+        await next({ duration: '0%' })
+      },
+      leave: { opacity: 0, height: 0 },
+      onRest: (result: AnimationResult, ctrl: Controller, item: NotificationRecord) => {
+        setNotificationItems(state => state.filter(i => i.id !== item.id))
+      },
+      /**
```
tec27: nit: only use `/**` for doc comments, otherwise just `/*` (or multiline with `//` is also fine)

(this is not a doc comment because it is in normal code, not on a class/interface)

client/notifications/notifications.tsx
```diff
@@ -3,12 +3,16 @@ import styled from 'styled-components'
 import { colorDividers } from '../styles/colors'
 import { body1 } from '../styles/typography'
 
-const Container = styled.div<{ unread: boolean }>`
+const Container = styled.div<{ read: boolean }>`
```
tec27: I probably added this so it's my mistake, but styled-components has a shorthand for props that should not be output to the DOM element now, which you should try to use (just prepend any such prop with a `$`)

tec27 (changes requested):

PR #674, opened 2021-05-20
Refactor the profile components so they're easier to extend.

2Pacalypse-:
This PR creates a ConnectedUserProfileOverlay component which allows the
usage of profile overlay without needing to send pre-made actions to it.
Threading required state/callbacks through the chat-channel components
just to create the user's profile overlay actions was a bit of a pain,
and was especially annoying once we started adding more actions to that
menu that depend on other application state, e.g. (the party invite actions
change depending on whether the user is in your party).

This makes it possible to use the user profile overlay at many different
places (something that we might start doing soon, e.g. in chat
messages), without a lot of boiler plate. However, this is done at the
expense of the flexibility in terms of which actions are displayed in
the overlay. If more flexibility is needed in certain scenarios, it is
still possible to use the un-connected user profile overlay component
and provide the actions manually to it.

Another thing that is left as a TODO in this PR is to be able to
configure the position of the user profile overlay through props,
something that will probably be required once we start showing this
overlay at other places.

Finally, this PR moves all the profile components to Typescript, so
that's nice.


client/profile/self-profile-overlay.tsx
```diff
@@ -0,0 +1,45 @@
...
+  from: { opacity: 0, scaleY: 0.5 },
+  enter: { opacity: 1, scaleY: 1 },
+  leave: { opacity: 0, scaleY: 0 },
+  config: (item, index, phase) => key =>
+    phase === 'leave' || key === 'opacity' ? { ...defaultSpring, clamp: true } : defaultSpring,
+}
+
+export function SelfProfileOverlay(props: SelfProfileOverlayProps) {
+  const { username, children, open, onDismiss, anchor } = props
+  const [, anchorX, anchorY] = useAnchorPosition('left', 'top', anchor ?? null)
```
tec27, line 25: These positions (and the origins below) don't seem right for this? It's extending upwards from the bottom of the screen, so you probably want center, top for the anchor, and center, bottom for the origin

tec27 (approved):

client/profile/self-profile-overlay.tsx
tec27: On second thought this probably identical given that the widths are the same, that's fine

PR #675, opened 2021-05-23
Add underline to activity buttons

### LINKED ISSUE #532: Underline letter on hotkey buttons
One thing I always thought was cool about the menus in Brood War was that they highlighted or underlined which letter can be pressed to activate the button. I thought it might be nice to mimic that. 

Possibly only add the underline or whatever other visual display when the alt key is pressed?

GrantJenk:
Fixes #532 

Underlines the first character of the label that matches its hotkey button.
Always shows, even if its modifier key is not pressed.

Preview: 
activity-buttons-underlined

### DIFF
--- a/client/activities/activity-button.tsx
+++ b/client/activities/activity-button.tsx
@@ -169,24 +169,24 @@ export const ActivityButton = React.memo(
         },
         [disabled, hotkey, onClick],
       )
-      const getLabelJsx = useMemo(() => {
+      const labelElems = useMemo(() => {
         if (disabled || !hotkey) {
           return label
         }
 
         const hotkeyString = String.fromCharCode(hotkey.keyCode).toLowerCase()
-        const labelJsx = []
+        const result = []
         let hasFoundHotkeyChar = false
         for (const char of label) {
           if (!hasFoundHotkeyChar && char.toLowerCase() === hotkeyString) {
-            labelJsx.push(<u key={char}>{char}</u>)
+            result.push(<u key={char}>{char}</u>)
             hasFoundHotkeyChar = true
           } else {
-            labelJsx.push(char)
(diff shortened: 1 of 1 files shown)


client/activities/activity-button.tsx
```diff
@@ -169,6 +169,25 @@ export const ActivityButton = React.memo(
         },
         [disabled, hotkey, onClick],
       )
+      const getLabelJsx = useMemo(() => {
```
tec27, line 172: nit: `getLabelJsx` makes this seem like a function when it's really the actual value

maybe call this `labelElems` and use that instead of `labelJsx` inside the function as well?

client/activities/activity-button.tsx
```diff
@@ -169,6 +169,25 @@ export const ActivityButton = React.memo(
...
+          return label
+        }
+
+        const hotkeyString = String.fromCharCode(hotkey.keyCode).toLowerCase()
+        const labelJsx = []
+        let hasFoundHotkeyChar = false
+        for (const char of label) {
+          if (!hasFoundHotkeyChar && char.toLowerCase() === hotkeyString) {
+            labelJsx.push(<u key={char}>{char}</u>)
+            hasFoundHotkeyChar = true
```
tec27, line 183: doesn't seem like this value is used for anything?

tec27 (approved):

client/activities/activity-button.tsx
GrantJenk: The purpose for this is to stop checking for more letters to underline after it's already found one. So, for example, the Ladder activity button would have both d's underlined without this check.

tec27: Oh sorry, didn't see it being used in the loop condition!

PR #678, opened 2021-05-28
Move the chat logic from WS API to a dedicated chat service.

2Pacalypse-:


server/lib/wsapi/chat.ts
```diff
@@ -78,22 +80,17 @@ export class ChatApi {
...
-    await addUserToChannel(user.session.userId, channel)
+    const { channel } = data.get('body')
 
-    this.state = this.state
-      .updateIn(['channels', channel], (s = Set()) => s.add(user.name))
-      .updateIn(['users', user.name], (s = Set()) => s.add(channel))
-    this._publishTo(channel, { action: 'join', user: user.name })
-    this._subscribeUserToChannel(user, channel)
+    try {
+      const chatService = container.resolve(ChatService)
```
tec27, line 89: Just a note (you don't need to fix this for this PR because I know this is moving to HTTP after this *anyway*), my advice before was only to not resolve this *at the top level*. It's perfectly fine to inject these into an instance of something and then use it as a field

server/lib/wsapi/chat.ts
```diff
@@ -43,6 +28,31 @@ const sendThrottle = createThrottle('chatsend', {
...
+
+  switch (err.code) {
+    case ChatServiceErrorCode.UserOffline:
+      throw new errors.NotFound(err.message)
+    case ChatServiceErrorCode.InvalidJoinAction:
+    case ChatServiceErrorCode.InvalidLeaveAction:
+    case ChatServiceErrorCode.InvalidSendAction:
+    case ChatServiceErrorCode.InvalidGetHistoryAction:
+    case ChatServiceErrorCode.InvalidGetUsersAction:
+      throw new errors.BadRequest(err.message)
```
tec27, line 48: we *really* need to do the error body thing sooner rather than later :) (Although I guess for bad requests it's not as bad)

server/lib/wsapi/chat.ts
```diff
@@ -168,29 +143,17 @@ export class ChatApi {
...
-      id: m.msgId,
-      user: m.userName,
-      sent: +m.sent,
-      data: m.data,
-    }))
+    try {
+      const chatService = container.resolve(ChatService)
+      return chatService.getChannelHistory(channel, user.name, limit, beforeTime)
+    } catch (err) {
+      return convertChatServiceError(err)
```
tec27, line 155: not gonna ask you to fix this either, because this is moving to HTTP, but in the future I think it's preferable to make this a `throw` to make the semantics clear. This works now because `convertChatServiceError` always throws, but looking at this code in isolation, this sure looks like it's returning a success response

tec27 (approved):

PR #679, opened 2021-05-28
Add a HTTP-based chat API (client is still using the WS API; will be

2Pacalypse-:
moved to HTTP in the next PR).

### DIFF
--- a/server/lib/chat/chat-api.ts
+++ b/server/lib/chat/chat-api.ts
@@ -1,0 +1,206 @@
+import Router, { RouterContext } from '@koa/router'
+import httpErrors from 'http-errors'
+import Joi from 'joi'
+import { container } from 'tsyringe'
+import { assertUnreachable } from '../../../common/assert-unreachable'
+import { CHANNEL_MAXLENGTH, CHANNEL_PATTERN } from '../../../common/constants'
+import { MULTI_CHANNEL } from '../../../common/flags'
+import { featureEnabled } from '../flags/feature-enabled'
+import { httpApi, HttpApi } from '../http/http-api'
+import ensureLoggedIn from '../session/ensure-logged-in'
+import createThrottle from '../throttle/create-throttle'
+import throttleMiddleware from '../throttle/middleware'
+import { validateRequest } from '../validation/joi-validator'
+import ChatService, { ChatServiceError, ChatServiceErrorCode } from './chat-service'
+
+const joinThrottle = createThrottle('chatjoin', {
+  rate: 3,
+  burst: 10,
+  window: 60000,
+})
+
+const leaveThrottle = createThrottle('chatleave', {
+  rate: 10,
+  burst: 20,
+  window: 60000,
+})
+
+const sendThrottle = createThrottle('chatsend', {
+  rate: 30,
+  burst: 90,
+  window: 60000,
+})
+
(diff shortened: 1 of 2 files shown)


2Pacalypse-:
Depends on https://github.com/ShieldBattery/ShieldBattery/pull/678

server/lib/chat/chat-api.ts
```diff
@@ -0,0 +1,206 @@
...
+@httpApi()
+export class ChatApi extends HttpApi {
+  constructor() {
+    super('/chat')
+    container.resolve(ChatService)
+  }
+
+  applyRoutes(router: Router): void {
+    router
+      .use(featureEnabled(MULTI_CHANNEL), ensureLoggedIn)
```
tec27, line 74: By putting this here you're disabling all chat APIs unless `MULTI_CHANNEL` is on

server/lib/chat/chat-api.ts
```diff
@@ -0,0 +1,206 @@
...
+        getChannelUsers,
+      )
+  }
+}
+
+function joinChannel(ctx: RouterContext) {
+  const {
+    params: { channelName },
+  } = validateRequest(ctx, {
+    params: Joi.object<{ channelName: string }>({
```
tec27, line 107: Would probably be good to pull both the interface and the validation out somewhere common, since this is the same across all these APIs

server/lib/chat/chat-api.ts
```diff
@@ -0,0 +1,206 @@
...
+function sendChatMessage(ctx: RouterContext) {
+  const {
+    params: { channelName },
+    body: { message },
+  } = validateRequest(ctx, {
+    params: Joi.object<{ channelName: string }>({
+      channelName: Joi.string().max(CHANNEL_MAXLENGTH).pattern(CHANNEL_PATTERN).required(),
+    }),
+    body: Joi.object<{ message: string }>({
+      message: Joi.string().min(1).required(),
```
tec27, line 150: Pretty sure this has a max length too?

server/lib/chat/chat-api.ts
```diff
@@ -0,0 +1,206 @@
...
+      channelName: Joi.string().max(CHANNEL_MAXLENGTH).pattern(CHANNEL_PATTERN).required(),
+    }),
+  })
+
+  try {
+    const chatService = container.resolve(ChatService)
+    chatService.joinChannel(channelName, ctx.session!.userName)
+
+    ctx.status = 204
+  } catch (err) {
```
tec27, line 117: You could write a simple thing to `.use` above that would do this try/catch logic for everything, btw:

```ts
router.use(ensureLoggedIn, convertChatServiceErrors)
// ...

function convertChatServiceErrors(ctx, next) {
  try {
    await next()
  } catch (err) {
    convertChatServiceError(err)
  }
}
```

tec27 (changes requested):

server/lib/chat/chat-api.ts
2Pacalypse-: Yeah, but we're simply cutting it off here: https://github.com/ShieldBattery/ShieldBattery/blob/master/server/lib/messaging/filter-chat-message.js instead of throwing an error.

PR #680, opened 2021-05-28
Make the client use the HTTP-based API in chat.

2Pacalypse-:


2Pacalypse-:
Depends on https://github.com/ShieldBattery/ShieldBattery/pull/678 and https://github.com/ShieldBattery/ShieldBattery/pull/679

client/chat/action-creators.js
```diff
@@ -54,7 +55,10 @@ export function sendMessage(channel, message) {
     })
     dispatch({
       type: CHAT_SEND_MESSAGE,
-      payload: siteSocket.invoke('/chat/send', params),
+      payload: fetch(apiUrl`chat/${channel}`, {
```
tec27, line 58: Seeing this in use here, maybe this would make more sense as a separate sub-url (e.g. `/chat/<channel>/messages`)? This feels like it'll be a bit weird when we let people change settings of the channel, which would likely be the same URL and method

client/chat/action-creators.js
```diff
@@ -80,7 +84,10 @@ export function getMessageHistory(channel, limit) {
     })
     dispatch({
       type: CHAT_LOAD_CHANNEL_HISTORY,
-      payload: siteSocket.invoke('/chat/getHistory', params),
+      payload: fetch(
+        apiUrl`chat/${channel}/chat?limit=${limit}&beforeTime=${earliestMessageTime}`,
```
tec27, line 88: And yeah, think this URL should match the one we use for posting a message, so `/chat/<channel>/messages` or `/chat/<channel>/chat`, one of the two (I vaguely prefer `messages` because it is less redundant/filled with multiple meanings for the same word :) )

Feel free to put these changes in the previous commit if you prefer

tec27 (changes requested):

PR #681, opened 2021-05-28
Move the chat models file to its feature folder.

2Pacalypse-:

PR #683, opened 2021-05-29
 Add types to the chat actions and events.

2Pacalypse-:


2Pacalypse-:
Depends on https://github.com/ShieldBattery/ShieldBattery/pull/678 and https://github.com/ShieldBattery/ShieldBattery/pull/679 and https://github.com/ShieldBattery/ShieldBattery/pull/680 and https://github.com/ShieldBattery/ShieldBattery/pull/681

common/chat.ts
```diff
@@ -0,0 +1,60 @@
...
+  user: string
+  sent: number
+  data: {
+    type: string
+    text: string
+  }
+}
+
+// TODO(2Pac): Make this into an interface and include more information here
+export type ChatUser = string
```
tec27, line 12: Seems like you intended to use this in actions but then never did? (I'd just delete tbh, I don't find it particularly useful for refactoring). Either delete or actually use it in these types, I have no particular preference as I don't think it makes refactoring any easier or harder

server/lib/chat/chat-models.ts
```diff
@@ -48,13 +48,18 @@ export interface UserChannelsEntry {
   joinDate: Date
 }
 
+export interface ChatMessageData {
```
tec27, line 51: This isn't a correct typing imo, this is only the form for one particular type (we may only have one type atm, but the way you have typed this does not indicate that we expect there to be more in the future and that adding more forms is acceptable/expected)

server/lib/chat/chat-service.ts
```diff
@@ -158,12 +176,16 @@ export default class ChatService {
       limit,
       beforeTime && beforeTime > -1 ? new Date(beforeTime) : undefined,
     )
-    return messages.map(m => ({
-      id: m.msgId,
-      user: m.userName,
-      sent: +m.sent,
-      data: m.data,
-    }))
+    return messages.map(m => {
```
tec27, line 179: nit: you can write this more succinctly by passing a generic parameter to `map`, i.e.

```ts
return messages.map<ChatMessage>(m => ({
  id: m.msgId,
  user: m.userName,
  sent: Number(m.sent),
  data: m.data,
}))
```

(applies to other places in this file too)

server/lib/chat/chat-service.ts
```diff
@@ -206,12 +232,11 @@ export default class ChatService {
...
   private subscribeUserToChannel(userSockets: UserSocketsGroup, channelName: string) {
-    userSockets.subscribe(getChannelPath(channelName), () => {
-      return {
-        action: 'init',
-        activeUsers: this.state.channels.get(channelName),
-      }
-    })
+    const initEventData: ChatInitEvent = {
+      action: 'init',
+      activeUsers: (this.state.channels.get(channelName) ?? Set()).toArray(),
```
tec27, line 237: nit: prefer creating the JS things as a fallback instead of creating an immutable thing and then converting it, e.g.
```ts
this.state.channels.get(channelName)?.toArray() ?? []
```

Although honestly for this particular case, if this isn't set, this seems like programmer error, so maybe `!` is just a more appropriate thing to do

tec27 (changes requested):

server/lib/chat/chat-service.ts
```diff
@@ -200,18 +206,16 @@ export default class ChatService {
...
   }
 
   private subscribeUserToChannel(userSockets: UserSocketsGroup, channelName: string) {
-    userSockets.subscribe(getChannelPath(channelName), () => {
-      return {
-        action: 'init',
-        activeUsers: this.state.channels.get(channelName),
-      }
-    })
+    const initEventData: ChatInitEvent = {
```
tec27, line 214: You changed the semantics of this. Before, every time a client connected for this user, it would retrieve the active users for the channel.

With this change, it retrieves the active users for the channel *when this user was first subscribed* and never updates it again. Please revert

server/lib/chat/chat-service.ts
```diff
@@ -180,7 +186,7 @@ export default class ChatService {
     }
 
     const users = await getUsersForChannel(originalChannelName)
-    return users.map(u => u.userName)
+    return users.map<ChatUser>(u => u.userName)
```
tec27, line 189: nit: this generic param is unnecessary, it can be easily inferred. If you'd prefer to note that this is a `ChatUser` and not a `string` (which maybe you can see why I don't find this type to provide much value :) ), do so by notating the return type on the function (which you should generally always be doing for non-simple methods, and this method certainly isn't simple)

tec27 (changes requested):

tec27 (approved):

PR #685, opened 2021-05-29
Add a middleware to parties API to convert party service errors.

2Pacalypse-:

PR #686, opened 2021-05-29
Validate the party request inline instead of through a middleware.

2Pacalypse-:


2Pacalypse-:
Depends on https://github.com/ShieldBattery/ShieldBattery/pull/685

PR #688, opened 2021-05-29
Finish the party invites system.

2Pacalypse-:
This PR only implements stuff related to party invites, utilizing our
brand new notification service. The actual party UI, along with chat and
everything will be done in a separate PR.


2Pacalypse-:
Depends on https://github.com/ShieldBattery/ShieldBattery/pull/687

client/parties/actions.ts
```diff
@@ -0,0 +1,121 @@
...
+  | RemoveInvite
+  | Init
+  | Invite
+  | Uninvite
+  | Decline
+  | Join
+  | Leave
+
+export interface InviteToPartyBegin {
+  type: '@parties/inviteToPartyBegin'
```
tec27, line 27: These `Begin` actions should *always* include information about the action that is being performed. "I'm about to do a party invite" is really useless information if you don't know *who* is being invited, there's nothing you can do with that info.

Apply throughout

client/parties/actions.ts
```diff
@@ -0,0 +1,121 @@
+import { PartyPayload, PartyUser } from '../../common/parties'
+import { BaseFetchFailure } from '../network/fetch-action-types'
+
+export type PartiesActions =
+  | InviteToPartyBegin
+  | InviteToPartySuccess
+  | BaseFetchFailure<'@parties/inviteToParty'>
```
tec27, line 7: Vastly, vastly prefer that you name these and put them next to the relevant Success type

client/parties/actions.ts
```diff
@@ -0,0 +1,121 @@
...
+}
+
+export interface RemoveInvite {
+  type: '@parties/removeInvite'
+  payload: {
+    partyId: string
+  }
+}
+
+export interface Init {
```
tec27, line 81: These names are a bit too short and will make these a pain to import anywhere, go with things like `InitParty` instead

client/parties/actions.ts
```diff
@@ -0,0 +1,121 @@
...
+  type: '@parties/acceptPartyInviteBegin'
+}
+
+export interface AcceptPartyInviteSuccess {
+  type: '@parties/acceptPartyInvite'
+  payload: void
+  error?: false
+}
+
+export interface AddInvite {
```
tec27, line 66: Add and Remove feel really unnecessary? At least from the structure of these it seems like they're for when *you* have been invited to a party, but this is meant to be handled via the notification system. Unless you want some UI to display every party you're currently invited to (and I don't think we do), I can't see how these would be useful or desirable

client/parties/actions.ts
```diff
@@ -0,0 +1,121 @@
...
+}
+
+export interface Init {
+  type: '@parties/init'
+  payload: {
+    party: PartyPayload
+  }
+}
+
+export interface Invite {
```
tec27, line 88: These sorts of actions need more context than you're giving here, because we can't guarantee they run in particular orders. All these sorts of actions should say not only the change the occurred but *what it was a change to*, e.g. the party ID. In the event that the reducer does not register that party as the current one, it discards the action without changing state

client/parties/actions.ts
```diff
@@ -0,0 +1,121 @@
...
+  | Decline
+  | Join
+  | Leave
+
+export interface InviteToPartyBegin {
+  type: '@parties/inviteToPartyBegin'
+}
+
+export interface InviteToPartySuccess {
+  type: '@parties/inviteToParty'
```
tec27, line 31: Here too, we have zero useful information on these actions. We invited someone to a party, great. Who was it? No clue. What party was it for? Not sure.

`meta` is the typical property used when we need to provide information that the server won't return to us, and that's what you should be using for all of these.

If you want all successes to be handled through websocket publishes, fine, but I'm not sure why these useless success actions exist in that case, and you'd be better off thinking more deeply about the structure of these things instead of just copy-pasting the structure of other systems

client/parties/parties-reducer.ts
```diff
@@ -0,0 +1,94 @@
...
+}) {}
+
+export class PartyInvite extends Record({
+  partyId: '',
+  from: new PartyUserRecord(),
+}) {}
+
+export class PartiesState extends Record({
+  party: new PartyRecord(),
+  invites: Map<string, typeof PartyInvite>(),
```
tec27, line 27: Yeah, I really don't get the purpose of this structure. If we're using notifications to accept party invites, the notification should contain the info about the party we've been invited to and are accepting. This is duplicated information that we don't have a use for, and is likely to fall out of sync

server/lib/notifications/__mocks__/notification-service.ts
```diff
@@ -0,0 +1,17 @@
+const retrieveNotifications = jest.fn().mockResolvedValue([])
```
tec27, line 1: I'm not a fan of doing this, it obscures the fact that this is a mock in tests and also makes it impossible to inspect the mock without a cast, would *vastly* prefer you either

a) mock directly in the test file
or
b) write a fake in a file that is not here and does not automatically replace the import. With injection, using mocked imports should be unnecessary

This file also makes zero references to the original types, and thus if it gets out of sync we will never know unless the test fails, and that's really not great.

server/lib/parties/party-service.test.ts
```diff
@@ -82,33 +93,52 @@ describe('parties/party-service', () => {
     let leader: PartyUser
     let party: PartyRecord
 
-    test('should throw if inviting yourself', () => {
-      expect(() =>
+    test('should throw if inviting yourself', async () => {
+      await expect(() =>
```
tec27, line 97: is the function form necessary on these? Seems like if it's a promise you can just pass the promise directly? (The point of the function is to give it something it can `try/catch` around)

server/lib/parties/party-service.ts
```diff
@@ -33,6 +36,7 @@ export enum PartyServiceErrorCode {
   PartyFull,
   UserOffline,
   InvalidAction,
+  NotificationFail,
```
tec27, line 39: nit: `NotificationFailure`

server/lib/parties/party-service.ts
```diff
@@ -95,11 +97,26 @@ export default class PartyService {
...
+      }
+
       party.invites.set(invitedUser.id, invitedUser)
-      this.publishToParty(party.id, {
+      const inviteEventData: PartyInviteEvent = {
         type: 'invite',
         invitedUser,
-      })
+      }
+      this.publishToParty(party.id, inviteEventData)
```
tec27, line 119: Now that you have these things typed, you could change `publishToParty` to:
```ts
private publishToParty<T extends PartyEvent>(partyId: string, data: T)
```

and then you could use it without pulling these into separate fields or needing to declare the event type.

tec27 (changes requested):

2Pacalypse-:
Depends on #707

server/lib/parties/party-service.ts
```diff
@@ -216,26 +220,33 @@ export default class PartyService {
...
 
-  private unsubscribeFromInvites(party: PartyRecord, user: PartyUser) {
-    this.nydus.publish(getInvitesPath(party.id, user.id), { type: 'removeInvite' })
-    // TODO(2Pac): Remove the invite notification once the server-side notification system is in.
-    const userSockets = this.userSocketsManager.getById(user.id)
-    if (userSockets) {
-      userSockets.unsubscribe(getInvitesPath(party.id, user.id))
-    }
+  private async clearInviteNotification(partyId: string, user: PartyUser) {
+    this.notificationService
```
tec27, line 231: This method is async, so it seems like the intention is that you can use `await`, but then you only attach promise handlers. If the intention is that we don't wait for this clearing to complete and it has no bearing on the initiating request, that's fine, but it probably shouldn't be an async function in that case

server/lib/parties/party-service.ts
```diff
@@ -216,26 +220,33 @@ export default class PartyService {
...
+      .retrieveNotifications({ userId: user.id, type: NotificationType.PartyInvite })
+      .then(
+        userInviteNotifications => {
+          const notification = userInviteNotifications.filter(n => n.data.partyId === partyId)[0]
+          if (notification) {
+            this.notificationService.clearById(user.id, notification.id).catch(err => {
```
tec27, line 237: If you make the containing callback `async`, you can `await` this instead of nesting attached callbacks. Then you could add a single catch handler (after the above `then` instead of as the second parameter) that would handle all errors inside of here.

(I don't think the separate log messages are particularly useful here, given that we have the error attached and it will have a stack trace)

At some point (doesn't need to be fixed as part of this PR), we should also really pass loggers down to services like this, so that they're linked to a `request` and such. Pretty annoying here that these operations are associated with a particular request but we'll have no idea which

tec27 (approved):

Issue #239, opened 2017-02-18
Improve error message for replay folder not existing


tec27:
Currently we just show an ENOENT or something, which is... crap

Ideally we:

- Have a decent error message here
- Possible create an empty folder in the case of this being the root missing

PR #689, opened 2021-05-30
Make missing filepath error more clear

williamhammond:
Closes https://github.com/ShieldBattery/ShieldBattery/issues/239. I figured the improved logging route would be better since this would effect more than looking up replays (I think) and on a good broodwar install the folder should be there anyway. 
missing-replays

### DIFF
--- a/client/file-browser/get-files.js
+++ b/client/file-browser/get-files.js
@@ -2,7 +2,16 @@ import { promises as fsPromises } from 'fs'
 import path from 'path'
 
 export default async function readFolder(folderPath) {
-  const names = await fsPromises.readdir(folderPath)
+  let names
+  try {
+    names = await fsPromises.readdir(folderPath)
+  } catch (err) {
+    if (err.code === 'ENOENT') {
+      throw new Error('Filepath ' + folderPath + " doesn't exist")
+    }
+    throw err
+  }
+
   const stats = await Promise.all(
     names.map(async name => {
       const targetPath = path.join(folderPath, name)


client/file-browser/get-files.js
```diff
@@ -1,7 +1,10 @@
-import { promises as fsPromises } from 'fs'
+import fs, { promises as fsPromises } from 'fs'
 import path from 'path'
 
 export default async function readFolder(folderPath) {
+  if (!fs.existsSync(folderPath)) {
```
tec27, line 5: Generally avoid `Sync` methods in `fs`, as they'll block the whole process until they return. Alternatives would be to use `fsPromises.access` or (probably better) to put a try/catch around the `fsPromises.readdir` call below and throw this error if `err.code === 'ENOENT'`

tec27 (changes requested):

client/file-browser/get-files.js
williamhammond: 👍, I handled it using `.catch`

client/file-browser/get-files.js
```diff
@@ -2,7 +2,13 @@ import { promises as fsPromises } from 'fs'
 import path from 'path'
 
 export default async function readFolder(folderPath) {
-  const names = await fsPromises.readdir(folderPath)
+  const names = await fsPromises.readdir(folderPath).catch(err => {
```
tec27, line 5: This is inside an `async` function and already being `await`ed so there's no real need to use `.catch` here, would prefer to just use try/catch, i.e.
```js
let names
try {
  names = await fsPromises.readdir(folderPath)
} catch (err) {
  // ...
}

tec27:
Thanks for making those changes!

PR #692, opened 2021-05-31
Kills server on bad redis connection

williamhammond:
If you have a bad redis password or host, you'll see the log message without any other indication of failure. 

```
{"level":30,"time":1622497667609,"pid":18068,"hostname":"DESKTOP-8GER2KL","req":{"id":"ckpd57vxc0002xwuxcvya3agl","method":"GET","url":"/__webpack_hmr","headers":{"host":"localhost:5555","user-agent":"Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:88.0) Gecko/20100101 Firefox/88.0","accept":"text/event-stream","accept-language":"en-US,en;q=0.5","accept-e
ncoding":"gzip, deflate","connection":"keep-alive","referer":"http://localhost:5555/ladder","cookie":"s=ckpa49sxs001o1suxcaiz0h49; s.sig=4sSvYlKGhsYLrBQm-RfWH0zT4C8","pragma":"no-cache","cache-control":"no-cache"}},"res":{"statusCode":500},"err":{"type":"Error","message":"failed with status code 500","stack":"Error: failed with status code 500\n    at
ServerResponse.onResFinished (C:\\Users\\willi\\Code\\ShieldBattery\\node_modules\\pino-http\\logger.js:73:38)\n    at ServerResponse.emit (events.js:388:22)\n    at ServerResponse.emit (domain.js:470:12)\n    at onFinish (_http_outgoing.js:792:10)\n    at callback (internal/streams/writable.js:513:21)\n    at afterWrite (internal/streams/writable.js:4
66:5)\n    at afterWriteTick (internal/streams/writable.js:453:10)\n    at processTicksAndRejections (internal/process/task_queues.js:81:21)"},"responseTime":952,"msg":"request errored"}
{"level":30,"time":1622497667613,"pid":18068,"hostname":"DESKTOP-8GER2KL","reqId":"ckpd57wnx0003xwuxc1uy8xdt","req":{"method":"GET","url":"/engine.io/?clientId=ckpcqf19h00004f67peod6yp9&EIO=4&transport=polling&t=Nd4PQeQ","headers":{"host":"localhost:5555","connection":"keep-alive","accept":"*/*","x-shield-battery-client":"true","user-agent":"Mozilla/5.
0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) ShieldBattery-Local/7.0.12 Chrome/91.0.4472.69 Electron/13.0.1 Safari/537.36","origin":"shieldbattery://app","sec-fe
```

Some context on ioredis error handling https://github.com/luin/ioredis/issues/68. 

I'm worried this might be an over correction, since this will bomb out the server without retrying if there is a transient ENOTFOUND or ECONNREFUSED. Although i really don't know how common that is. 

A few other approaches i thought of are 

Just adding a log message like so, and then relying on external monitoring of redis health in production if that becomes necessary. 
```
redis.on('error', err => {
  console.log(err)
})
```

Or logging the error, building a slightly more comprehensive client that'll track if it's initial startup and fail hard then, and otherwise rely on logging. 

I've been staring at the ioredis docs and code a bit but i can't find a knob to turn that'll say something like throw and error after 30 seconds of failed requests or something like that. We could do that per client call, but we don't have full control over that in the case of the session-store and it'd be pretty annoying to have to remember to do that everywhere.

### DIFF
--- a/server/app.ts
+++ b/server/app.ts
@@ -17,6 +17,7 @@ import logMiddleware from './lib/logging/log-middleware'
 import log from './lib/logging/logger'
 import userIpsMiddleware from './lib/network/user-ips-middleware'
 import { RallyPointService } from './lib/rally-point/rally-point-service'
+import redis from './lib/redis'
 import checkOrigin from './lib/security/check-origin'
 import { cors } from './lib/security/cors'
 import secureHeaders from './lib/security/headers'
@@ -154,6 +155,21 @@ const rallyPointInitPromise = rallyPointService.initialize(
     )
     app.use(koaConvert(koaWebpackHot(getWebpackCompiler())))
   }
+
+  log.info('Testing connection to redis.')
+  try {
+    await redis.ping()
+  } catch (err) {
+    log.error(
+      { err },
+      'Could not connect to Redis instance, redis host/port configuration may be incorrect',
+    )
+    process.exit(1)
+  }
+
+  redis.on('error', err => {
+    log.error({ err }, 'redis error')
+  })
 
   fileStoreMiddleware(app)
 


williamhammond:
Also if you'd prefer I always make a ticket before making a PR please let me know!

tec27:
Not exactly what I was intending :)

I definitely don't want to kill the server on *every* redis error, as there's a lot of possible reasons for those to occur. What I meant was that if your redis config is completely wrong (e.g. we are never able to make a connection), the server ideally would bail out on startup.

Perhaps the right way to accomplish that is by attempting to send something to the redis server during startup, and if that fails, bail out then. (PING)[https://redis.io/commands/ping] seems like it might be a reasonable option for that.

Adding logging on redis errors seems fine too, in addition to that, but you'd want to use our logger, not just put stuff in the console.

williamhammond:
> Or logging the error, building a slightly more comprehensive client that'll track if it's initial startup and fail hard then, and otherwise rely on logging.

yeah I was thinking we could use ping for this. Does it matter if the ping check happens everytime is redis imported or should I take the time to make sure it happens only once? It's kind of odd because I think it would make more sense to check in app.ts but the client is used as a side effect of one of the imports

williamhammond:
Oh actually I might have missed a good solution because i misunderstood how promises worked

williamhammond:
Yeah so I didn't understand that throwing in a promise catch block just returned another promise. 

So my current solution seems reasonable, using `redis.ping` instead of `redis.connect` will let it do it's normal retrying before killing the process so there will be some leeway during startup. It'll run everytime a new client is requested, which seems mostly reasonable since currently there are only about 5 and it looks like they all get setup at boot time. I kind find a way to do the ping just once though if you'd like.

tec27:
For future reference: imported files only ever get executed once, and then the same values are returned for all future imports

server/lib/redis/index.ts
```diff
@@ -1,8 +1,20 @@
...
 // TODO(tec27): provide some better wrapper around this that deals with connects/disconnects, etc.
-export default new Redis({
+const redis = new Redis({
   port: Number(process.env.SB_REDIS_PORT),
   host: process.env.SB_REDIS_HOST,
   dropBufferSupport: true,
 })
+
+redis.on('error', err => {
+  logger.error(err)
```
tec27, line 12: So the way the logger works is that the first parameter passed to it is an object where the keys can have special meaning. Depending on the key it filters out certain properties to log as well, the ones that have special handling are:

 - `err` (for `Error`s)
 - `req` (for `HttpRequest`s)
 - `res` (for `HttpResponse`s)

so it's important that when you're logging things with it, you pass an object that has the right keys. It's also important that you include some message so we have a better idea of what is occurring

So a better call here is probably something like:
```ts
logger.error({ err }, 'redis error')
```

server/lib/redis/index.ts
```diff
@@ -1,8 +1,20 @@
...
+})
+
+redis.ping().catch(err => {
+  logger.error('Error initializing Redis connection' + err)
```
tec27, line 16: Same idea as above, it's better to pass the error on the first parameter so we get better structured data, instead of just a message string:

```ts
logger.error({ err }, 'Error initializing Redis connection')
```

I also think in this case, since this likely indicates a configuration problem, it might ne nice to use a message that hints at what can be done to fix this, e.g. `'Could not connect to Redis instance, redis host/port configuration may be incorrect'`

tec27 (changes requested):
Feels pretty weird to me to put this stuff in this file, I think I'd prefer that error logging + a ping test are configured at the top level instead.

We have an `async` block in `app.ts` that could easily contain both of these things, and would allow you to wait on the redis ping before continuing application initialization (so the crash might be a bit more obvious instead of happening in some random spot):

https://github.com/ShieldBattery/ShieldBattery/blob/master/server/app.ts#L157

williamhammond:
> Feels pretty weird to me to put this stuff in this file, I think I'd prefer that error logging + a ping test are configured at the top level instead.

Yeah good point. I was originally thinking to try and synchronously block until we knew startup was good to get the clearest error message since https://github.com/ShieldBattery/ShieldBattery/blob/b0062d415bee0d4ceb35e6216e6a56a3417b05be/server/lib/session/session-store.js#L4 ends up starting to make requests. But with a good error message and a the process crashing it should give people all the context they need

williamhammond:
Thanks for the feedback I like this solution 100x better 🙏

server/app.ts
```diff
@@ -155,6 +156,21 @@ const rallyPointInitPromise = rallyPointService.initialize(
     app.use(koaConvert(koaWebpackHot(getWebpackCompiler())))
   }
 
+  log.info('Testing connection to redis.')
```
2Pacalypse-, line 159: I would personally remove this line. I don't think it's terribly important to let developers know this is about to happen, and it might confuse them since they'll only see the conclusion of this declaration in case of an error.

But then again, I don't think it hurts _too_ much to have more logging, so I'm willing to change my mind on this if others agree ^^.

williamhammond: I just dropped it in because it'll hang there for a bit if the connection is bad. It was an after thought though, best case it'll just save someone from `ctrl-c`'ing before they get a clear error message if they get impatient

tec27: Yeah, I'm fine with it being there

tec27:
Thank you!

PR #693, opened 2021-06-03
Move the user sockets to use primarily `userId` instead of `userName`.

2Pacalypse-:
All other services have only been updated by a bare minimum to make them
work again. Subsequent PR(s) should update them to use the `userId`s as
well, to make things more internally consistent and easier to work with.

### DIFF
--- a/server/lib/chat/chat-service.ts
+++ b/server/lib/chat/chat-service.ts
@@ -51,8 +51,8 @@ export default class ChatService {
     private userSocketsManager: UserSocketsManager,
   ) {
     userSocketsManager
-      .on('newUser', (userSockets: UserSocketsGroup) => this.handleNewUser(userSockets))
-      .on('userQuit', (userId: number) => this.handleUserQuit(userId))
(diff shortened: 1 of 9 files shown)


server/lib/chat/chat-service.ts
```diff
@@ -48,16 +48,15 @@ export default class ChatService {
 
   constructor(
     private publisher: TypedPublisher<ChatEvent>,
-    private nydus: NydusServer,
     private userSocketsManager: UserSocketsManager,
   ) {
     userSocketsManager
       .on('newUser', (userSockets: UserSocketsGroup) => this.handleNewUser(userSockets))
-      .on('userQuit', (userName: string) => this.handleUserQuit(userName))
+      .on('userQuit', (userId: number) => this.handleUserQuit(userId))
```
tec27, line 55: This type annotation should be unnecessary (the one above it is also unnecessary)

tec27 (approved):

PR #694, opened 2021-06-05
Update scr-analysis to support newer versions

neivv:
Tested that this worked whenplaying alone with the patch from yesterday that got rolled back, but probably would be good to test with the this "good patch" that the game was reverted to before merging this PR. (It is supposed to work with both patches)


tec27:
Tested on the latest patch (lobbies, replays, multiplayer lobbies) and it all seems to work okay 👍

PR #695, opened 2021-06-05
Move the chat stuff away from using usernames internally.

2Pacalypse-:
The main concept that this PR is trying to introduce is the split of user properties in chat channels into two distinct interfaces, typed in code as `ChatUser` and `User`. The idea behind a `ChatUser` type is to include all the user properties that are relevant to a user in a chat channel, e.g.:
```js
interface ChatUser {
  id: number
  name: string
  joinTime: Number
  permissions: ChannelPermissions
  etc.
}
```
The `User` type already exists in the system and is meant to describe general user info, e.g.:
```js
interface User {
  id: number
  name: string
  avatar: string
  permissions: Permissions
  etc.
}
```
The reason why `name` property specifically is included in both types is because it makes things easier, at least currently, since it's needed in the chat reducer to create message records. Eventually this might not be needed anymore.

So in essence, this PR basically updates everything that previously used username to denote a chat user to a proper `ChatUser` type. In addition to including `ChatUser` in various responses to the client from the server, we also include the general `User` info type in particular requests and events, which are:
- ~~`init` websocket event~~
- `join` websocket event
- `GET /chat/:channelName/users` HTTP request

~~I feel like including user info in the `init` event is not even necessary since the other two cases should cover it. This would simplify the server code a bit too, so it'd be nice if we could remove it.~~ This is now removed.

Future work:
- make the usernames in chat messages clickable that open a user profile overlay when clicked


2Pacalypse-:
Depends on #705

2Pacalypse-:
Reading the PR description again, `ChatUser` interface might not be necessary at all and could be replaced with just a `userId`, with additional properties sent alongside with it (for now that would only be `name` since it's needed in the chat reducer for message records, but could be removed once the message records are updated to work with `userId`s as well).

tec27 (approved):

PR #696, opened 2021-06-06
Handle pg syntax error

williamhammond:
This is still a work in progress but wanted to share some work because I have some questions. 

1.) I've been googling everywhere and I can't find a way other than using something like https://www.npmjs.com/package/verror to get a full stack trace from where the query is made back to where it's parsed. I'm not sure if it's a super big deal though. 
2.) I kind of dislike relying on postgres' convention of using the first two values of the error code to show what class the error is. Would typing the errors by class, or building an enum of classes be overkill?  I don't think most error types will require a ton of customization 
3.) I was kind of thinking that hijacking the reject error of the Promise returned by the wrappedclient to do the parsing, and letting the caller handle it properly would be best, but I couldn't find a way to get to the error without the `.catch` on the promise in the db index file. I think I'm doing the correct thing but let me know since I'm still wrapping my head around nodes async model. 

Even on the off chance you all think everything is good, I still plan on incorporating the the `hint` value sometimes returned, and writing some unit tests for the error parsing. And I need to either pass long the rest of the pg error with my new error or be more opinionated to handle stuff like this https://github.com/ShieldBattery/ShieldBattery/blob/master/server/lib/api/users.js#L106-L110. 

p.s.. I can keep the .idea/ line in my personal git ignore if you like. I moved the pino config because that's what it said in the docs.

### DIFF
--- a/server/lib/db/index.ts
+++ b/server/lib/db/index.ts
@@ -1,6 +1,7 @@
 import pg from 'pg'
 import { isTestRun } from '../../../common/is-test-run'
 import log from '../logging/logger'
+import handlePgError from './pg-error-handler'
 
 /**
  * The amount of time queries are allowed to run for before they are considered "slow" and logged
@@ -63,7 +64,9 @@ export class DbClient {
     const queryText = isQueryConfig(queryTextOrConfig) ? queryTextOrConfig.text : queryTextOrConfig
     const startTime = Date.now()
     try {
-      return this.wrappedClient.query(queryTextOrConfig, values)
+      return await this.wrappedClient.query(queryTextOrConfig, values)
+    } catch (error) {
+      throw handlePgError(queryText, error)
     } finally {
       const totalTime = Date.now() - startTime
       if (totalTime > SLOW_QUERY_TIME_MS) {
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@ app/styles/
 dist/
 node_modules/
 server/.env
+.idea/
 .env
 npm-debug.log
 yarn-error.log
--- a/server/lib/db/pg-error-handler.ts
+++ b/server/lib/db/pg-error-handler.ts
@@ -1,0 +1,50 @@
+import { DatabaseError } from 'pg'
+import { SYNTAX_ERROR_OR_ACCESS_RULE_VIOLATION } from './pg-error-codes'
+
+export default function handlePgError(query: string, error: Error): Error {
+  if (
+    isDatabaseError(error) &&
+    error.code &&
+    getErrorClass(error.code) === getErrorClass(SYNTAX_ERROR_OR_ACCESS_RULE_VIOLATION)
+  ) {
+    return handlePgSyntaxError(query, error)
+  } else {
+    return error
+  }
+}
+
+function getErrorClass(code: string): string {
+  // By the standard, the first two character denote the class of error
+  // https://www.postgresql.org/docs/13/errcodes-appendix.html
+  return code.substring(0, 2)
+}
+
+function isDatabaseError(error: Error): error is DatabaseError {
+  return error.hasOwnProperty('code') && error.hasOwnProperty('position')
+}
+
+function handlePgSyntaxError(queryText: string, error: DatabaseError): Error {
+  const messageLines = queryText.split('\n')
+  const position = Number(error.position)
+
+  if (!isNaN(position)) {
+    let curLine = 0
+    let summedLength = 0
+    do {
+      summedLength += messageLines[curLine].length + 1 // 1 extra for the \n
+      curLine += 1
+    } while (curLine < messageLines.length && summedLength < position)
+
+    if (summedLength >= position) {
+      const foundLine = curLine - 1
+      const relativePosition = position - (summedLength - messageLines[foundLine].length)
+      const caretLine = '^'.padStart(relativePosition + 1)
+      messageLines.splice(curLine, 0, caretLine)
+    }
+  }
+
+  if (error.hint) {
+    messageLines.push(`hint: ${error.hint}`)
+  }
+  return new Error(`${error.message} in query: \n${messageLines.join('\n')}`)
+}


server/lib/logging/logger.ts
```diff
@@ -8,5 +8,6 @@ export function getLoggerOptions() {
     level: process.env.NODE_ENV === 'production' ? 'info' : 'debug',
     genReqId: cuid,
     serializers: stdSerializers,
+    prettyPrint: { colorize: true, translateTime: true },
```
tec27, line 11: I think if you do things this way you're likely to break the stuff we do on the real servers (and I think this probably breaks the database logging too?). Would prefer to keep this as it is

server/lib/db/pg-error-parser.ts
```diff
@@ -0,0 +1,42 @@
+import { SYNTAX_ERROR_OR_ACCESS_RULE_VIOLATION } from './pg-error-codes'
+
+export default function parseError(query: string, error: any): Error {
```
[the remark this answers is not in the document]
tec27: I also think `parse` is a sort of weird naming for both of the functions in here, as neither of them is really doing any parsing. `handle` or `maybeHandle` is probably a more suitable prefix

server/lib/db/pg-error-parser.ts
```diff
@@ -0,0 +1,42 @@
...
+  let lineNumber = 0
+  let relativePosition = 0
+  let foundPosition = false
+  for (let i = 0; i < queryLines.length && !foundPosition; i++) {
+    lineNumber += 1
+    // We need to count the new line character
+    cursor += 1
+    const line = queryLines[i]
+    for (let j = 0; j < line.length; j++) {
+      if (cursor === parseInt(error.position)) {
```
tec27, line 27: prefer `Number(...)` to `parseInt(...)`, `parseInt` has a bunch of unexpected gotchas. It'd also be good to pull this conversion out of the loops, since it will be the same result every time.

server/lib/db/pg-error-parser.ts
```diff
@@ -0,0 +1,42 @@
...
+  let queryLines = queryText.split('\n')
+
+  let cursor = 0
+  let lineNumber = 0
+  let relativePosition = 0
+  let foundPosition = false
+  for (let i = 0; i < queryLines.length && !foundPosition; i++) {
+    lineNumber += 1
+    // We need to count the new line character
+    cursor += 1
```
tec27, line 24: This seems like it'd be an off-by-one thing because the lines do not start with a newline, they end with it.

This whole logic seems a bit convoluted for what you're actually trying to do here, as well. We know the position is within the range `[0, length]`, we want to know what line it's on. Given an array of lines (which you got above from the `split`), you can iterate through the lines until the summed length is `> error.position`. Once that is the case, the position is on the previous line.

```ts
const position = Number(error.position)
if (isNaN(position)) {
  return new Error(`${error.message} in query: \n${queryLines.join('\n')}`)
}

let curLine = 0
let summedLength = 0
do {
  summedLength += queryLines[curLine].length
  curLine += 1
} while (curLine < queryLines.length && summedLength < position)

if (summedLength >= position) {
  const foundLine = curLine - 1
  const relativePosition = position - (summedLength - queryLines[foundLine].length)
  const caretLine = '^'.padStart(relativePosition + 1)
  queryLines.splice(curLine, 0, caretLine)
}

return new Error(`${error.message} in query: \n${queryLines.join('\n')}`)
```

tec27 (changes requested):

server/lib/db/index.ts
```diff
@@ -63,7 +64,9 @@ export class DbClient {
     const queryText = isQueryConfig(queryTextOrConfig) ? queryTextOrConfig.text : queryTextOrConfig
     const startTime = Date.now()
     try {
-      return this.wrappedClient.query(queryTextOrConfig, values)
+      return this.wrappedClient.query(queryTextOrConfig, values).catch(error => {
```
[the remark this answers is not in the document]
williamhammond: I'm not quite sure what I'm doing wrong, but I changed this to use a regular `catch` block and regardless of it was just `try/finally` or `try/catch/finally` in the calling methods it'd never hit the catch block in here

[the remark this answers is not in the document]
tec27: You need to change it to a `return await ...` probably. Without the `await` it just returns the promise without waiting for it to resolve, thus errors are handled after this function has already completed. (Honestly the fact that it wasn't already an `await` is a mistake and means this timing code doesn't actually measure the query time in the current form)

server/lib/db/pg-error-parser.ts
```diff
@@ -0,0 +1,37 @@
...
+    return handlePgSyntaxError(query, error)
+  }
+  return error
+}
+
+function handlePgSyntaxError(queryText: string, error: DatabaseError): DatabaseError {
+  const queryLines = queryText.split('\n')
+  const position = Number(error.position)
+  if (isNaN(position)) {
+    error.message = `${error.message} in query: \n${queryLines.join('\n')}`
```
tec27, line 18: Don't think I'm really a fan of doing this, we don't have any guarantees that this field is writable (or even a field, it could just be a getter). Would prefer to return new Errors rather than mutating existing ones.

tec27 (changes requested):

server/lib/db/pg-error-parser.ts
williamhammond: Do you think it makes sense to take make a deep copy to preserve all the DatabaseError specific fields or do you think it makes sense to pick and choose?

tec27: I would just make a new error and not worry about anything that is present on the one we received. It won't be logged anyway, and if we're handling it here, the relevant thing is "the syntax of this SQL statement is incorrect", and the message already contains the relevant context for that.

tec27 (approved):

tec27:
Thank you!

PR #698, opened 2021-06-09
Move the whispers service to use a HTTP-based API.

2Pacalypse-:


server/lib/whispers/whisper-api.ts
```diff
@@ -0,0 +1,171 @@
...
+    case WhisperServiceErrorCode.UserOffline:
+    case WhisperServiceErrorCode.UserNotFound:
+      throw new httpErrors.NotFound(err.message)
+    case WhisperServiceErrorCode.InvalidCloseAction:
+    case WhisperServiceErrorCode.InvalidGetSessionHistoryAction:
+      throw new httpErrors.BadRequest(err.message)
+    case WhisperServiceErrorCode.NoSelfMessaging:
+      throw new httpErrors.Forbidden(err.message)
+    default:
+      assertUnreachable(err.code)
```
tec27, line 58: Didn't really think about this in the other places you've done this, but this isn't *really* safe because `code` doesn't necessarily guarantee that the thing we have here is actually a WhisperServiceError. Might be good to add a Symbol or something on the prototype that you can check in `isWhisperServiceError` instead, and then this assertion wouldn't be replacing e.g. errors from `fs`.

(Alternative to the Symbol approach if you know the classes will always be constructed WhisperServiceErrors, which seems reasonable, is simply to do an `instanceof WhisperServiceError` check)

tec27 (approved):

PR #701, opened 2021-06-13
Change Userlist to use react-window - Partially fixes #690

### LINKED ISSUE #690: Move react-virtualized usages to react-windowed
It'd be nice to not pull in two libraries for this same purpose, and react-windowed is better suited to our purposes, along with not being based around inline styles and specific class names for customization.

GrantJenk:
Changed:
* Changed Userlist from being a class component to using hooks
* Refactored the logic to render the rows and get height
* Removed padding from being a row that's rendered to padding on the container

### DIFF
--- a/client/chat/channel.tsx
+++ b/client/chat/channel.tsx
@@ -261,7 +261,7 @@ const UserList = React.memo((props: UserListProps) => {
 
   useMemo(() => {
     listRef.current?.resetAfterIndex(0, false)
-  }, [getRowHeight])
+  }, [getRowHeight]) // eslint-disable-line react-hooks/exhaustive-deps
 
   return (
     <UserListContainer ref={dimensionsRef}>
--- a/client/chat/channel.tsx
+++ b/client/chat/channel.tsx
@@ -167,13 +167,6 @@ const UserListEntry = React.memo<UserListEntryProps>(props => {
   )
 })
 
-const UsersVirtualizedList = styled(VariableSizeList)`
-  &:focus,
-  & > div:focus {
-    outline: none;
-  }
-`
-
 interface UserListProps {
   users: UsersRecord
 }
@@ -181,12 +174,10 @@ interface UserListProps {
 const UserList = React.memo((props: UserListProps) => {
   const { active, idle, offline } = props.users
   const [dimensionsRef, containerRect] = useObservedDimensions()
-  const containerRef = useRef<HTMLElement | null>(null)
   const listRef = useRef<VariableSizeList | null>(null)
   const containerCallback = useCallback(
     (node: HTMLElement | null) => {
       dimensionsRef(node)
-      containerRef.current = node
     },
     [dimensionsRef],
   )
@@ -281,15 +272,15 @@ const UserList = React.memo((props: UserListProps) => {
 
   return (
     <UserListContainer ref={containerCallback}>
-      <UsersVirtualizedList
+      <VariableSizeList
         ref={listRef}
-        style={{ height: 'auto', display: 'inline-block', overflow: 'unset' }}
+        style={{ overflowX: 'hidden' }}
         width='100%'
         height={containerRect?.height ?? 0}
         itemCount={rowCount}
         itemSize={getRowHeight}>
         {renderRow}
-      </UsersVirtualizedList>
+      </VariableSizeList>
     </UserListContainer>
   )
 })
--- a/client/chat/channel.tsx
+++ b/client/chat/channel.tsx
@@ -1,9 +1,9 @@
-import React, { useCallback, useEffect, useRef, useState } from 'react'
-import { List as VirtualizedList, ListRowRenderer } from 'react-virtualized'
+import React, { useCallback, useEffect, useMemo, useRef, useState } from 'react'
+import { VariableSizeList } from 'react-window'
 import styled, { css } from 'styled-components'
 import { MULTI_CHANNEL } from '../../common/flags'
 import Avatar from '../avatars/avatar'
-import WindowListener from '../dom/window-listener'
+import { useObservedDimensions } from '../dom/dimension-hooks'
 import Chat from '../messaging/chat'
 import { Message } from '../messaging/message-records'
 import { push } from '../navigation/routing'
@@ -40,6 +40,9 @@ const UserListContainer = styled.div`
   width: 256px;
   flex-grow: 0;
   flex-shrink: 0;
+  overflow-y: auto;
+  overflow-x: hidden;
+  padding-bottom: 8px;
 
   background-color: ${background700};
 `
@@ -119,13 +122,6 @@ const UserListEntryItem = styled.div<UserListEntryItemProps>`
   }}
 `
 
-const USER_LIST_PADDING_HEIGHT = 8
-
-const UserListPadding = styled.div`
-  width: 100%;
-  height: ${USER_LIST_PADDING_HEIGHT}px;
-`
-
 const UserListName = styled.span`
   ${singleLine};
   display: inline-block;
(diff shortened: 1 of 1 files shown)


client/chat/channel.tsx
```diff
@@ -182,140 +178,121 @@ interface UserListProps {
...
+    [dimensionsRef],
+  )
 
-  componentDidUpdate(prevProps: UserListProps) {
-    if (prevProps.users !== this.props.users) {
-      this._listRef.current?.recomputeRowHeights()
-    }
-  }
+  useEffect(() => {
+    listRef.current?.resetAfterIndex(0, false)
```
tec27, line 195: It seems to me like if this is necessary, this is actually too late to be doing this (and the `false` second parameter here means the cache wouldn't be updated until a follow-on render call or if the UI was scrolled?).

I think the right place to call this would be directly during the render, so that react-window will re-request sizes in its own render

client/chat/channel.tsx
```diff
@@ -182,140 +178,121 @@ interface UserListProps {
...
-  _contentRef = React.createRef<HTMLDivElement>()
-  _listRef = React.createRef<VirtualizedList>()
-
-  componentDidMount() {
-    this.updateDimensions()
-  }
+const UserList = React.memo((props: UserListProps) => {
+  const { active, idle, offline } = props.users
+  const [dimensionsRef, containerRect] = useObservedDimensions()
+  const containerRef = useRef<HTMLElement | null>(null)
```
tec27, line 184: This is never used for anything, so I think you can remove it and just use `dimensionsRef` directly instead of doing the multiplexing function ref thing

client/chat/channel.tsx
```diff
@@ -182,140 +178,121 @@ interface UserListProps {
...
-    if (this.state.width !== width || this.state.height !== height) {
-      this.setState({ width, height })
-    }
-  }
-}
+  return (
+    <UserListContainer ref={containerCallback}>
+      <UsersVirtualizedList
+        ref={listRef}
+        style={{ height: 'auto', display: 'inline-block', overflow: 'unset' }}
```
tec27, line 286: I feel like this probably isn't necessary. I did this on the ladder table because we want the scrollbar to be on the root of the content, rather than on the table, but the Users list has its own scrollbar anyway.

client/chat/channel.tsx
```diff
@@ -171,7 +167,7 @@ const UserListEntry = React.memo<UserListEntryProps>(props => {
   )
 })
 
-const UsersVirtualizedList = styled(VirtualizedList)`
+const UsersVirtualizedList = styled(VariableSizeList)`
```
tec27, line 170: These styles were to override something that react-virtualized set, shouldn't be necessary with the new library.

tec27 (changes requested):

client/chat/channel.tsx
GrantJenk: I added this because the `Overline`s in the `UserList` would shrink to the size of a `UserListEntry` if a user logged in/out. I initially left out the `false`, but it seems to be working now when I set it to `true`? Not sure what I missed there.

By calling directly during the render, do you mean something like this?
```
  return (
    <UserListContainer ref={containerCallback}>
      {listRef.current?.resetAfterIndex(0)}
      <VariableSizeList
```
If so, I get an error `Warning: Cannot update during an existing state transition`

client/chat/channel.tsx
GrantJenk: :+1: I added in a `overflowX: 'hidden'` to hide the scrollbar that appears.

client/chat/channel.tsx
tec27: By "directly in render" I just mean at some point in the component function (putting it in the JSX part is weird but probably works the same, although I'm not entirely certain on that). Passing `true` as the second param likely forces a second render on every render, so I think it would re-render over and over again forever? Unsure to be honest. I'll have to explore more, but from looking at their code before, calling it in a `useEffect` doesn't seem like the right thing to me.

Honestly I'd prefer they just let us turn off the caching entirely, as this lookup is likely very cheap for this usecase anyway =/

GrantJenk: Ahh, I gotchya. Also, looks like the error I was getting was actually from the `true` parameter, and it only appeared once on loading the page.

tec27: Ah okay, that makes sense I think because the `true` parameter basically means "re-render now", and React probably doesn't like doing that if it's currently rendering :)

client/chat/channel.tsx
```diff
@@ -171,151 +167,121 @@ const UserListEntry = React.memo<UserListEntryProps>(props => {
...
-  componentDidUpdate(prevProps: UserListProps) {
-    if (prevProps.users !== this.props.users) {
-      this._listRef.current?.recomputeRowHeights()
-    }
-  }
+const UserList = React.memo((props: UserListProps) => {
+  const { active, idle, offline } = props.users
+  const [dimensionsRef, containerRect] = useObservedDimensions()
+  const listRef = useRef<VariableSizeList | null>(null)
+  const containerCallback = useCallback(
```
tec27, line 178: What I meant here is that you don't actually need this callback at all, you can simply pass `dimensionsRef` as the `ref` on the component. When we're creating a function like this in other places, it's so that we can store a `ref` value in 2 or more places.

client/chat/channel.tsx
```diff
@@ -171,151 +167,121 @@ const UserListEntry = React.memo<UserListEntryProps>(props => {
...
 
-    let i = index - (active.size + 1)
-    if (idle.size) {
-      if (i === 0) {
-        return OVERLINE_HEIGHT
-      } else if (i < idle.size + 1) {
-        return USER_ENTRY_HEIGHT
+  const getRowHeight = useCallback(
+    (index: number) => {
+      if (index > rowCount) {
```
tec27, line 192: Should be `>=` I think, since it's 0-indexed

client/chat/channel.tsx
```diff
@@ -171,151 +167,121 @@ const UserListEntry = React.memo<UserListEntryProps>(props => {
...
-          width={this.state.width}
-          height={this.state.height}
-          rowCount={rowCount}
-          rowHeight={this.getRowHeight}
-          rowRenderer={this.renderRow}
-        />
-      </UserListContainer>
-    )
-  }
+  listRef.current?.resetAfterIndex(0, false)
```
tec27, line 269: Might be good to memoize this (I think a good candidate for the `deps` to use would be `getRowHeight`, since the values for row height can only change if that's changed)

client/chat/channel.tsx
```diff
@@ -40,6 +40,9 @@ const UserListContainer = styled.div`
   width: 256px;
   flex-grow: 0;
   flex-shrink: 0;
+  overflow-y: auto;
+  overflow-x: hidden;
+  padding-bottom: 8px;
```
tec27, line 45: I don't think this is the right place to add the padding. If you add it here, you end up with a 8px background-colored bar at the bottom that cuts off the content (the list of users). What we *actually* want is 8px of extra space at the end of the content (only visible when you scroll all the way to the bottom). To get that, you can probably add padding to the `VariableSizeList` (although I can't remember if that actually works with `react-window`, may need to be applied to some deeper element)

tec27 (changes requested):

client/chat/channel.tsx
GrantJenk: The linter warns that `getRowHeight` is an unnecessary dependency and to remove it

tec27: You can add a `// eslint-disable-next-line` comment to disable that particular warning (VSCode should have an action to do this on the offending line)

tec27 (approved):

tec27:
Thank you!

PR #704, opened 2021-06-15
Add types to whisper actions and events.

2Pacalypse-:


common/whispers.ts
```diff
@@ -0,0 +1,73 @@
...
+
+export interface WhisperMessage {
+  id: string
+  from: WhisperUser
+  to: WhisperUser
+  sent: number
+  data: WhisperMessageData
+}
+
+// TODO(2Pac): Make this into an interface and include more information here
```
tec27, line 24: not sure that's really desirable with where you're using this tbh, feels like this should just be an id in those places and any extra info about the users should be passed adjacently

PR #705, opened 2021-06-15
Move the whisper models to its feature folder and add types to it

2Pacalypse-:


2Pacalypse-:
Depends on https://github.com/ShieldBattery/ShieldBattery/pull/704

server/lib/whispers/whisper-models.ts
```diff
@@ -0,0 +1,167 @@
...
+
+export interface WhisperSessionEntry {
+  targetUserId: number
+  targetUserName: string
+}
+
+export async function getWhisperSessionsForUser(userId: number): Promise<WhisperSessionEntry[]> {
+  const { client, done } = await db()
+  try {
+    const result = await client.query(sql`
```
tec27, line 13: Generally you always want to pass a generic parameter to `client.query`, otherwise it's `any` and that's bad for typing anything further than that in the function.

Apply throughout this file

server/lib/whispers/whisper-models.ts
```diff
@@ -0,0 +1,167 @@
...
+    if (result.rowCount < 1) {
+      throw new Error('No rows deleted')
+    }
+  } finally {
+    done()
+  }
+}
+
+export interface DbWhisperMessage {
+  msgId: string
```
tec27, line 71: In these structures I would try *very hard* to match what the names are in the database (with camel case instead of snake case, obviously). If you start changing stuff it makes using `Dbify` hard/impossible and you're gonna have a bad time

tec27 (changes requested):

server/lib/whispers/whisper-models.ts
```diff
@@ -0,0 +1,172 @@
...
+    SELECT *
+    FROM messages
+    ORDER BY sent ASC;
+  `)
+
+  try {
+    const result = await client.query<DbWhisperMessage>(query)
+
+    return result.rows.map(row => ({
+      id: row.id,
```
tec27, line 163: Would prefer that this conversion is pulled out into a function if you're doing it in multiple places

tec27 (approved):

PR #706, opened 2021-06-21
Move the whispers stuff away from using usernames internally.

2Pacalypse-:
Whispers are kind of in a weird place because we still want to use the
usernames in all the routing-related paths, which means it's still
preferable to use the usernames in various data structures (e.g. in the
reducer).

To be honest, we could probably get away with using usernames everywhere
(which makes this PR kinda useless huh), since we only need user IDs in
a couple of places, like in each whisper message.


2Pacalypse-:
Depends on #695

tec27 (approved):

PR #707, opened 2021-06-21
Make the username in messaging-related components clickable.

2Pacalypse-:
Clicking the username will open the profile overlay for the clicked
user.

NOTE 1: There's no hover effect atm besides changing the cursor to
pointer. Might be worth it to add some color change as well.

NOTE 2: There's some weird ass bug atm where hovering on some names
doesn't make the whole name clickable, only parts of it.

### DIFF
--- a/client/messaging/connected-username.tsx
+++ b/client/messaging/connected-username.tsx
@@ -5,7 +5,10 @@ import { ConnectedUserProfileOverlay } from '../profile/user-profile-overlay'
 import { useAppSelector } from '../redux-hooks'
 
 const Username = styled.span`
-  cursor: pointer;
+  &:hover {
+    cursor: pointer;
+    text-decoration: underline;
+  }
 `
 
 /**
--- a/client/messaging/connected-username.tsx
+++ b/client/messaging/connected-username.tsx
@@ -1,0 +1,55 @@
+import React, { useCallback, useRef, useState } from 'react'
+import styled from 'styled-components'
+import { ConnectedUserProfileOverlay } from '../profile/user-profile-overlay'
+import { useAppSelector } from '../redux-hooks'
+
+const Username = styled.span`
+  cursor: pointer;
+`
+
+/**
+ * A component which displays the username in messaging-related services (e.g. chat, lobbies,
+ * whispers) and any additional functionality that we might want to do with them, like opening a
+ * user profile when clicked.
+ *
+ * This component is connected to the store where it tries to find the user. All the services using
+ * it should ensure that the user is loaded in the store properly.
+ */
+export function ConnectedUsername(props: { userId: number }) {
+  const [overlayOpen, setOverlayOpen] = useState(false)
+  const usernameRef = useRef(null)
+
+  const onOpenOverlay = useCallback(() => {
+    setOverlayOpen(true)
+  }, [])
+  const onCloseOverlay = useCallback(() => {
+    setOverlayOpen(false)
+  }, [])
+
+  const user = useAppSelector(s => s.users.byId.get(props.userId))
+  if (!user) {
+    return <span>Unknown user</span>
+  }
+
+  return (
+    <>
+      <ConnectedUserProfileOverlay
+        key={'overlay'}
+        userId={user.id}
+        popoverProps={{
+          open: overlayOpen,
+          onDismiss: onCloseOverlay,
+          anchor: usernameRef.current,
+          anchorOriginX: 'right',
+          anchorOriginY: 'top',
+          anchorOffsetX: 4,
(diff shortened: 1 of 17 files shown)


2Pacalypse-:
Depends on #706

client/messaging/connected-username.tsx
```diff
@@ -0,0 +1,55 @@
...
+  }, [])
+
+  const user = useAppSelector(s => s.users.byId.get(props.userId))
+  if (!user) {
+    return <span>Unknown user</span>
+  }
+
+  return (
+    <>
+      <ConnectedUserProfileOverlay
```
tec27, line 36: I worry a *lot* about introducing this here tbh. This is already kind of an issue with the user list, every single entry there introduces a Portal + managed div outside the react root, so this means every single message will do this as well (and messages are not virtualized so potentially this is a *lot* of elements to be managing). Would vastly prefer that this communicates with some other component (whether via context or redux actions or whatever) so we only maintain a single element for the whole containing view

client/messaging/message-layout.tsx
```diff
@@ -44,6 +44,8 @@ const longTimestamp = new Intl.DateTimeFormat(navigator.language, {
 export const Separator = styled.i`
   position: absolute;
   display: inline-block;
+  top: -10000px;
```
tec27, line 47: I think doing this screws up the copy-pasteability, I'm fairly certain I tried this before and it did. So, test that before adding this randomly please :)

client/profile/user-profile-overlay.tsx
```diff
@@ -1,14 +1,20 @@
...
-import { Popover, useAnchorPosition } from '../material/popover'
+import { OriginX, OriginY, Popover, useAnchorPosition } from '../material/popover'
 import { useAppSelector } from '../redux-hooks'
 import { navigateToWhisper } from '../whispers/action-creators'
 import { Actions, Header, PopoverContents, StyledAvatar, Username } from './profile-overlay-content'
 
 export interface PopoverRelatedProps {
   open: boolean
   anchor: HTMLElement | null
+  anchorOriginX: OriginX
```
tec27, line 12: is there a reason we can't just use `PopoverProps` here? Maybe `PopoverProps` with an `Omit` if you really need to remove something? (but it seems like they're the same, just some stuff is missing because it didn't get added)

Also get rid of the `anchorOffset` stuff please, that stuff was a bad idea imo and it's a lot better to do the math at the point that is setting the position

tec27 (changes requested):

client/messaging/connected-username.tsx
2Pacalypse-: I tried to do this, and turns out it's not that easy at all. Unless I'm doing something wrong, which is also very likely.

So I tried doing this in the same way we're opening dialogs, through the store. My initial thinking was to save the popover props in the store and then have a connected popover somewhere in the app that would pull those props from the store and render the popover. However, it's pretty much impossible to make it work like this unless we watch for `anchorX/Y` changes in the component that calculates them, and we update the store each time they change.

The other option is calculating the anchor position inside the connected popover itself, but then we'd need to save in store enough information to make it possible to calculate (along with potential offsets), which I'm not a fan of at all.

So yeah, I'm kinda stuck here and not sure what's the best way to deal with it.

Another option might be trying to see if it's possible to change the Popover/Portal code to not create `div` elements _until_ the Popover is actually opened.

client/messaging/message-layout.tsx
2Pacalypse-: I did test it and didn't notice any issues from what I can tell.

client/messaging/connected-username.tsx
tec27: I think Context is actually a better/easier API to do this with since you can pass non-serializable objects in, and have some easier JS API on top of it. If you use keys "properly" (not sure how easy this is to do with our current popover code, but ideally it would be simple :) ) then you could get proper show/hide animation even when people click on different users in succession.

> Another option might be trying to see if it's possible to change the Popover/Portal code to not create div elements until the Popover is actually opened.
This was another suggestion I was thinking of last night, but I'm not 100% sure if this will break things (especially worried about LegacyPopover, but ideally we would finish migrating those, so maybe if it does break those we could just do that first).

If we get to a place where this stuff only renders when it's "in use" then I'm cool with that. I think it might be pretty hard to do that without the previous mentioned approaches though, purely because of the close animations (which effectively require the transition parent to stay mounted). Perhaps react-spring allows us to remove the parent post-animation easily, though?

tec27 (approved):

Issue #276, opened 2017-04-01
Add an option to start our application automatically when OS starts.


2Pacalypse-:
Maybe we could use this library to do it more easily - https://www.npmjs.com/package/auto-launch

2Pacalypse-:
Looks like this is supported by Electron out of the box now and seems to be fairly straightforward to implement:
https://www.electronjs.org/docs/api/app#appsetloginitemsettingssettings-macos-windows

We would also need an option in system tray menu to toggle this option though (perhaps even one in the settings dialog?).

tec27:
Settings dialog seems like a better spot for it, yeah :)

PR #708, opened 2021-06-23
Add setting to open on boot

williamhammond:
This closes https://github.com/ShieldBattery/ShieldBattery/issues/276

I was also thinking about the more verbose `Run Shield Battery when my computer starts.` which is effectively what steam's equivalent setting is. Also this defaults to not opening on start, i'm not sure what your preference is.

### DIFF
--- a/app/app.ts
+++ b/app/app.ts
@@ -25,7 +25,6 @@ const ipcMain = new TypedIpcMain()
 
 getUserDataPath()
 app.setAppUserModelId('net.shieldbattery.client')
-
 autoUpdater.logger = logger
 // We control the download ourselves to avoid problems with double-downloading that this library
 // sometimes has
@@ -151,6 +150,11 @@ function setupIpc(localSettings: LocalSettings, scrSettings: ScrSettings) {
     })
 
   localSettings.on('change', settings => {
+    if (settings.runAppAtSystemStart !== app.getLoginItemSettings().openAtLogin) {
+      app.setLoginItemSettings({
+        openAtLogin: settings.runAppAtSystemStart,
+      })
+    }
     TypedIpcSender.from(mainWindow?.webContents).send('settingsLocalChanged', settings)
   })
   scrSettings.on('change', settings => {
--- a/app/settings.ts
+++ b/app/settings.ts
@@ -7,7 +7,7 @@ import { TypedEventEmitter } from '../common/typed-emitter'
 import { findInstallPath } from './find-install-path'
 import log from './logger'
 
-const VERSION = 6
+const VERSION = 7
 const SCR_VERSION = 3
 
 async function findStarcraftPath() {
@@ -150,6 +150,7 @@ export class LocalSettings extends Settings<LocalSettingsData> {
   private async createDefaults(): Promise<LocalSettingsData> {
     return {
       version: VERSION,
+      runAppAtSystemStart: true,
       starcraftPath: await findStarcraftPath(),
       winX: -1,
       winY: -1,
@@ -211,6 +212,11 @@ export class LocalSettings extends Settings<LocalSettingsData> {
       delete (newSettings as any).displayMode
       delete (newSettings as any).maintainAspectRatio
       delete (newSettings as any).mouseSensitivity
+    }
+
+    if (!settings.version || settings.version < 7) {
+      log.verbose('Found settings version 6, migrating to version 7')
+      newSettings.runAppAtSystemStart = true
     }
 
     newSettings.version = VERSION
--- a/client/settings/app-settings.jsx
+++ b/client/settings/app-settings.jsx
@@ -7,7 +7,9 @@ import SubmitOnEnter from '../forms/submit-on-enter'
 import PlayIcon from '../icons/material/play_arrow-24px.svg'
 import StopIcon from '../icons/material/stop-24px.svg'
 import { TextButton } from '../material/button'
+import CheckBox from '../material/check-box'
 import Slider from '../material/slider'
+import { FormContainer } from './settings-content'
 
 const VolumeSettings = styled.div`
   display: flex;
@@ -39,7 +41,7 @@ class AppForm extends React.Component {
   }
 
   render() {
-    const { bindCustom, isPlayingTestSound, onTestSoundClick, onSubmit } = this.props
+    const { bindCustom, isPlayingTestSound, onTestSoundClick, onSubmit, bindCheckable } = this.props
     const testSoundLabel = isPlayingTestSound ? (
       <>
         <StopIcon />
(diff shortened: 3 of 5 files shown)


williamhammond:
shieldbattery

app/app.ts
```diff
@@ -108,6 +107,9 @@ function setupIpc(localSettings: LocalSettings, scrSettings: ScrSettings) {
     })
     .on('settingsLocalMerge', (event, settings) => {
       // This will trigger a change if things changed, which will then emit a LOCAL_SETTINGS_CHANGED
+      app.setLoginItemSettings({
```
williamhammond, line 110: I'm not sure if there's a better place to do this, It'd be most natural code organization wise to put this at the top of the file with the other electron settings but i'm not sure that's possible since it relies on an ipc call

williamhammond:
To test I rebooted without ever touching the box, rebooted after setting it, and rebooted after unchecking. To test the use of the FormContainer I just changed the size of the main window to max/min. I verified i didn't break the sound bar along the way

app/app.ts
2Pacalypse-: A better place for this would probably be in the `change` event handler that the local settings emit (we're listening to it in the few lines below). That event is emitted only when something actually changes in the settings file, and is emitted if someone manually changes their settings file.

Also, I'm not sure about the performance implications of running this each time _any_ setting changes. I assume it's simply writing something in the registry (at least on Windows machines), so it shouldn't be too bad. Alternatively, you could call the [getLoginItemSettings()](https://www.electronjs.org/docs/api/app#appgetloginitemsettingsoptions-macos-windows) and check if this setting actually changed before trying to set it.

app/settings.ts
```diff
@@ -150,6 +150,7 @@ export class LocalSettings extends Settings<LocalSettingsData> {
   private async createDefaults(): Promise<LocalSettingsData> {
     return {
       version: VERSION,
+      openOnStart: false,
```
2Pacalypse-, line 153: This is not enough to do when adding new settings. This only creates the default values if the settings file doesn't exist, i.e. when users first install the application. Additionally, we need to add new settings to the existing setting files for people who already have the application installed.

You can easily do this by increasing the `VERSION` (for local settings) number in this file and then adding another block in the `migrateOldSettings` function which checks for that version number and adds the new setting.

2Pacalypse- (changes requested):
I think we should definitely be a bit more verbose with the settings label. Your suggestion of `Run ShieldBattery when my computer starts` sounds good to me, but I'll let native English speakers sign off on that ^^.

As for the default value, I feel like it should be set to `true` instead. Not sure how obtrusive that is deemed by general public in this day and age, but it's very easily changeable to whatever the user wants to set it to, so I think we'll be in the clear.

common/local-settings.ts
```diff
@@ -16,6 +16,7 @@ export interface ShieldBatteryAppSettingsData {
 }
 
 export interface LocalSettingsData extends ShieldBatteryAppSettingsData {
+  openOnStart: boolean
```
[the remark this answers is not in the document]
williamhammond: I think i'm going to do `openAppAtLogin` so it reflects the electron setting a bit more clearly

client/settings/app-settings.jsx
```diff
@@ -55,7 +51,7 @@ class AppForm extends React.Component {
     return (
       <form noValidate={true} onSubmit={onSubmit}>
         <SubmitOnEnter />
-        <VolumeSettings>
+        <FormContainer>
```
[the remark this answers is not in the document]
williamhammond: I had it grouped by divs, using the `FormContainer` but it looked inf of odd. I it row oriented and it looked like this
row-gap

client/settings/app-settings.jsx
```diff
@@ -7,11 +7,13 @@ import SubmitOnEnter from '../forms/submit-on-enter'
 import PlayIcon from '../icons/material/play_arrow-24px.svg'
 import StopIcon from '../icons/material/stop-24px.svg'
 import { TextButton } from '../material/button'
+import CheckBox from '../material/check-box'
 import Slider from '../material/slider'
 
-const VolumeSettings = styled.div`
-  display: flex;
-  align-items: flex-end;
+const AppContainer = styled.div`
```
williamhammond, line 13: maybe AppSettingsContainer?

common/local-settings.ts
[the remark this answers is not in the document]
2Pacalypse-: I wanted to avoid using "login" since it might've been a bit ambiguous from our point of view, but I'll take it ^^.

client/settings/app-settings.jsx
2Pacalypse-: Sure.

app/app.ts
```diff
@@ -151,6 +150,12 @@ function setupIpc(localSettings: LocalSettings, scrSettings: ScrSettings) {
     })
 
   localSettings.on('change', settings => {
+    const loginSettings = {
+      openAtLogin: settings.openAppAtLogin,
+    }
+    if (loginSettings !== app.getLoginItemSettings()) {
```
2Pacalypse-, line 156: In JavaScript, objects are compared by their reference so this condition will always be true. This should probably be rewritten to something like:
```ts
if (settings.openAppAtLogin !== app.getLoginItemSettings().openAtLogin) {
  //
}
```

2Pacalypse- (changes requested):

client/settings/app-settings.jsx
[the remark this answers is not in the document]
2Pacalypse-: I think your initial choice to use the `FormControl` component was correct. We want to display all settings in two columns, at least for now. The only tricky thing then becomes how to display the test sound button, since it should be visually connected to the volume slider. One option is either this:

image

or something like this:

image

For what it's worth, I think I prefer the former.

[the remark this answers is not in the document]
2Pacalypse-: Also, it's quite annoying that the checkbox label doesn't fit on one line :d. Changing it to something like `Run ShieldBattery on system startup` might help with that.

app/app.ts
williamhammond: ah that's embarassing, i thought != was reference and !== was value 😅

williamhammond:
I went with 
image
to give a little more space for volume control and because it didn't require any new styling

app/app.ts
```diff
@@ -151,6 +150,12 @@ function setupIpc(localSettings: LocalSettings, scrSettings: ScrSettings) {
     })
 
   localSettings.on('change', settings => {
+    const loginSettings = {
+      openAtLogin: settings.openAppAtSystemStart,
```
2Pacalypse-, line 154: Nit: I would rather set this directly inside the `if` condition. It's very likely that if we add more settings to set at login, they would have their own, separate `if` conditions and we couldn't use this general object at all.

common/local-settings.ts
```diff
@@ -16,6 +16,7 @@ export interface ShieldBatteryAppSettingsData {
 }
 
 export interface LocalSettingsData extends ShieldBatteryAppSettingsData {
+  openAppAtSystemStart: boolean
```
2Pacalypse-, line 19: Sorry to do this to you, but maybe `runAppAtSystemStart` might be a bit more inline with our new label text :d. Ignore if you don't feel like changing it anymore.

client/settings/app-settings.jsx
```diff
@@ -55,19 +51,28 @@ class AppForm extends React.Component {
...
-            max={100}
-            step={1}
-            showTicks={false}
-            onChange={this.onMasterVolumeChange}
-          />
-          <TestSoundButton label={testSoundLabel} onClick={onTestSoundClick} />
-        </VolumeSettings>
+        <FormContainer>
+          <div>
+            <StyledSlider
```
2Pacalypse-, line 56: You can just wrap this component and the test sound button component in a `VolumeSettingsContainer` component that already existed to get them aligned as they were previously. This would also eliminate the need to fiddle with margins of the button that probably would've needed to be done with the new layout.

But also make sure to leave the outer `div` in place as that one wraps the first column and will be needed when we add more items to it.

2Pacalypse- (changes requested):

client/settings/app-settings.jsx
williamhammond: So like?

```
        <FormContainer>
          <div>
            <VolumeSettings>
              <StyledSlider
                {...bindCustom('masterVolume')}
                label='Master volume'
                tabIndex={0}
                min={0}
                max={100}
                step={1}
                showTicks={false}
                onChange={this.onMasterVolumeChange}
              />
              <TestSoundButton label={testSoundLabel} onClick={onTestSoundClick} />
            </VolumeSettings>
          </div>
          <div>
            <CheckBox
              {...bindCheckable('runAppAtSystemStart ')}
              label='Run ShieldBattery on system startup'
              inputProps={{ tabIndex: 0 }}
            />
          </div>
        </FormContainer>
        ```

2Pacalypse-: Yep, that ought to do it. I think ^^

2Pacalypse- (approved):

2Pacalypse-:
Thanks!

PR #711, opened 2021-06-28
Implement better rules to limit order queues

neivv:
Fixes flying SCVs and a mineral hack


tec27:
Have done some cursory tests that everything seems to work here, going to merge this with the expectation that we'll do more testing on the staging server

Issue #710, opened 2021-06-27
Add a "start minimized" setting


tec27:
Now that we can set the app to run at startup, might be good to let it start in the system tray instead of directly visible

PR #714, opened 2021-06-30
Add a "start minimized" setting (#710)

joel-rich:


common/local-settings.ts
```diff
@@ -17,6 +17,7 @@ export interface ShieldBatteryAppSettingsData {
 
 export interface LocalSettingsData extends ShieldBatteryAppSettingsData {
   runAppAtSystemStart: boolean
+  startAppMinimized: boolean
```
2Pacalypse-, line 20: I would rather name this setting something like `runAppAtSystemStartMinimized` to make it clear it's only applied to the `runAppAtSystemStart` setting, even if it's a bit of a mouthful :d

client/settings/app-settings.jsx
```diff
@@ -79,6 +79,11 @@ class AppForm extends React.Component {
               label='Run ShieldBattery on system startup'
               inputProps={{ tabIndex: 0 }}
             />
+            <CheckBox
```
2Pacalypse-, line 82: This should probably be disabled if the `Run ShieldBattery on system startup` setting is `false`. There's an example in some of our other settings that shows how you can reference the value of one setting in another.

app/settings.ts
```diff
@@ -217,6 +218,7 @@ export class LocalSettings extends Settings<LocalSettingsData> {
     if (!settings.version || settings.version < 7) {
       log.verbose('Found settings version 6, migrating to version 7')
       newSettings.runAppAtSystemStart = true
+      newSettings.startAppMinimized = false
```
2Pacalypse-, line 221: You should increase the `VERSION` of settings whenever they change, since some users might already be on this version and this migration won't run for them.

2Pacalypse- (changes requested):
Hey, thanks for the PR!

However, I think this PR misses the point of the #710 issue just a little bit. The "start minimized" setting should only apply when running the application on system startup. Your implementation applies to scenarios where a user starts the application manually as well, which could definitely be a bit confusing to start it as minimized then.

You should be able to achieve the former by setting the `args` property in `setLoginItemSettings()` API: https://www.electronjs.org/docs/api/app#appsetloginitemsettingssettings-macos-windows
There is an example there as well which could be helpful (you can  ignore the `path` property).

app/app.ts
```diff
@@ -150,11 +150,11 @@ function setupIpc(localSettings: LocalSettings, scrSettings: ScrSettings) {
...
-    if (settings.runAppAtSystemStart !== app.getLoginItemSettings().openAtLogin) {
-      app.setLoginItemSettings({
-        openAtLogin: settings.runAppAtSystemStart,
-      })
-    }
+    app.setLoginItemSettings({
+      openAtLogin: settings.runAppAtSystemStart,
+      args: [settings.runAppAtSystemStartMinimized ? '--hidden' : ''],
+    })
+
```
joel-rich, line 157: @2Pacalypse- I removed the check to see if runAppAtSystemStart changed since I didn't see a simple way to also check if --hidden was already in the args. It seems to be in getLoginSettings().launchItems[].args. I'm not sure how I would determine which of the launchItems[] to check args for. I can look into it more if you'd like the checks back.

2Pacalypse-: Yeah I would still like to have the check on, but maybe a better way to write it would be to save the last value of these two values outside this handler, and then check against them. Something like:
```ts
let lastRunAppAtSystemStart: boolean | undefined
let lastRunAppAtSystemStartMinimized: boolean | undefined

localSettings.on('change', settings => {
  if (
    lastRunAppAtSystemStart !== settings.runAppAtSystemStart ||
    lastRunAppAtSystemStartMinimized !== settings.runAppAtSystemStartMinimized
  ) {
    app.setLoginItemSettings({
      openAtLogin: settings.runAppAtSystemStart,
      args: [settings.runAppAtSystemStartMinimized ? '--hidden' : ''],
    })
    lastRunAppAtSystemStart = settings.runAppAtSystemStart
    lastRunAppAtSystemStartMinimized = settings.runAppAtSystemStartMinimized
  }

  //
```

2Pacalypse- (changes requested):

app/app.ts
joel-rich: I made this change, pretty much verbatim to what you have here. It works as expected and it seems the best way to do it.

client/settings/app-settings.jsx
joel-rich: I made a whole styled Checkbox for this since it seems I can't add styles/classes directly to Checkbox. Could you point me towards the documentation for how the whole styling thing works for this project? Is it all basic electron features (or react?) or is there a theme/library (like bootstrap) that's being used?

2Pacalypse-: Yeah, that's exactly what you should've done. We use [Styled Components](https://styled-components.com/) library for styling, so you can poke around its documentation to learn a bit more about it. You can also browse through our client components to see various ways in which we have used it so far.

2Pacalypse- (approved):
Thanks a lot for the PR! It looks good to me and I'm going to merge it now.

PR #717, opened 2021-07-01
Implement the party nav entry with accompanying actions.

2Pacalypse-:
The following two actions are implemented:
- invite players by entering their name
- leave party action


client/parties/action-creators.ts
```diff
@@ -110,3 +110,23 @@ export function acceptPartyInvite(partyId: string): ThunkAction {
     })
   }
 }
+
+export function leaveParty(): ThunkAction {
+  return dispatch => {
+    dispatch({
+      type: '@parties/leavePartyBegin',
```
tec27, line 117: should probably include info about what party is being left

client/parties/party-invite-dialog.tsx
```diff
@@ -0,0 +1,98 @@
...
+import TextField from '../material/text-field'
+import { useAppDispatch, useAppSelector } from '../redux-hooks'
+import { openSnackbar } from '../snackbars/action-creators'
+import { inviteToParty } from './action-creators'
+
+const userValidator = composeValidators(
+  required('Enter a username'),
+  minLength(USERNAME_MINLENGTH),
+  maxLength(USERNAME_MAXLENGTH),
+  regex(USERNAME_PATTERN, 'User name contains invalid characters'),
```
tec27, line 17: Should probably be `Username` for consistency

server/lib/parties/party-service.ts
```diff
@@ -242,25 +260,25 @@ export default class PartyService {
...
-      sockets => this.handleClientQuit(sockets),
+      sockets => this.removeClientFromParty(sockets),
     )
   }
 
-  private handleClientQuit(clientSockets: ClientSocketsGroup) {
+  private removeClientFromParty(clientSockets: ClientSocketsGroup) {
     const party = this.getClientParty(clientSockets)
     if (!party) {
       logger.error('error while handling client quitting, party not found')
```
tec27, line 278: Missed this in previous reviews, but I'd create an Error when logging this and include it, so that we get a stack trace on this

server/lib/websockets/testing/websockets.ts
```diff
@@ -44,6 +44,9 @@ export class FakeNydusServer
     }
   })
   unsubscribeClient = jest.fn((client: NydusClient, path: string) => {
+    this.fakeSubscriptions.get(path)?.forEach(client => {
```
tec27, line 47: This seems wrong?

The previous implementation of this method removes only a specific client, returning true if the client was subscribed. Your implementation adds to this by telling *every* client that was subscribed to the path that they've been unsubscribed.

I think the more correct implementation here would be to look at the return value below, if true, just do `(client as InspectableNydusClient).unsubscribeClient(client, path))`. There's no real need to go through the roundabout way of retrieving the same value out of the subscriptions

tec27: You also didn't add similar functionality to `unsubscribeAll` below, which you should definitely do if your intention is to be able to assert against this at the client level

server/lib/websockets/testing/websockets.ts
```diff
@@ -76,6 +79,7 @@ export class InspectableNydusClient extends NydusClient {
   // NOTE(tec27): We add an explicit type because it makes calling the method have better
   // documentation (the arguments are named, vs the arg_0, arg_1 stuff from Jest)
   publish: (path: string, data: any) => void = jest.fn()
+  unsubscribeClient: (client: NydusClient, path: string) => boolean = jest.fn()
```
tec27, line 82: I'd just call this `unsubscribe`, `unsubscribeClient` is weird when this is already a Client. I also don't get why this needs to take a `NydusClient`? That should always be `===` itself or this call makes no sense

tec27 (changes requested):

client/parties/action-creators.ts
2Pacalypse-: Initially I had `partyId` here, but later removed it since I realized I don't use it anywhere. And I don't mean that alone makes it useless, but once I realized I don't even send the `partyId` to the server, I stopped looking at this as a "leave _this_ party" action, and more like a "leave the client's party, whatever it is, and if there's one".

Now I'm not sure if that's the best way of looking at this action, but that's the least amount of information I needed to perform it. The main reason why I'm skeptical of approaching it like I did is when looking at the HTTP route. Currently I defined the API route to leave the party as `DELETE /api/1/parties`. While speaking from a strictly REST perspective, `DELETE /api/1/parties/:partyId/:clientId` maybe makes the most amount of sense to me?

tec27: Yes, I think that URL makes more sense (I'd view `:clientId` as pretty optional, but I don't think it hurts to include it)

tec27: I think in general you should try to include information about what entity the client is trying to update, rather than figuring this out all automatically on the server, simply because it gives us a better opportunity of identifying bugs and mismatched client state. It also prevents a class of timing issues, because the server will only update the state about the thing the client was explicit it was talking about

2Pacalypse-: Yeah, I only included `:clientId` because without it, the URL seemed like a party itself was being deleted :d

tec27 (approved):

PR #722, opened 2021-07-03
Implement the party view, compromising of a chat and a user list.

2Pacalypse-:
Still missing:
- logic when leader leaves the party
- kick action

Those will come in subsequent PRs.

### DIFF
--- a/client/parties/party-reducer.ts
+++ b/client/parties/party-reducer.ts
@@ -1,6 +1,7 @@
 import cuid from 'cuid'
 import { List, OrderedMap, Record } from 'immutable'
 import { PartyUser } from '../../common/parties'
+import { NETWORK_SITE_CONNECTED } from '../actions'
 import { TextMessageRecord } from '../messaging/message-records'
 import { keyedReducer } from '../reducers/keyed-reducer'
 import {
@@ -178,5 +179,9 @@ export default keyedReducer(new PartyRecord(), {
     return state
       .set('messages', state.messages.slice(-INACTIVE_PARTY_MAX_HISTORY))
       .set('activated', false)
+  },
+
+  [NETWORK_SITE_CONNECTED as any]() {
+    return new PartyRecord()
   },
 })
--- a/client/parties/party-reducer.ts
+++ b/client/parties/party-reducer.ts
@@ -1,6 +1,17 @@
-import { List, Map, Record } from 'immutable'
+import cuid from 'cuid'
+import { List, OrderedMap, Record } from 'immutable'
 import { PartyUser } from '../../common/parties'
+import { TextMessageRecord } from '../messaging/message-records'
 import { keyedReducer } from '../reducers/keyed-reducer'
+import {
+  InviteToPartyMessageRecord,
+  JoinPartyMessageRecord,
+  LeavePartyMessageRecord,
+  SelfJoinPartyMessageRecord,
+} from './party-message-records'
+
+// How many messages should be kept when a party is inactive
+const INACTIVE_PARTY_MAX_HISTORY = 500
 
 export class PartyUserRecord
   extends Record({
@@ -11,10 +22,12 @@ export class PartyUserRecord
 
 export class PartyRecord extends Record({
   id: '',
-  invites: Map<number, PartyUser>(),
-  members: Map<number, PartyUser>(),
+  invites: OrderedMap<number, PartyUser>(),
+  members: OrderedMap<number, PartyUser>(),
   leader: new PartyUserRecord(),
   messages: List(),
+  hasUnread: false,
+  activated: false,
 }) {}
 
 export default keyedReducer(new PartyRecord(), {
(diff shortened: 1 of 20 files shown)


client/parties/party-reducer.ts
```diff
@@ -25,9 +39,16 @@ export default keyedReducer(new PartyRecord(), {
...
       id,
-      invites: Map(invites.map(i => [i.id, new PartyUserRecord(i)])),
-      members: Map(members.map(m => [m.id, new PartyUserRecord(m)])),
+      invites: OrderedMap(invites.map(i => [i.id, new PartyUserRecord(i)])),
+      members: OrderedMap(members.map(m => [m.id, new PartyUserRecord(m)])),
       leader: new PartyUserRecord(leader),
+      messages: List([
+        new SelfJoinPartyMessageRecord({
+          id: cuid(),
+          time: Date.now(),
```
tec27, line 48: I know we decided to leave this in place for Chat because it was already like this, but now you're copy-pasting this to other stuff, please don't :(

server/lib/parties/party-service.test.ts
```diff
@@ -451,4 +476,39 @@ describe('parties/party-service', () => {
...
+    })
+
+    test('should throw if the user is not in party', () => {
+      expect(() =>
+        partyService.sendChatMessage(party.id, user3.id, 'Hello World!'),
+      ).toThrowErrorMatchingInlineSnapshot(`"User not in party"`)
+    })
+
+    test('should publish "chatMessage" event to the party path', () => {
+      Date.now = jest.fn(() => 272727)
```
tec27, line 503: Does this get reset between tests? My guess is probably not, so you should probably do this in a `before` and clean it up in an `after` (or, better yet, don't do this at all and fix the code to allow this to be passed in or at the very least inject the thing that retrieves the current date)

server/lib/parties/party-service.ts
```diff
@@ -234,8 +239,38 @@ export default class PartyService {
...
+    if (!user) {
+      throw new PartyServiceError(PartyServiceErrorCode.UserNotInParty, 'User not in party')
+    }
+
+    const text = filterChatMessage(message)
+
+    this.publisher.publish(getPartyPath(partyId), {
+      type: 'chatMessage',
+      from: user,
+      time: Date.now(),
```
tec27, line 271: As mentioned in the test above, don't retrieve the Date this way. This is an undeclared dependency and causes headaches for testing

server/lib/parties/party-service.ts
```diff
@@ -234,8 +239,38 @@ export default class PartyService {
   }
 
   leaveParty(partyId: string, userId: number, clientId: string) {
+    const party = this.parties.get(partyId)
+    if (!party) {
```
tec27, line 243: imo these checks present a security issue. I know it feels *nice* to just retrieve the party and check the members, but by doing things this way you're letting users verify the existence of parties they aren't in.

I'd prefer that you retrieve the user's current party from their client ID and check that it matches the party ID they passed, and the order of the errors would then be:
- You're not in a party
- You're not in *that* party

At no point would we tell them "that party exists but you're not in it"

server/lib/parties/party-service.ts
```diff
@@ -234,8 +239,38 @@ export default class PartyService {
...
+    }
+
     const clientSockets = this.getClientSockets(userId, clientId)
-    this.removeClientFromParty(clientSockets)
+    this.removeClientFromParty(clientSockets, party)
+  }
+
+  sendChatMessage(partyId: string, userId: number, message: string) {
+    const party = this.parties.get(partyId)
+    if (!party) {
```
tec27, line 257: Same as above, this is revealing far too much information to the user

tec27: And honestly you should audit all the APIs in here for this, I think they have similar problems.

We're not dealing with trusted clients here, we can't be returning errors like we are. Keep the information limited to what we need for reasonable client behavior.

tec27 (changes requested):

server/lib/parties/party-service.ts
```diff
@@ -287,12 +285,13 @@ export default class PartyService {
...
   }
 
-  private subscribeToParty(clientSockets: ClientSocketsGroup, party: PartyRecord) {
+  private subscribeToParty(clientSockets: ClientSocketsGroup, party: PartyRecord, time: number) {
     clientSockets.subscribe<PartyInitEvent>(
       getPartyPath(party.id),
       () => ({
         type: 'init',
         party: toPartyJson(party),
+        time,
```
tec27, line 294: Not sure that this is really safe, since this function will be called every time a new client connects, and this time feels like it's meant to be "when this action occurs" not "when this client was initially subscribed"? You probably need to pass a closure to retrieve the current time instead (which likely makes the injection approach more appealing here)

Note that I don't think this breaks any code *right now* but it feels semantically weird and likely to cause weird breakages in the future

2Pacalypse-: Hmm, not sure I understand how to do this. Does this mean the `init` event should not reflect the time a user accepted the party invite?

Also, how would I test this?

tec27: Is the init event about a user accepting an invite? Because it sure seems to me (from the name and all) that it's about initializing a client to their current state. To me, the `time` field in an event should almost always be "the time this event was sent by the server, in server time", because we use it to sort events.

If you don't want that to be the case here, I mean, name it something else I guess, but it seems like a problem waiting to happen and not particularly useful on the client (as implemented, you could accomplish the same thing by assuming that anything that comes in an `init` event was prior to anything else you have received, because it will always sort that way with this time)

As for testing a closure-provided time, you would pass a closure that returns a constant time, it's not really different from how you were already testing. But like I said, injection is probably a cleaner approach, in which case you would inject a `TimeProvider` or a `Clock` or whatever that would return a constant value (again, fairly similar to what you already have)

server/lib/time/get-current-time.ts
```diff
@@ -0,0 +1,7 @@
+/**
+ * Returns the current time, in milliseconds. This function mainly exists so it can be injected into
+ * classes that needs to get the current time and also be easily testable.
+ */
+export function getCurrentTime(): number {
```
tec27, line 5: I think in general, unless you *have* to inject a function for some reason, creating a class is just a lot cleaner and doesn't require manual registration or `@inject(name)` things

PR #723, opened 2021-07-03
Handle user changing a party.

2Pacalypse-:
This can happen when a user gets invited to a party while they were
already in a different party. By accepting the invite to a new party,
they will immediately leave the old party and be put in a new party.

TODO: The transition from one party to another might be a bit jarring,
so maybe eventually we want to display some kind of a confirmation
dialog before doing it?

### DIFF
--- a/client/parties/party-view.tsx
+++ b/client/parties/party-view.tsx
@@ -12,10 +12,9 @@ import { SlotActions } from '../lobbies/slot-actions'
 import { TextButton } from '../material/button'
 import Chat from '../messaging/chat'
 import { Message } from '../messaging/message-records'
-import { push } from '../navigation/routing'
+import { replace } from '../navigation/routing'
 import { urlPath } from '../network/urls'
 import { useAppDispatch, useAppSelector } from '../redux-hooks'
-import { usePrevious } from '../state-hooks'
 import { background700, background800 } from '../styles/colors'
 import { activateParty, deactivateParty, leaveParty, sendChatMessage } from './action-creators'
 import {
@@ -130,25 +129,29 @@ function renderPartyMessage(msg: Message) {
   }
 }
 
-export function PartyView() {
+interface PartyViewProps {
+  params: { partyId: string }
+}
+
+export function PartyView(props: PartyViewProps) {
   const dispatch = useAppDispatch()
   const party = useAppSelector(s => s.party)
   const partyId = party.id
-  const prevPartyId = usePrevious(partyId)
+  const routePartyId = decodeURIComponent(props.params.partyId)
   // Party can change when a user accepts an invite to a new party while already being a member of a
   // different party that is currently being rendered.
-  const partyChanged = !!prevPartyId && prevPartyId !== partyId
+  const partyChanged = routePartyId !== partyId
 
   useEffect(() => {
     const isInParty = !!partyId
 
     if (isInParty) {
       if (partyChanged) {
-        push(urlPath`/parties/${partyId}`)
+        replace(urlPath`/parties/${partyId}`)
       }
       dispatch(activateParty(partyId))
     } else {
-      push('/')
+      replace('/')
     }
 
     return () => dispatch(deactivateParty(partyId))
--- a/client/parties/party-view.tsx
+++ b/client/parties/party-view.tsx
@@ -149,8 +149,9 @@ export function PartyView(props: PartyViewProps) {
       // party view while you're in a different party.
       if (routePartyId !== partyId) {
         replace(urlPath`/parties/${partyId}`)
+      } else {
+        dispatch(activateParty(partyId))
       }
-      dispatch(activateParty(partyId))
     } else {
       replace('/')
     }
--- a/client/parties/party-view.tsx
+++ b/client/parties/party-view.tsx
@@ -138,15 +138,16 @@ export function PartyView(props: PartyViewProps) {
   const party = useAppSelector(s => s.party)
   const partyId = party.id
   const routePartyId = decodeURIComponent(props.params.partyId)
-  // Party can change when a user accepts an invite to a new party while already being a member of a
-  // different party that is currently being rendered.
-  const partyChanged = routePartyId !== partyId
 
   useEffect(() => {
     const isInParty = !!partyId
 
     if (isInParty) {
-      if (partyChanged) {
+      // The mismatch between the current URL and the party state can occur in a couple of ways. One
+      // is when a user accepts an invite to a new party while already being a member of a different
+      // party that is currently being rendered. Another way is that someone links to their own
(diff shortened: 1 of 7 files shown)


2Pacalypse-:
Depends on https://github.com/ShieldBattery/ShieldBattery/pull/722

client/parties/party-view.tsx
```diff
@@ -129,35 +130,29 @@ function renderPartyMessage(msg: Message) {
   }
 }
 
-interface PartyViewProps {
-  params: { partyId: string }
-}
-
-export function PartyView(props: PartyViewProps) {
+export function PartyView() {
```
tec27, line 133: I don't think removing the params here is a very good idea tbh. If we're putting the party ID in the URL, you should keep that up to date. e.g. if they come to this view and the ID in the URL doesn't match their current party, you should `replace` to the correct URL.

If we're not putting the party ID in the URL, fine, but that doesn't seem to be the case here.

client/parties/party-view.tsx
```diff
@@ -129,35 +130,29 @@ function renderPartyMessage(msg: Message) {
...
-    if (isLeavingParty) {
-      push('/')
-    }
-  }, [isLeavingParty])
+    const isInParty = !!partyId
 
-  useEffect(() => {
     if (isInParty) {
+      if (partyChanged) {
+        push(urlPath`/parties/${partyId}`)
```
tec27, line 147: `push` doesn't make sense here, the previous URL is effectively a location that no longer exists for this user, so why are we keeping it in the back stack?

client/parties/party-view.tsx
```diff
@@ -129,35 +130,29 @@ function renderPartyMessage(msg: Message) {
...
+      }
       dispatch(activateParty(partyId))
-    } else if (!isLeavingParty) {
+    } else {
       push('/')
```
tec27, line 151: push also doesn't make sense here

tec27 (changes requested):

client/parties/party-view.tsx
```diff
@@ -130,25 +129,29 @@ function renderPartyMessage(msg: Message) {
...
+export function PartyView(props: PartyViewProps) {
   const dispatch = useAppDispatch()
   const party = useAppSelector(s => s.party)
   const partyId = party.id
-  const prevPartyId = usePrevious(partyId)
+  const routePartyId = decodeURIComponent(props.params.partyId)
   // Party can change when a user accepts an invite to a new party while already being a member of a
   // different party that is currently being rendered.
-  const partyChanged = !!prevPartyId && prevPartyId !== partyId
+  const partyChanged = routePartyId !== partyId
```
tec27, line 143: This isn't a good naming or a description of what is occurring here, imo

Yes, changing parties is *one way* that a mismatch between the current URL and the party state can occur, but it is not the *only* way. Another way is that someone links to their own party view while you're in a different party.

So I would either A) not name this, just inline this condition in the useEffect below, since the condition is self-explanatory or B) name it something better, like `urlMatchesState`. I probably prefer A

server/lib/parties/party-service.ts
```diff
@@ -332,7 +332,13 @@ export default class PartyService {
       })
     }
 
-    this.clientSocketsToPartyId.delete(clientSockets)
+    const clientPartyId = this.clientSocketsToPartyId.get(clientSockets)!
```
tec27, line 335: If this wasn't caught by tests, then test coverage is not sufficient. Thus, this fix should include an additional test or a modification to existing tests

2Pacalypse-: I wasn't testing this because `clientSocketsToPartyId` is a private property in the party service. But I guess it doesn't make sense to have private stuff in services when writing unit tests, since we most likely want to test the internals of the service?

tec27: Generally, no. The service should be testable by controlling inputs/examining outputs. Stuff that is "private to the service" is an implementation detail, and shouldn't really be depended upon for tests.

(This isn't always possible, but in new code that has tests written for it directly, it is nearly always possible to design things in that way)

Think more deeply about what the issue you're seeing here is and how you would detect that in a test. What test could you write that would fail with the old code but succeed with the new code? (Then ideally, delete this fix, run the test, ensure it fails, and then re-add the fix and ensure it passes)

tec27: I'll just give you an example so that we don't go back and forth on this forever and actually get this done:

The issue you're seeing here is that in the case of switching parties, a user ends up only halfway into a party, because after they switched parties the system comes along and deletes their registration? How would this manifest in the application, what error states would occur? What did you see when this happened that even made you realize it was an issue?

I'd guess the user would be unable to leave the new party because the system would fail to look up their client ID. Thus a reasonable test might be:

1) Join a party
2) Switch to a different party
3) Try to leave the different party
4) Verify that leaving the party worked

Think less about this stuff being "I want to call function X, Y, Z" and more about actions that users are taking, and how those actions relate to the APIs present. The more realistic the sequence to what users are doing, and the more closely we verify state that is actually being used by those users, the better our tests will be at finding "real" issues and the more resilient they'll be to changes in the underlying implementation.

2Pacalypse-: Actually, the user was able to leave the new party just fine; the problem occurred when trying to get the party by client's sockets (after it was erroneously deleted), which was mostly used in cleaning up the user when they disconnected (I only even noticed this bug while getting the error logged in the console). But anyway, added the test for that too.

For what is worth, should probably add tests at various other places to make sure stuff gets cleaned up properly when a user arbitrarily disconnects. But gonna do that outside this PR. The reason why I didn't do it in the first place is because I wasn't sure how to simulate the user disconnect with our fake nydus client we use in testing, but after spending two hours on it now, I finally figured it out! And it's one line of code ><

tec27: Some lines of code are more expensive than others :P

client/parties/party-view.tsx
```diff
@@ -138,15 +138,16 @@ export function PartyView(props: PartyViewProps) {
...
     if (isInParty) {
-      if (partyChanged) {
+      // The mismatch between the current URL and the party state can occur in a couple of ways. One
+      // is when a user accepts an invite to a new party while already being a member of a different
+      // party that is currently being rendered. Another way is that someone links to their own
+      // party view while you're in a different party.
+      if (routePartyId !== partyId) {
         replace(urlPath`/parties/${partyId}`)
       }
       dispatch(activateParty(partyId))
```
tec27, line 153: I think putting this in an `else` is fine/preferable here as well (this effect will re-run after the URL changes anyway, so this prevents a double activation)

tec27 (approved):

PR #724, opened 2021-07-05
Improve the party handling on non-electron clients.

2Pacalypse-:
This PR does two things:
- doesn't display party invite notifications on web clients
- throws an error if user tries to accept a party invitation on a
  non-electron client

### DIFF
--- a/client/notifications/socket-handlers.ts
+++ b/client/notifications/socket-handlers.ts
@@ -1,6 +1,8 @@
 import type { NydusClient, RouteHandler, RouteInfo } from 'nydus-client'
-import { NotificationEvent } from '../../common/notifications'
+import { NotificationEvent, NotificationType } from '../../common/notifications'
 import { dispatch, Dispatchable } from '../dispatch-registry'
+
+const ELECTRON_ONLY_NOTIFICATION_TYPES = [NotificationType.PartyInvite]
 
 type EventToActionMap = {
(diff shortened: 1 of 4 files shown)


2Pacalypse-:
Depends on https://github.com/ShieldBattery/ShieldBattery/pull/723

client/notifications/socket-handlers.ts
```diff
@@ -1,7 +1,9 @@
 import type { NydusClient, RouteHandler, RouteInfo } from 'nydus-client'
-import { NotificationEvent } from '../../common/notifications'
+import { NotificationEvent, NotificationType } from '../../common/notifications'
 import { dispatch, Dispatchable } from '../dispatch-registry'
 
+const ELECTRON_ONLY_NOTIFICATION_TYPES = [NotificationType.PartyInvite]
```
tec27, line 5: You should provide an explicit type for this array, TS generally isn't gonna be able to pick a good one (and you should also make it `Readonly` so that we don't accidentally mutate it

tec27: Honestly it's probably better to just make this a `Set` too. It doesn't matter that much with the amount of types we're talking about here, but may as well pay the one-time init cost to get O(n) inclusion checks

tec27 (approved):

PR #725, opened 2021-07-05
Assign new leader in a party when the old one leaves.

2Pacalypse-:
Generally the person who has joined the party the earliest will be
chosen as a new leader, but there's no robust solution implemented that
ensures this. My thinking was that since the multiple invites can be
sent at once, and the order in which those invites are accepted are
pretty arbitrary already, there's really no reason why someone
"deserves" to be a new leader more than someone else. Or well, no reason
big enough to spend time implementing a robust solution for this at
least :)

### DIFF
--- a/client/parties/actions.ts
+++ b/client/parties/actions.ts
@@ -30,6 +30,7 @@ export type PartyActions =
   | UpdateJoin
   | UpdateLeave
   | UpdateLeaveSelf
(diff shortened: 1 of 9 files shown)


2Pacalypse-:
Depends on https://github.com/ShieldBattery/ShieldBattery/pull/724

tec27 (approved):

PR #726, opened 2021-07-05
Implement an action to kick a player from a party.

2Pacalypse-:


2Pacalypse-:
Depends on https://github.com/ShieldBattery/ShieldBattery/pull/725

client/parties/action-creators.ts
```diff
@@ -161,6 +161,31 @@ export function sendChatMessage(partyId: string, message: string): ThunkAction {
...
+  return dispatch => {
+    const params = { partyId, targetId }
+    dispatch({
+      type: '@parties/kickFromPartyBegin',
+      payload: params,
+    })
+
+    dispatch({
+      type: '@parties/kickFromParty',
+      payload: fetch<void>(apiUrl`parties/${partyId}/${targetId}/kick`, {
```
tec27, line 174: This URL feels fairly weird from a REST perspective, I think the options for fixing that are:

1) remove `/kick`( so this is `DELETE parties/:id/:target`). I assume this has issues with the distinguishing between leave/kick on the server and that's perhaps why you didn't do it? That seems likely solveable to me, either by using state about the client or by adding a query parameter (`?type=kick` or something)

2) Make this a POST, which is how we normally handle "command" URLs like this

client/parties/socket-handlers.ts
```diff
@@ -126,6 +127,31 @@ const eventToAction: EventToActionMap = {
       },
     }
   },
+
+  kick: (partyId, event) => (dispatch, getState) => {
+    const { target, time } = event
+    const selfUser = getState().auth.user
+    if (selfUser.id === target.id) {
+      // It was us who have been kicked from the party
```
tec27, line 135: nit: 'has been'

server/lib/parties/party-service.ts
```diff
@@ -280,6 +282,48 @@ export default class PartyService {
...
+        PartyServiceErrorCode.InvalidAction,
+        "Can't kick player who is not in your party",
+      )
+    }
+
+    if (kickingUser.id === target.id) {
+      throw new PartyServiceError(PartyServiceErrorCode.InvalidAction, "Can't kick yourself")
+    }
+
+    party.members.delete(target.id)
```
tec27, line 312: Feels kinda bad that this can't reuse `removeClientFromParty`, I wonder if instead you could separate the two pieces here:

1) A `'kick'` action that notifies the party that someone is being kicked
2) A `'leave'` action that notifies the party that this person is now gone

And then you could just publish this first action, followed by a call to `removeClientFromParty` to do all the cleanup

Probably needs a bit more thinking on the client as far as showing messages and stuff, but seems doable (on the client the `kick` could update the members + add a message, and then `leave` could only add a message if the person was in `members`? That or just start managing those messages entirely on the server)

tec27 (changes requested):

client/parties/action-creators.ts
2Pacalypse-: Yeah, initially I went with 1., but I didn't know how to distinguish this API route from leave. To be honest, I still don't know how to do that, besides adding a query param, which I will do here.

The reason why I didn't go with 2. is that pretty much _all_ of these routes seem like "command" URLs to me, so should probably make them all POST as well.

server/lib/parties/party-service.ts
2Pacalypse-: Yeah, I was trying to avoid the situation we had in the lobby code where we had to pass "removal type" through several methods to do the right thing. And yeah, could split this into two pieces with one piece having a shared logic that removes the player from the party. The reason why I didn't do that here is because the two features of removing a player from the party are "changing the host if it leaves" and "removing the party if last person leaves"; both of which are not applicable here since you can't kick yourself from a party. But I guess it doesn't hurt having that logic even when cleaning the kicked player, especially since we might add additional logic that needs to be shared between these two scenarios.

client/parties/action-creators.ts
tec27: I mean, all REST actions are effectively "commands" at the end of the day, I don't think I'd make *everything* a POST. The question really is, I guess: does this action map to one of the REST verbs well or not? If we're creating something or destroying something, it probably does map quite well (but it may be on some sub-entity, like how we use `/messages`). Sometimes the same action might need more description (so a kick and a leave may both be `DELETE`s but need a way to distinguish so we can return the right errors to clients or whatever)

server/lib/parties/party-service.ts
tec27: It definitely does not hurt to run logic that won't change anything, yes. Running (cheap) things unnecessarily is far, far, far preferable to having logic copy-pasted a bunch of places. We (I) have already caused bugs in the past from not properly tracking down all the places lobbies do certain things that would ideally be shared, and the easy way to prevent that is to share it.

Optimize for stuff that will keep working when we need to add or change the implementation

2Pacalypse-:
Depends on https://github.com/ShieldBattery/ShieldBattery/pull/726

server/lib/parties/party-api.ts
```diff
@@ -112,19 +112,30 @@ export class PartyApi extends HttpApi {
       )
       .delete(
         '/:partyId/:clientId',
-        throttleMiddleware(partyThrottle, ctx => String(ctx.session!.userId)),
-        leave,
+        async (ctx, next) => {
+          let throttle
+          if (ctx.query.type === 'leave') {
```
tec27, line 117: Maybe just put this logic in the leaveOrKick method, after validations run? (I also question whether we really need a separate kick and party action throttle at all, seems kinda unnecessarily granular for how inexpensive the requests are)

side note: you can also use the `apiEndpoint` helper I wrote to make including the validations easier

server/lib/parties/party-api.ts
```diff
@@ -220,18 +231,35 @@ async function accept(ctx: RouterContext) {
...
+    const foundTarget = await findUserById(clientId as number)
+    if (!foundTarget) {
+      throw new httpErrors.NotFound('Target user not found')
+    }
+
+    const kickingUser: PartyUser = { id: ctx.session!.userId, name: ctx.session!.userName }
+    const target: PartyUser = { id: foundTarget.id, name: foundTarget.name }
+
+    await partyService.kickPlayer(partyId, kickingUser, target)
+  }
```
tec27, line 262: probably wanna add:

```ts
} else {
  assertUnreachable(type)
}
```

tec27 (approved):

PR #730, opened 2021-07-13
Improve the handling of the "decline invite" logic in parties.

2Pacalypse-:
This PR does the following:
- don't emit the "decline" event to the party when someone declines an
  invite
- don't display an error in snackbar if decline fails (these errors are
  mostly not of interest to the decliner)
- don't throw an error when inviting someone who is already invited
  (this would reveal to the party the the user has declined the first
  invite)
- only send invite notification if one already doesn't exist that is
  visible
- don't remove user from the invited users list if they decline

### DIFF
--- a/server/lib/parties/party-service.ts
+++ b/server/lib/parties/party-service.ts
@@ -70,6 +70,14 @@ export default class PartyService {
     private clock: Clock,
   ) {}
 
+  /**
+   * Invites a user to the party. If a person who is inviting someone else is not already in a
+   * party, a new party is created and they're made its leader. Note that this should be treated
+   * more as an "allow user to join a party" action, which means that we're not checking if someone
+   * is already invited. So inviting someone multiple times should work without a problem. The
+   * invite notification is only sent if there are no currently visible notifications for the
+   * invited user already though.
+   */
   async invite(
     leader: PartyUser,
     leaderClientId: string,
@@ -93,19 +101,16 @@ export default class PartyService {
         )
       }
 
-      if (party.invites.has(invitedUser.id)) {
-        throw new PartyServiceError(
-          PartyServiceErrorCode.InvalidAction,
-          'An invite already exists for this user',
-        )
-      }
-
       if (party.members.has(invitedUser.id)) {
         throw new PartyServiceError(
           PartyServiceErrorCode.InvalidAction,
           'This user is already a member of this party',
         )
       }
+
+      // An invite might already exist for a user, but we don't treat that as an error as that would
+      // reveal to the inviter that the person has declined their first invite, which should be
+      // treated as private information.
 
       party.invites.set(invitedUser.id, invitedUser)
       this.publisher.publish(getPartyPath(party.id), {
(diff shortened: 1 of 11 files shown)


2Pacalypse-:
Depends on https://github.com/ShieldBattery/ShieldBattery/pull/726

tec27:
Think there may be some issues with this approach that aren't fully thought through (not directly caused by these changes so they don't need to be fixed in this PR, but we should fix them). Namely:

1) User A invites User B
2) User B clears notifications
3) User B realizes they received an invite from A and *do* want to join their party, User A reinvites

In this case, User B doesn't receive the invite because there was already one outstanding (I think?), the only way out of this situation is to recreate the party (not great).

Feels like we should treat the lists we have here a bit differently. The list in the party service should probably be seen as something like, "users in this list are *allowed* to join the party", rather than "users in this list have an active invite". And the logic for notifying users should probably depend on the contents of the notifications table rather than something inherent to the party service (if a user already has a non-dismissed notification for the party, don't send them another one).

tec27 (approved):

server/lib/parties/party-service.ts
```diff
@@ -191,12 +186,10 @@ export default class PartyService {
       )
     }
 
-    if (!party.invites.has(target.id)) {
-      throw new PartyServiceError(
-        PartyServiceErrorCode.InvalidAction,
-        "Can't remove invite for a user that wasn't invited",
-      )
-    }
+    // If an invite doesn't exist for a user it most likely means that the user has already declined
```
tec27, line 189: This in particular feels like something that would be fine as an error case with the semantics change I mentioned previously. You can decide whether you'd like to do that semantics change in this PR or separately.

2Pacalypse-:
> Think there may be some issues with this approach that aren't fully thought through (not directly caused by these changes so they don't need to be fixed in this PR, but we should fix them). Namely:
> 
> 1. User A invites User B
> 2. User B clears notifications
> 3. User B realizes they received an invite from A and _do_ want to join their party, User A reinvites
> 
> In this case, User B doesn't receive the invite because there was already one outstanding (I think?), the only way out of this situation is to recreate the party (not great).

I actually thought about this use case and this was not possible to do through UI; well, sort of :d. The reason why I thought this was not possible is because in order for someone to "reinvite" someone, they would first have to uninvite them (which would delete their invite) and then invite again (which would send them a notification again, regardless of if they cleared it or not; although it would be cleared automatically when they get uninvited so they wouldn't get two notifications). But now I noticed that you could technically still "reinvite" someone through the "invite players" dialog, in which case their invite wouldn't be removed first and the case that you outlined would happen, i.e. player wouldn't receive the notification again.

So your suggestion is still applicable which is what I have now done. Honestly, thinking about the "invites" as an "allowed to join" list and not thinking about what's possible through UI changed everything in my understanding of the party invite system and is much more clear to me now in how it should work. This lead me to also not removing the player from the invites list when they decline an invite, since they're technically still "allowed to join", even if there's no UI at the moment that lets them do that (although there might be one in the future?).

server/lib/parties/party-service.ts
2Pacalypse-: The reason why I removed this error case is because the user was removed from the invites list when they decline, so it could potentially leak information about whether they declined or not. Which made me now think that we should probably not remove the user from the invites list if they decline and this error case can then stay :). See the comment below (above?) where I mentioned this as well.

tec27: Right, I think we're trying to say the same thing here (this error is fine to send provided the invite list becomes an "allowed to join" list), sorry if that wasn't clear :)

server/lib/parties/party-service.ts
```diff
@@ -321,18 +324,52 @@ export default class PartyService {
...
+      await this.notificationService.addNotification({
+        userId: user.id,
+        data: {
+          type: NotificationType.PartyInvite,
+          from: party.leader.name,
+          partyId: party.id,
+        },
       })
+    } catch (err) {
+      throw new PartyServiceError(
```
tec27, line 350: If we're gonna drop whatever this error was, probably it should be manually logged. Definitely be careful about handling *all* errors like this, you can easily swallow e.g. syntax errors, problems with database queries, etc. and make it exceptionally hard for us to track down issues. If you know that a particular kind of error can occur and it's fine (e.g. trying to insert something that may violate a unique constraint, but that's a fine problem), you should check explicitly that it is that error.

(honestly this might be a good feature to have in our error payload middleware: an optional Error parameter to log when the response is sent as the "cause" of the handled error, such that it gets linked up with the proper request ID)

tec27 (approved):

server/lib/parties/party-service.ts
2Pacalypse-: Yeah, I didn't have a specific error in mind here, just wanted to have a way to let the inviter know in case a party invitation was not sent to the user they invited and to try inviting them again (I added client-side handling of this error in #733).

Good point about swallowing syntax errors and other errors that might indicate something being wrong, so I'm gonna log the error as well.

PR #731, opened 2021-07-13
Navigate to the party screen immediately when accepting an invite.

2Pacalypse-:


2Pacalypse-:
Depends on https://github.com/ShieldBattery/ShieldBattery/pull/730

server/lib/parties/party-service.ts
```diff
@@ -336,6 +336,8 @@ export default class PartyService {
         type: 'init',
         party: toPartyJson(party),
         time: this.clock.now(),
+        // TODO(2Pac): This will have to be done differently once we start sending more information
```
tec27, line 339: You can just fix this now tbh, see the user profile stuff in the API for an example. Basically:

1) Collect unique user IDs
2) Call `findUsersById` with them
3) Pass result here

tec27: Could also probably avoid storing anything besides the user ID in this service, but that's a bigger change I'm sure :)

tec27 (approved):

server/lib/parties/party-service.ts
2Pacalypse-: I left this out for a separate PR, which I hope I will do... eventually. The problem that I encountered is that using user model methods directly in the party service breaks the tests. So those methods would need to be mocked in some way, and the only way I found to mock them by using import overrides with Jest, which we already agreed that is probably not the best thing to use. So a better solution would probably be writing a user service which abstracts those model methods and can be injected into tests easily.

PR #732, opened 2021-07-13
Implement an action to manually change the party leader.

2Pacalypse-:

PR #733, opened 2021-07-13
Display the actual reason why accepting a party invite failed.

2Pacalypse-:
These are more or less the only errors that are likely to happen to
users in their normal usage of the party system. Now we display the
actual reason to them if they try to accept an invite to a party that
doesn't exist anymore, or if the party is currently full.

### DIFF
--- a/server/lib/parties/party-api.ts
+++ b/server/lib/parties/party-api.ts
@@ -9,9 +9,11 @@ import {
   AcceptPartyInviteServerBody,
   ChangeLeaderServerBody,
   InviteToPartyServerBody,
+  PartyServiceErrorCode,
   PartyUser,
   SendChatMessageServerBody,
 } from '../../../common/parties'
+import { HttpErrorWithPayload } from '../errors/error-with-payload'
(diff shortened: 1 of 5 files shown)


2Pacalypse-:
Depends on https://github.com/ShieldBattery/ShieldBattery/pull/732

server/lib/parties/party-api.ts
```diff
@@ -58,15 +60,15 @@ function convertPartyServiceError(err: Error) {
     case PartyServiceErrorCode.NotFoundOrNotInvited:
     case PartyServiceErrorCode.NotFoundOrNotInParty:
     case PartyServiceErrorCode.InvalidAction:
-      throw new httpErrors.BadRequest(err.message)
+      throw new HttpErrorWithPayload(400, err.message, { code: err.code })
```
tec27, line 63: Small thing, but feel like we could probably write a helper to make this less repetitive, pass it an HTTP error code + a `Error & { code: string }` and it makes this call?

`asHttpError(400, err)` feels a bit easier to parse to me, anyway

tec27 (approved):

PR #734, opened 2021-07-24
Refactor the matchmaking preferences to support multiple matchmaking types.

2Pacalypse-:


server/lib/session/init.ts
```diff
@@ -1,8 +1,14 @@
...
+import { MatchmakingType } from '../../../common/matchmaking'
 import { Permissions } from '../../../common/users/permissions'
 import { SelfUser } from '../../../common/users/user-info'
 
-export default function initSession(ctx: Context, user: SelfUser, permissions: Permissions) {
+export default function initSession(
+  ctx: Context,
+  user: SelfUser,
+  permissions: Permissions,
+  lastQueuedMatchmakingType: MatchmakingType = MatchmakingType.Match1v1,
```
tec27, line 10: At this point I would pass an object for this instead (leave `ctx` out of it, can probably share typing with the session stuff in some way)

server/migrations/20210719151134-make-matchmaking-preferences-based-on-type.js
```diff
@@ -0,0 +1,77 @@
...
+  `)
+
+  await Promise.all(
+    preferences.map(p => {
+      const data = {
+        useAlternateRace: p.use_alternate_race,
+        alternateRace: p.alternate_race,
+      }
+
+      return db.runSql(sql`
```
tec27, line 22: This can all be written as a single update query, without pulling all the data into JS-land first and creating 1000s of promises. Would vastly prefer that you do that.

Generally don't write migrations that need to operate on a bunch of data on the JS side unless you *have* to (and since this is just a simple modification on existing data in the table, that is definitely *not* the case here)

server/migrations/20210719151134-make-matchmaking-preferences-based-on-type.js
```diff
@@ -0,0 +1,77 @@
...
+      `)
+    }),
+  )
+
+  await db.runSql(`
+    ALTER TABLE matchmaking_preferences
+    RENAME COLUMN preferred_maps
+    TO map_selections;
+  `)
+  await db.runSql(`ALTER TABLE matchmaking_preferences ALTER COLUMN map_selections SET NOT NULL;`)
```
tec27, line 35: This was nullable before and this migration does not update any data in the table to ensure there are no nulls, so, you should probably be doing that first.

server/migrations/20210719151134-make-matchmaking-preferences-based-on-type.js
```diff
@@ -0,0 +1,77 @@
...
+    TO preferred_maps;
+  `)
+  await db.runSql(`ALTER TABLE matchmaking_preferences ALTER COLUMN preferred_maps DROP NOT NULL;`)
+
+  const { rows: preferences } = await db.runSql(`
+    SELECT user_id, matchmaking_type, data
+    FROM matchmaking_preferences;
+  `)
+
+  await Promise.all(
```
tec27, line 56: Same as above, very unnecessary to pull this all into JS like this

server/migrations/20210803222419-add-2v2-matchmaking-type.js
```diff
@@ -0,0 +1,85 @@
+exports.up = async function (db) {
```
tec27, line 1: Do we need to support anything prior to Postgres 9.1? I don't think we really do, that's pretty old at this point (and way older than what we use on the server). Thus you can do this a lot more simply (and not need to update every single table that uses it):

https://stackoverflow.com/questions/1771543/adding-a-new-value-to-an-existing-enum-type/7834949#7834949

tec27: Guess you can't really do the same to remove it in the reverse migration though, so maybe it's not really that much of a benefit

tec27 (changes requested):

server/migrations/20210803222419-add-2v2-matchmaking-type.js
tec27: Ah, annoying :( Guess we're stuck doing this forever then

client/matchmaking/browse-map-selections.jsx
```diff
@@ -256,22 +254,10 @@ export default class MapSelections extends React.Component {
     const { auth, matchmaking, matchmakingPreferences, type } = this.props
 
     const mapPool = matchmaking.mapPoolTypes.get(type)
-    const preferencesRecord = matchmakingPreferences.typeToPreferences.get(type)
+    const prefs = matchmakingPreferences.typeToPreferences.get(type)
```
tec27, line 257: Just a note, no need to fix in this PR, but when we update this code in the future to move to hooks, we probably want to select more granular things in the `useAppSelector` call. e.g. it's better to return this preference record from the `useAppSelector` call instead of returning the entire state object, since it will prevent re-renders when state in non-relevant parts of the state changes.

client/matchmaking/matchmaking-preferences-reducer.ts
```diff
@@ -21,7 +20,7 @@ export class MatchmakingPreferencesRecord extends ImmutableRecord({
   race: 'r' as RaceChar,
   mapPoolId: 1,
   mapPoolOutdated: false,
-  mapSelections: List<MapInfoJson>(),
+  mapSelections: List<typeof MapRecord>(),
```
tec27, line 23: don't think this is the right typing, You probably want `ReturnType<typeof MapRecord>` but a better thing to do here might be to just make `MapRecord` do the `extends Record` thing we do elsewhere so that you can use it as a type directly. I don't think it should change other usages of the type very much (or it will at the very least be easy to find with the TS compiler)

tec27 (approved):

client/matchmaking/matchmaking-preferences-reducer.ts
2Pacalypse-: Hmm, looks like this is not that easy without full TS-ification of the maps reducer, I think. When `MapRecord` extends `Record`, the typing here works, but then creation of the map record fails typecheck, as we're trying to create a `MapRecord` from `MapInfoJson`. Not really sure how to fix that, but I also saw that `maps2-reducer` doesn't even use the `MapRecord`, so maybe when we fully move to that it won't even matter?

tec27: yes, maps2-reducer is immer so these things are simpler. I don't see why creating a MapRecord from a MapInfoJson would be a problem here, could you link me to the code you're talking about? If there is a typing mismatch on the fields then MapRecord is probably just wrong, MapInfoJson is as accurate to what we actually store/pass as anything is (and with MapRecord I spent a bit less time trying to get things right).

If it *needs* to differ, you could also add a constructor/method on MapRecord to convert from a MapInfoJson until we can deleted it entirely.

PR #737, opened 2021-07-31
Move the matchmaking to a HTTP-based API backed by a service.

2Pacalypse-:


server/lib/matchmaking/matchmaking-api.ts
```diff
@@ -0,0 +1,124 @@
...
+      )
+      .post(
+        '/',
+        throttleMiddleware(matchmakingThrottle, ctx => String(ctx.session!.userId)),
+        acceptMatch,
+      )
+  }
+}
+
+async function findMatch(ctx: RouterContext) {
```
tec27, line 84: Don't have to fix this in this PR, but I'd *really* recommend moving to the API helper I wrote (so you don't have to call `validateRequest` manually and the response body is properly typed) + probably putting these methods on the `HttpApi` object (so you don't have to inject in every method)

server/lib/matchmaking/matchmaking-service.ts
```diff
@@ -259,13 +278,13 @@ export class MatchmakingApi {
...
           })
-          this.handleLeave(client)
+          this.removeClientFromMatchmaking(client)
         }
       }
     },
   }
 
-  matchmakerDelegate = {
+  private matchmakerDelegate = {
```
tec27, line 287: Not a problem you caused, but would probably be good to add explicit types for this + `gameLoaderDelegate`, think error messages would be better if the type definition changes

server/lib/matchmaking/matchmaking-service.ts
```diff
@@ -462,82 +490,49 @@ export class MatchmakingApi {
...
+    userId: number,
+    clientId: string,
+    type: MatchmakingType,
+    race: RaceChar,
+    useAlternateRace: boolean,
+    alternateRace: AssignedRaceChar,
+    preferredMaps: string[],
+  ) {
+    const userSockets = this.getUserSockets(userId)
+    const clientSockets = this.getClientSockets(userId, clientId)
```
tec27, line 517: This method doesn't do any validation that the client in question should be capable of entering matchmaking (e.g. that the client is an Electron client)

tec27: Note that I think you could make the argument that `activityRegistry` should be the thing to make that check, but it doesn't do so currently, either, and the error message/code here wouldn't be able to differentiate between "client isn't valid" and "user is already in a game"

server/lib/matchmaking/matchmaking-service.ts
```diff
@@ -575,74 +570,115 @@ export class MatchmakingApi {
...
+
+      cancelToken.cancel()
+
+      this.clientTimers = this.clientTimers.delete(client.name)
+    }
+
+    this.unregisterActivity(client)
+  }
+
+  private getUserSockets(userId: number): UserSocketsGroup {
```
tec27, line 642: Not something you caused, but it might be good to name these with `OrFail` as a postfix. We write these different ways sometimes, and it's really not obvious in the code using them whether or not they throw if the retrieval fails (and we've caused problems before by adding this sort of call somewhere, not realizing it will break when the user is offline, etc.). So I think e.g. `getUserSocketsOrFail` makes that a lot clearer, and makes us less likely to introduce such bugs

tec27 (approved):

server/lib/matchmaking/matchmaking-api.ts
2Pacalypse-: Gotta be honest here and mention that I actually prefer the syntax of explicitly calling the `validateRequest` in the handler body, instead of sending it as a parameter to a helper function. Purely from an aesthetic/readability perspective.

I like the benefit of using it to handle the API response though.

server/lib/matchmaking/matchmaking-service.ts
2Pacalypse-: Well, I'd say that's a good practice to have going forward as immediately as I renamed this method I noticed I'm calling it when unregistering an activity, which shouldn't really throw errors if it fails ^^.

PR #740, opened 2021-08-02
Add types to matchmaking events and actions.

2Pacalypse-:


2Pacalypse-:
Depends on https://github.com/ShieldBattery/ShieldBattery/pull/737

client/matchmaking/matchmaking-reducer.ts
```diff
@@ -1,52 +1,40 @@
...
+  randomMaps: Set<MapInfoJson>(),
+  chosenMap: undefined as MapInfoJson | undefined,
+}) {}
+
+export class MapPoolRecord extends Record({
+  id: '',
+  type: MatchmakingType.Match1v1,
+  startDate: new Date(),
+  maps: List<string>(),
+  byId: Map<string, MapInfoJson>(),
```
tec27, line 31: You're storing records in here, so that's not an accurate typing. At the very least this should be `Readonly<MapInfoJson>`, but even that is fairly fake/not true and you'd be better off using `MapRecord`, which is what you're actually storing

tec27: I also don't really know why we're storing the map info here at all, we already have a reducer that stores the map info by ID (`maps2-reducer`), you should really be handling the appropriate actions in there and only storing IDs here

client/matchmaking/socket-handlers.ts
```diff
@@ -163,57 +167,60 @@ const eventToAction = {
...
+      },
       setup: {
         gameId: event.setup.gameId,
         name: 'Matchmaking game', // Does this even matter for anything?
         map: event.chosenMap,
-        gameType: 'oneVOne',
-        slots: event.slots,
-        host: event.slots[0], // Arbitrarily set first player as host
+        gameType: GameType.OneVsOne,
+        slots: event.slots as PlayerInfo[],
```
tec27, line 188: extremely not a fan of these casts, if you're doing this either A) `event` is typed badly, or B) you're assuming that the thing that is being passed in the event will always have the same fields as the thing you're casting to. There's zero reason to assume that (and indeed, if the fields were exactly the same, this cast would be unnecessary in the first place), you should be creating an object of the correct type and mapping the fields manually

common/lobbies/slot.ts
```diff
@@ -2,8 +2,19 @@ import cuid from 'cuid'
...
+  ControlledOpen = 'controlledOpen',
+  ControlledClosed = 'controlledClosed',
+  UmsComputer = 'umsComputer',
+  Open = 'open',
+  Closed = 'closed',
+}
+
 export class Slot extends Record({
-  type: 'open',
+  type: undefined as SlotType | undefined,
```
tec27, line 17: Why make this potentially undefined? There is never a point where we expect a Slot to not have a known type, and in the previous iteration of this code that default type was 'open'. `type: SlotType.Open` would be the correct mapping of the previous code

common/lobbies/slot.ts
```diff
@@ -17,7 +28,7 @@ export class Slot extends Record({
 
 export function createOpen(race: RaceChar = 'r', hasForcedRace = false, playerId = 0): Slot {
   return new Slot({
-    type: 'open',
+    type: SlotType.Open as typeof SlotType.Open,
```
tec27, line 31: These casts are nonsensical. You're treating these like they're a union type, but there is no union type. There are no subclasses of `Slot`. Every one of these functions returns just a `Slot`, there is nothing to discriminate.

server/lib/matchmaking/matchmaking-service.ts
```diff
@@ -43,10 +44,10 @@ interface GameLoaderCallbacks {
     matchInfo: Match
     clients: List<ClientSocketsGroup>
     slots: List<Slot>
-    setup?: Partial<{
+    setup?: {
```
tec27, line 47: Don't really understand this change, why move from `undefined` meaning "this thing is not present" to `random value I've chosen` serving that purpose?

tec27 (changes requested):

server/lib/matchmaking/matchmaking-service.ts
2Pacalypse-: I thought `Partial` made all of the fields of the object optional, and seemed to me like `gameId` and `seed` are required to successfully start a game. Reverted it now though.

client/matchmaking/matchmaking-reducer.ts
2Pacalypse-: Gonna just fix the types for now, and then I'll move to using map IDs similar to what we do with users in a different PR.

server/lib/matchmaking/matchmaking-service.ts
tec27: It does. I don't get why this was ever optional in the first place and would be fine with you moving it away from being optional (and Partial) if you can verify that it was just a mistake. But moving from it defaulting to `{}` (and thus each of these values being `undefined`) to it defaulting to an object with dummy values for each of these keys (which is way harder and more confusing to verify as meaning to indicate "not present") was a bad change imo.

server/lib/matchmaking/matchmaking-service.ts
```diff
@@ -344,7 +345,7 @@ export class MatchmakingService {
       matchInfo,
       clients,
       slots,
-      setup = {},
+      setup = { gameId: '', seed: 0 },
```
tec27, line 348: So *this* was the part of the `Partial<>` change I had a problem with, and you didn't roll this back. Either roll this back as well, or do the necessary investigation to figure out why any of this is optional in the first place

2Pacalypse-: I'm just gonna revert it for now and investigate at some later point. Gonna file an issue for it.

PR #751, opened 2021-08-19
Add types to map actions and move its reducers to immer.

2Pacalypse-:


client/admin/map-pools-reducer.js
```diff
@@ -52,12 +51,7 @@ export default keyedReducer(new MapPoolsState(), {
 
     const history = new MapPoolHistory({
       mapPools: new List(payload.pools.map(p => p.id)),
-      byId: new Map(
-        payload.pools.map(p => [
-          p.id,
-          new MapPool({ ...p, maps: new List(p.maps.map(m => new MapRecord(m))) }),
-        ]),
-      ),
+      byId: new Map(payload.pools.map(p => [p.id, new MapPool({ ...p, maps: new List(p.maps) })])),
```
tec27, line 54: Seems like we're gonna wanna move this reducer to immer ASAP as well, I'm not totally sure of the effects of throwing these JS objects into a map directly

client/maps/action-creators.ts
```diff
@@ -0,0 +1,221 @@
...
+    }
+
+    dispatch({
+      type: '@maps/uploadLocalMapBegin',
+      payload: { path },
+    })
+
+    dispatch({
+      type: '@maps/uploadLocalMap',
+      payload: upload(path, apiUrl`maps`).then(({ map }: { map: MapInfoJson }) => {
```
tec27, line 33: You added a generic parameter to this, so you should use that instead of typing things in the `then` (and ideally the type specified would be a Payload type in `common/` that the server uses as well)

client/maps/action-creators.ts
```diff
@@ -0,0 +1,221 @@
...
+        if (onMapSelect) {
+          onMapSelect(map)
```
tec27, line 35: This is a bad place to do this because if `onMapSelect` throws, it will turn the action into a failure (even though it succeeded and the map is now on the server). I would instead pull the `upload` call out of the dispatch call and store the promise, pass it into the dispatch, and attach a `then` to handle this call that won't affect the result given to the reducers

client/maps/action-creators.ts
```diff
@@ -0,0 +1,221 @@
...
+        }
+
+        return map
+      }),
+      meta: { path },
+    })
+  }
+}
+
+export function getMapsList(
```
tec27, line 45: this should almost certainly just take an object directly, especially since you immediately convert it into an object

client/maps/action-creators.ts
```diff
@@ -0,0 +1,221 @@
...
+    dispatch({
+      type: '@maps/toggleFavoriteMapBegin',
+      payload: params,
+    })
+
+    const reqUrl = apiUrl`maps/favorites/${map.id}`
+    dispatch({
+      type: '@maps/toggleFavoriteMap',
+      payload: fetch<void>(reqUrl, { method: map.isFavorited ? 'DELETE' : 'POST' }).then<void>(
+        () => {
```
tec27, line 90: Should probably also be attached separately from the promise given to `dispatch` (see `regenMapImage` below)

client/maps/actions.ts
```diff
@@ -0,0 +1,341 @@
...
+    tileset: Tileset
+    searchQuery: string
+  }
+}
+
+/**
+ * A request is being made to the server to toggle the favorite status of a map.
+ */
+export interface ToggleFavoriteMapBegin {
+  type: '@maps/toggleFavoriteMapBegin'
```
tec27, line 130: nit: we already know this is for `@maps`, can't we just shorten this to `toggleFavoriteBegin` and `toggleFavorite`?

client/maps/browse-server-maps.jsx
```diff
@@ -25,7 +25,7 @@ import { BrowserFooter as Footer } from './browser-footer'
 import MapPreview from './map-preview'
 import { MapThumbnail } from './map-thumbnail'
 
-const MAPS_LIMIT = 30
+const MAPS_LIMIT = 10
```
tec27, line 28: Intentional?

client/maps/local-maps-reducer.ts
```diff
@@ -0,0 +1,32 @@
...
+export interface LocalMapsState {
+  /** A flag indicating whether we're currently uploading a local map. */
+  isUploading: boolean
+  /** A potential error of our request to upload a local map. */
+  lastError?: FetchError
+}
+
+const DEFAULT_STATE: Immutable<LocalMapsState> = {
+  isUploading: false,
+  lastError: undefined as FetchError | undefined,
```
tec27, line 14: We don't need to do this any more, the TS type encodes this

client/maps/map-details-reducer.ts
```diff
@@ -0,0 +1,50 @@
...
+  /** A map that was requested, with all of the details included. */
+  map?: MapInfoJson
+  /** A flag indicating whether we're currently uploading a local map. */
+  isRequesting: boolean
+  /** A potential error of our request to upload a local map. */
+  lastError?: FetchError
+}
+
+const DEFAULT_STATE: Immutable<MapDetailsState> = {
+  map: undefined as MapInfoJson | undefined,
```
tec27, line 19: not necessary in this file either

client/maps/map-preferences-reducer.ts
```diff
@@ -0,0 +1,70 @@
...
+}
+
+const DEFAULT_STATE: Immutable<MapPreferencesState> = {
+  visibility: MapVisibility.Private,
+  thumbnailSize: 0,
+  sortOption: MapSortType.Name,
+  numPlayersFilter: new Set(),
+  tilesetFilter: new Set(),
+  isRequesting: false,
+  lastError: undefined as FetchError | undefined,
```
tec27, line 31: not necessary here either

client/maps/map-preferences-reducer.ts
```diff
@@ -0,0 +1,70 @@
...
+}
+
+function createPreferences(state: MapPreferencesState, preferences: MapPreferences) {
```
tec27, line 34: This doesn't really create anything, `updatePreferences` is probably a better name

client/maps/maps-reducer.ts
```diff
@@ -0,0 +1,126 @@
...
+  lastError?: FetchError
+}
+
+const DEFAULT_STATE: Immutable<MapsState> = {
+  byId: new Map(),
+  favoritedById: new Map(),
+  total: -1,
+  isRequesting: false,
+  favoriteStatusRequests: new Set(),
+  lastError: undefined as FetchError | undefined,
```
tec27, line 27: not necessary here either

client/maps/maps-reducer.ts
```diff
@@ -0,0 +1,126 @@
...
+  /** A set of map IDs that the user is currently (un)favoriting. */
+  favoriteStatusRequests: Set<string>
+  /** A potential error of our request to get maps. */
+  lastError?: FetchError
+}
+
+const DEFAULT_STATE: Immutable<MapsState> = {
+  byId: new Map(),
+  favoritedById: new Map(),
+  total: -1,
```
tec27, line 24: would probably prefer `0` over `-1` for stuff like this, it functions better if we end up with that value at some point (e.g. `if (!state.total)` will function properly with `0`, but not `-1`)

tec27 (changes requested):

client/maps/browse-server-maps.jsx
2Pacalypse-: No :(

client/admin/map-pools-reducer.js
2Pacalypse-: I'll keep it near the top of my TODO list :d

client/maps/action-creators.ts
```diff
@@ -0,0 +1,248 @@
...
+import { ThunkAction } from '../dispatch-registry'
+import fetch from '../network/fetch'
+import { apiUrl } from '../network/urls'
+import { openSnackbar } from '../snackbars/action-creators'
+import { ClearMaps } from './actions'
+
+const uploadModule = IS_ELECTRON ? import('./upload') : Promise.resolve()
+
+export function uploadLocalMap(path: string, onMapSelect: (map: MapInfoJson) => void): ThunkAction {
+  return async dispatch => {
```
tec27, line 22: Sorry I didn't notice this before, don't make thunks that return Promises, this isn't safe. Nothing will be attached to handle errors on this. You seemed to have realized this since you're no longer awaiting anything, but the async keyword still needs to be removed

client/maps/action-creators.ts
```diff
@@ -0,0 +1,248 @@
...
+    const uploadPromise = uploadModule.then(upload =>
+      upload && upload.default
+        ? upload.default<UploadMapPayload>(path, apiUrl`maps`)
```
tec27, line 25: This code is a lot messier than just pulling this out into a function, e.g.
```ts
async function uploadMap(filePath: string) {
  if (IS_ELECTRON) {
    const module = await import('./upload')
    return module?.default<UploadMapPayload>(filePath, apiUrl`maps`)
  } else {
    throw new Error('cannot upload maps on non-electron clients')
  }
}

// ... in action creator
const promise = uploadMap(foo)
promise.then(/* ... */)
```

tec27 (approved):

PR #753, opened 2021-08-25
Add a disabled overlay to create/join lobby, and find-match overlays

2Pacalypse-:
when in a party (and under certain conditions in case of find-match).

### DIFF
--- a/client/matchmaking/find-match.tsx
+++ b/client/matchmaking/find-match.tsx
@@ -1,23 +1,24 @@
 import { Immutable } from 'immer'
-import { rgba } from 'polished'
 import React, { useCallback, useRef, useState } from 'react'
 import styled from 'styled-components'
 import { assertUnreachable } from '../../common/assert-unreachable'
 import { MatchmakingPreferences, MatchmakingType } from '../../common/matchmaking'
 import { closeOverlay } from '../activities/action-creators'
+import { DisabledOverlay } from '../activities/disabled-content'
+import { useSelfUser } from '../auth/state-hooks'
 import { ComingSoon } from '../coming-soon/coming-soon'
 import KeyListener from '../keyboard/key-listener'
 import { RaisedButton } from '../material/button'
 import { useScrollIndicatorState } from '../material/scroll-indicator'
 import { TabItem, Tabs } from '../material/tabs'
 import { useAppDispatch, useAppSelector } from '../redux-hooks'
-import { colorDividers, dialogScrim } from '../styles/colors'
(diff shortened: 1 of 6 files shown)


client/matchmaking/find-match.tsx
```diff
@@ -160,6 +155,25 @@ export function FindMatch() {
       contents = assertUnreachable(activeTab)
   }
 
+  let disabledContents: React.ReactNode | undefined
```
tec27, line 158: nit: generally prefer to pull code flows like this into their own component that can do early returns or return a ternary or something

client/matchmaking/matchmaking-disabled-card.tsx
```diff
@@ -62,15 +43,12 @@ const CountdownItemText = styled.span`
   color: ${colorTextSecondary};
 `
 
-export interface ConnectedMatchmakingDisabledCardProps {
+export interface DisabledCardProps {
```
tec27, line 46: This rename seems fairly confusing given that there is now a separate component called `DisabledCard` and these are not props for it

client/matchmaking/matchmaking-disabled-card.tsx
```diff
@@ -152,6 +130,35 @@ export function ConnectedMatchmakingDisabledCard({
...
+    </DisabledCard>
+  )
+}
+
+export interface PartyDisabledCardProps extends DisabledCardProps {
+  isPartyLeader: boolean
+  partySize: number
+}
+
+export function PartyDisabledCard({
```
tec27, line 142: imo these components should either all be connected or none of them should be connected, it's confusing that this one needs all these props and the other one pulls the info out of the store itself

client/matchmaking/matchmaking-disabled-card.tsx
```diff
@@ -152,6 +130,35 @@ export function ConnectedMatchmakingDisabledCard({
       ) : (
         <Headline6>Soon™</Headline6>
       )}
-    </Contents>
+    </DisabledCard>
+  )
+}
+
+export interface PartyDisabledCardProps extends DisabledCardProps {
```
tec27, line 137: I see zero relation between these two components, so I'm not sure why you felt the need to extend this interface. Just because they have the same props under the same names doesn't mean they should have this kind of relation, just write `className` and `type` here as well (but honestly `type` should probably be `matchmakingType` in this context)

tec27 (changes requested):

PR #757, opened 2021-09-25
Move the chat reducer to immer.

2Pacalypse-:


client/chat/chat-reducer.ts
```diff
@@ -64,248 +66,271 @@ function updateMessages(
...
+
+    const channel = state.byName.get(lowerCaseChannelName)
+    if (!channel) {
+      return
     }
 
-    return updateMessages(updated, channel, false, messages => newMessages.concat(messages))
+    // Even though the payload here is `ServerChatMessage`, we expand its type so it can be
+    // concatenated with the existing messages which could also contain client chat messages.
+    const newMessages = action.payload as ChatMessage[]
```
tec27, line 258: kinda surprised this would be necessary to do at all, is it because the type field is a different enum or something?

2Pacalypse-: I don't know the exact reason to be perfectly honest. Your theory sounds likely to me though.

client/navigation/action-creators.js
```diff
@@ -12,7 +12,7 @@ export function goToIndex(transitionFn = push) {
     if (lobby.inLobby && IS_ELECTRON) {
       transitionFn(`/lobbies/${encodeURIComponent(lobby.info.name)}`)
     } else if (channels.size) {
-      transitionFn(`/chat/${encodeURIComponent(channels.first())}`)
+      transitionFn(`/chat/${encodeURIComponent([...channels][0])}`)
```
tec27, line 15: Not sure what the intention here is, but I'm pretty sure this is just an array and creating a new array just to grab the first value and throw it away seems pretty unnecessary?

client/navigation/connected-left-nav.jsx
```diff
@@ -34,7 +34,7 @@ function stateToProps(state) {
   return {
     activeGame: state.activeGame,
     auth: state.auth,
-    chatChannels: state.chat.channels.map(c => ({
+    chatChannels: Array.from(state.chat.channels.values()).map(c => ({
```
tec27, line 37: this is already an array, no? You can just call `map` directly on it. (If it's *not* and you still need to use `Array.from`, you can remove the call to `map` and just use the second parameter of `Array.from` to do the mapping)

common/chat.ts
```diff
@@ -1,33 +1,80 @@
...
 }
 
-// TODO(2Pac): Include more information here, e.g. channel permissions, join date, etc.
-export interface ChatUser {
-  id: SbUserId
-  name: string
+/** A message that is displayed in the chat when someone joins the channel. */
+export interface JoinChannelMessage extends BaseChatMessage {
+  type: typeof ClientChatMessageType.JoinChannel
+  userId: SbUserId
```
tec27, line 42: Unsure if these are new structures or moved from somewhere else or whatever, but probably this is not enough information for this particular message (this would be a user we do not necessarily have a name, etc. for, so it probably needs full info)

client/chat/action-creators.ts
```diff
@@ -70,7 +70,7 @@ export function getMessageHistory(channel: string, limit: number): ThunkAction {
     }
 
     const chanData = byName.get(lowerCaseChannel)!
-    const earliestMessageTime = chanData.messages.first({ time: -1 }).time
+    const earliestMessageTime = chanData.messages[0]?.time ?? -1
```
tec27, line 73: This code is too clever, prefer a ternary where you check the length and only access the first index if you're sure data is there

tec27 (approved):

client/chat/channel.tsx
```diff
@@ -381,16 +385,16 @@ export default function Channel(props: ChatChannelProps) {
...
+        idle: [],
+        offline: [],
       }
     }
 
     return {
-      active: channelUsers.active.toList().sort(sortUsers),
-      idle: channelUsers.idle.toList().sort(sortUsers),
-      offline: channelUsers.offline.toList().sort(sortUsers),
+      active: Array.from(channelUsers.active.values()).sort(sortUsers),
```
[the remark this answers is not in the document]
2Pacalypse-: These are actually `Map`s in the reducer, we're only using them as arrays in this component.

client/navigation/action-creators.js
2Pacalypse-: This is actually a `Set` in the reducer, but now that I think about it, this should probably be an `OrderedSet`, since getting the "first" element from it doesn't make much sense otherwise. We could also make the list of channels that the user has joined an `array`, but then we lose the uniqueness guarantee. Though I guess if it ever happens that we try to insert the same channel into the array twice, then something else is broken  ^^

client/navigation/connected-left-nav.jsx
2Pacalypse-: No, it's a `Set`; see the comment above.

common/chat.ts
2Pacalypse-: Yeah, these are just the structures that were moved from the existing records that we used before. I didn't feel like reevaluating them at this moment to be honest, but you're right, this particular structure should include more information.

It's not so bad at the moment since we're returning user infos for all users when we retrieve the user list for a channel, but that doesn't cover the case of displaying this message when someone leaves the channel (what #758 enables), or even if we allow hiding the user list in some way in future.

tec27: I actually think this is okay now that I'm looking at it today. I was thinking this was the event, but this is just the output of the event.

So the event itself will result in us population the user reducer state, this message just refers to that state indirectly (and referring to that indirectly is definitely better in case some state (e.g. their name) changes in the meantime)

PR #758, opened 2021-09-28
Persist the "join channel" message on the server to the DB.

2Pacalypse-:
This makes it so the "join channel" message is visible even after
reloading the app. Note that we're only persisting the "join channel"
message for now, and not the "leave channel" message. This makes it
possible to a person to join and leave a channel multiple times, and
other users will only see the "join channel" message when they reload,
which might be a bit weird, but doesn't seem like a big deal to me atm.


2Pacalypse-:
Depends on #757

tec27 (approved):

tec27:
Before merging, I just want to verify that you've tested this with old clients (that is, run the new server code with the old client code from before this change, and have that old client receive some persisted joins). It seems like it should work okay, but I'd just like to make sure of that now and not find it out when we deploy to production.

2Pacalypse-:
I'm not 100% sure I tested it correctly, but I think it works fine from what I saw :d

PR #759, opened 2021-10-03
Add server-side code to moderate users in chat channels.

2Pacalypse-:
"moderate" means kick or ban for now. Also added a global user
permission that makes someone a super moderator, i.e. allows them to
kick/ban people in all chat channels (except #ShieldBattery).

Users with "editPermissions" channel permission can kick/ban anyone in
that channel, while users with "kick"/"ban" channel permission can only
kick/ban people that don't have "editPermissions", "kick", or "ban"
permission.

TODO: We might want to introduce a concept of a channel "owner", which
makes them unkickable/unbannable even by other users with
"editPermissions" permission. Only super moderators with a global
moderation permissions should be able to moderate them.


2Pacalypse-:
Depends on #758

server/lib/chat/chat-service.ts
```diff
@@ -174,6 +181,90 @@ export default class ChatService {
...
+      !this.state.users.has(targetId) ||
+      !this.state.users.get(targetId)!.has(originalChannelName)
+    ) {
+      throw new ChatServiceError(
+        ChatServiceErrorCode.InvalidModerationAction,
+        'User must be in channel to moderate them',
+      )
+    }
+
+    const targetSockets = this.getUserSockets(targetId)
```
tec27, line 248: This doesn't seem safe to do to me, it implies we can't kick or ban people that are currently offline

server/lib/chat/chat-service.ts
```diff
@@ -174,6 +181,90 @@ export default class ChatService {
...
+    // NOTE(2Pac): New owner can technically be selected if a global moderator removes all users
+    // with channel permissions.
+    const newOwner = await this.removeUserFromChannel(originalChannelName, targetSockets)
+
+    if (moderationAction === ChannelModerationAction.Ban) {
+      await banUserFromChannel(originalChannelName, moderatorId, targetId, moderationReason)
```
tec27, line 254: This should happen first imo, #1 because it may fail and  that should probably stop the whole process here, and #2 because if you kick before banning, there's a timing window where the user can rejoin and throw off a lot of expectations here

server/lib/chat/chat-service.ts
```diff
@@ -174,6 +181,90 @@ export default class ChatService {
...
+      throw new ChatServiceError(
+        ChatServiceErrorCode.InvalidModerationAction,
+        'User must be in channel to moderate them',
+      )
+    }
+
+    const targetSockets = this.getUserSockets(targetId)
+    // NOTE(2Pac): New owner can technically be selected if a global moderator removes all users
+    // with channel permissions.
+    const newOwner = await this.removeUserFromChannel(originalChannelName, targetSockets)
```
tec27, line 251: this should really be done in a way that notifies the user they've been kicked, shouldn't be handled in a way where that is ambiguous to leaving (because the user may have multiple clients connected, so the clients really can't tell either way)

server/lib/chat/chat-service.ts
```diff
@@ -321,6 +412,24 @@ export default class ChatService {
     user.unsubscribe(getChannelPath(channelName))
   }
 
+  private async removeUserFromChannel(
```
tec27, line 415: As per above, this makes it so users can't be removed unless they're currently online. I think you should rewrite this to take a user ID and find the userSockets as part of the logic here (optionally, if they aren't found you just skip that part)

tec27 (changes requested):

PR #765, opened 2021-10-19
Implement basic support for user mentions in various chat services.

2Pacalypse-:
Two things are currently implemented:
- Highlight text messages if they mention logged-in user
- Make mentions clickable in chat (opens user profile overlay)

User is considered mentioned in any text message contains user's name
preceded by the @ character. This applies to chat channels, lobby chat,
party chat, and whispers.

This PR also includes the code for setting the special icon in the
system tray. But first we need to make one. That might happen in this
PR, or at a later point.


server/lib/messaging/process-chat-message.ts
```diff
@@ -0,0 +1,36 @@
...
+ * can then easily be used on the client to replace the markup with the appropriate UI.
+ *
+ * Any mention that doesn't map to a valid user in our system is left unreplaced.
+ *
+ * @returns A tuple of the processed chat message, along with a map of userId -> user of each
+ *   mentioned user.
+ */
+export async function processMessageContents(
+  text: string,
+  allowedMentionUsers: Set<string>,
```
tec27, line 17: I haven't reviewed the rest of this yet, but I'll mention this now since it's something you can change earlier:

I don't think this is desirable or good. I think mentioning people that aren't present is fine, it makes the username clickable, you can use it to whisper them, access their profile, etc. Alongside that, doing this kind of filtering can be VERY expensive.

With this change, you've made it so we have to look up all the users in #ShieldBattery on *every message* to #ShieldBattery, so we're grabbing and returning thousands of users every single time, converting all their names to lowercase, putting them in a Set... That's gonna wreak havoc on stuff and make everything slower.

Just leave this how it was before, there's really no harm in mentioning a user that isn't present. It's a different story when we're dealing with what to auto-complete and what triggers a notification, but this code here really has nothing to do with either of those things.

client/messaging/common-message-layout.test.tsx
```diff
@@ -0,0 +1,328 @@
+import React from 'react'
+import ShallowRenderer from 'react-test-renderer/shallow'
+import { SbUserId } from '../../common/users/user-info'
+import { TextMessage } from './common-message-layout'
+
+describe('client/messaging/common-message-layout/TextMessage', () => {
+  const renderer = ShallowRenderer.createRenderer()
+  const doMatch = (text: string): React.ReactElement => {
```
tec27, line 8: nit: this isn't really doing any matching, maybe `doRender` or something would be more appropriate

client/messaging/common-message-layout.test.tsx
```diff
@@ -0,0 +1,328 @@
...
+    renderer.render(
+      <TextMessage userId={2 as SbUserId} selfUserId={1 as SbUserId} time={0} text={text} />,
```
tec27, line 10: 3 things:
1) don't cast to `SbUserId` like this, use `makeSbUserId`, that's what it's for
2) The value for `selfUserId` here should probably be pulled into a constant at the top level so it can be easily referenced in tests that want to check highlighting behavior
3) You don't have tests for highlighting behavior atm (in the case that the message *is* highlighted, at least), and those would definitely be good to have

client/messaging/common-message-layout.test.tsx
```diff
@@ -0,0 +1,328 @@
...
+            />
+          </common-message-layout__UserMention>
+          
+        </common-message-layout__Text>
+      </TimestampMessageLayout>
+    `)
+  })
+
+  test('message with a link containing a mention', () => {
+    expect(doMatch('http://www.<@123>.com')).toMatchInlineSnapshot(`
```
tec27, line 186: I appreciate testing if there's a mention inside the link, but this is not a valid domain and thus this test's result actually seems incorrect to me. The test for this particular case should probably place the mention in the path of the URL (rather than the domain), and we should ensure that the link matching stuff ignores invalid characters.

Valid characters are: `[A-z0-9-]` (this may be better handled in the url matching tests, I dunno). Other restriction on that seems to be it must be at least 3 characters, the first character must be a letter or number, and the last character must be a letter or a number (so not a dash).

client/messaging/common-message-layout.tsx
```diff
@@ -34,55 +36,84 @@ const Text = styled.span`
...
-    const matches = matchLinks(text)
+const UserMention = styled.span`
+  background-color: ${rgba(blue400, 0.48)};
+
+  &:hover {
+    background-color: ${rgba(blue400, 0.87)};
+  }
+`
+
+function* getAllMatches(text: string) {
```
tec27, line 47: nice :)

client/messaging/common-message-layout.tsx
```diff
@@ -34,55 +36,84 @@ const Text = styled.span`
...
+        elements.push(
+          <a key={match.index} href={match.text} target='_blank' rel='noopener nofollow'>
+            {match.text}
+          </a>,
+        )
+      }
 
-      lastIndex = match.index! + match[0].length
+      lastIndex = match.index! + match.text.length
     }
```
tec27, line 97: I think since these types are now effectively a discriminated union, you can add an `else` here with a call to `assertUnreachable`. (For reference, the thing you pass into `assertUnreachable` is the thing you expect TS to determine as being `never` at that point, so here it is probably `assertUnreachable(match)`)

client/messaging/common-message-layout.tsx
```diff
@@ -34,55 +36,84 @@ const Text = styled.span`
...
-    for (const match of matches) {
+    for (const match of sortedMatches) {
       // Insert preceding text, if any
       if (match.index! > lastIndex) {
         elements.push(text.substring(lastIndex, match.index))
       }
 
-      elements.push(
+      if (match.type === 'mentionMarkup') {
+        const userId = Number(match.groups.userId) as SbUserId
```
tec27, line 73: Same as in the tests, basically never do this cast. It works *now* but if the type changes or we change how we generate the opaque type, it may not. *Always* use `makeSbUserId`, it makes the intention very clear and can be updated to deal with implementation changes in one place.

client/messaging/common-message-layout.tsx
```diff
@@ -34,55 +36,84 @@ const Text = styled.span`
...
+    const matches = getAllMatches(text)
+    const sortedMatches = Array.from(matches).sort((a, b) => a.index - b.index)
     const elements = []
     let lastIndex = 0
+    let isHighlighted = false
 
-    for (const match of matches) {
+    for (const match of sortedMatches) {
       // Insert preceding text, if any
       if (match.index! > lastIndex) {
```
tec27, line 68: nit: think the `!` is unnecessary now and should be removed

client/messaging/common-message-layout.tsx
```diff
@@ -34,55 +36,84 @@ const Text = styled.span`
...
-      )
+        elements.push(
+          <a key={match.index} href={match.text} target='_blank' rel='noopener nofollow'>
+            {match.text}
+          </a>,
+        )
+      }
 
-      lastIndex = match.index! + match[0].length
+      lastIndex = match.index! + match.text.length
```
tec27, line 96: Same as above, I think this `!` is unnecessary. If it *is* necessary, it seems like something may be screwed up about the types here.

client/messaging/common-message-layout.tsx
```diff
@@ -34,55 +36,84 @@ const Text = styled.span`
...
+  const { userId, selfUserId, time, text } = props
+  const [parsedText, isHighlighted] = useMemo(() => {
+    const matches = getAllMatches(text)
+    const sortedMatches = Array.from(matches).sort((a, b) => a.index - b.index)
     const elements = []
     let lastIndex = 0
+    let isHighlighted = false
 
-    for (const match of matches) {
+    for (const match of sortedMatches) {
```
tec27, line 66: This code doesn't seem to handle the case that `match.index < lastIndex`, which seems necessary. Not sure if we don't have any tests that trigger this (it seems like you *tried* to have that, but maybe the mention pattern is restrictive enough that it didn't actually trigger like you thought). In any case, if this occurs, the match should be skipped entirely (as it overlaps with a previous match).

Thinking more about the mention pattern, I can't think of a way to test that currently, but this logic should exist so we don't get surprised in the future when we add other patterns.

client/messaging/connected-username.tsx
```diff
@@ -20,7 +20,7 @@ const Username = styled.span`
  * This component is connected to the store where it tries to find the user. All the services using
  * it should ensure that the user is loaded in the store properly.
  */
-export function ConnectedUsername(props: { userId: SbUserId }) {
+export function ConnectedUsername(props: { userId: SbUserId; isMention?: boolean }) {
```
tec27, line 23: Setting a default value for `isMention` here is preferable to just allowing it to be undefined, imo. It documents the behavior better in the case that a value isn't provided.

client/profile/user-reducer.ts
```diff
@@ -83,6 +83,10 @@ export default immerKeyedReducer(DEFAULT_STATE, {
     state.usernameToId.set(user.name, user.id)
   },
 
+  ['@chat/updateMessage'](state, action) {
+    updateUsers(state, [action.payload.user, ...action.payload.mentions])
```
tec27, line 87: Prefer just calling `updateUsers` twice, once with each array. Less work to not allocate a new array and move all the references.

Apply throughout this reducer.

common/text/mentions.ts
```diff
@@ -0,0 +1,83 @@
...
+ * mentions pattern, it doesn't also verify that the mentioned users actually exist in the system.
+ *
+ * @returns A generator of matches for mentions, where each match includes a named capture group
+ *   called "username" which contains just the matched username of the user.
+ */
+export function* matchUserMentions(text: string): Generator<{
+  type: 'mention'
+  text: string
+  index: number
+  groups: Record<'prefix' | 'username' | 'postfix', string>
```
tec27, line 44: nit: Prefer an interface for this instead of using `Record` in this way. The errors will likely be a lot better and it's less unwieldy and just generally easier to understand

common/text/mentions.ts
```diff
@@ -0,0 +1,83 @@
...
+/**
+ * Matches all user mentions in a given text. The user is considered mentioned if the text contains
+ * their name, preceded by the @ sign. Note that this function only matches things that fit the
+ * mentions pattern, it doesn't also verify that the mentioned users actually exist in the system.
+ *
+ * @returns A generator of matches for mentions, where each match includes a named capture group
+ *   called "username" which contains just the matched username of the user.
+ */
+export function* matchUserMentions(text: string): Generator<{
+  type: 'mention'
```
tec27, line 41: You might find it better for errors and things to extract these return types (that is, the generic param to `Generator` here) to a named interface. Also might be good if you need to pass one of these around at some point

server/lib/chat/chat-api.ts
```diff
@@ -144,7 +145,13 @@ export class ChatApi {
 
   @httpGet('/:channelName/messages')
   @httpBefore(throttleMiddleware(retrievalThrottle, ctx => String(ctx.session!.userId)))
-  async getChannelHistory(ctx: RouterContext): Promise<ServerChatMessage[]> {
+  getChannelHistoryOld(ctx: RouterContext) {
```
tec27, line 148: Probably want to leave a comment here for why this exists

server/lib/chat/chat-service.ts
```diff
@@ -360,6 +367,20 @@ export default class ChatService {
           return assertUnreachable(m.data)
       }
     })
+
+    let mentionIds = Set<SbUserId>()
+    for (const msg of dbMessages) {
+      if (msg.data.type === ServerChatMessageType.TextMessage) {
```
tec27, line 373: imo do this logic when mapping the `TextMessage` above. (It's a bit weird to have an extra side effect in a `map`, but I think it's preferable to iterating the whole list of messages twice)

server/lib/chat/chat-service.ts
```diff
@@ -360,6 +367,20 @@ export default class ChatService {
           return assertUnreachable(m.data)
       }
     })
+
+    let mentionIds = Set<SbUserId>()
```
tec27, line 371: would probably prefer to use a native Set here, the immutable one really provides no benefits here

client/messaging/common-message-layout.test.tsx
2Pacalypse-: Hmm, you're right that this is better handled in the link tests. I'll delete this test from here and open an issue to better handle link matching.

client/messaging/connected-username.tsx
```diff
@@ -4,12 +4,15 @@ import { SbUserId } from '../../common/users/user-info'
 import { useAnchorPosition } from '../material/popover'
 import { ConnectedUserProfileOverlay } from '../profile/user-profile-overlay'
 import { useAppSelector } from '../redux-hooks'
+import { blue100 } from '../styles/colors'
 
-const Username = styled.span`
+const Username = styled.span<{ $isMention: boolean }>`
```
tec27, line 9: Think you'd be better off passing through a `className` prop to this element and styling this how you want in the mention code. Otherwise every time we need a different style we'll have to add more props here, or a whole enum, etc. The typical styled-components pattern is more the subclassing thing (e.g. `styled(ConnectedUsername)`).

server/lib/chat/chat-service.ts
```diff
@@ -340,45 +341,46 @@ export default class ChatService {
...
+          messages.push({
+            id: msg.msgId,
+            type: msg.data.type,
+            channel: msg.channelName,
+            from: msg.userId,
+            time: Number(msg.sent),
+            text: msg.data.text,
+          })
+          users.push({ id: msg.userId, name: msg.userName })
+          mentionIds = new global.Set<SbUserId>([...mentionIds, ...(msg.data.mentions || [])])
```
tec27, line 360: so when I was saying to use the built-in Set I didn't mean to also use it as if it were immutable :P

This is gonna be horribly inefficient, just mutate the existing set

server/lib/whispers/whisper-service.ts
```diff
@@ -165,28 +165,31 @@ export default class WhisperService {
...
+
+    for (const msg of dbMessages) {
+      messages.push({
+        id: msg.id,
+        from: msg.from,
+        to: msg.to,
+        sent: Number(msg.sent),
+        data: msg.data,
+      })
+      mentionIds = new global.Set<SbUserId>([...mentionIds, ...(msg.data.mentions || [])])
```
tec27, line 186: Same as the other file, mutate the existing set

PR #772, opened 2021-11-01
Confirmation dialog when clicking links in chat

### LINKED ISSUE #739: Add a confirmation dialog when clicking links in chat
We should confirm with users that they're okay visiting a particular untrusted host (which would be everything except for our own host(s) ). Ideally we'd also let them mark particular hosts as trusted, and save that trust locally. Discord is a decent model for the right behavior here.

T1mL3arn:
Closes #739.

Dialog has two options:

- trust this host
- trust all links (i.e. never show this dialog again)

Either of them can be chosen or none, and both options can be changed later in shb settings.

Bonus: generic checkable component that can be used as a radio and as a checkbox


2Pacalypse-:
Hey, thanks for the PR!

I haven't yet looked too deeply into it yet, but a few of the general comments on things that I've noticed:
- TypeScript should be used in all of the new files that are created (we're trying to move the whole codebase to TypeScript, so it doesn't make much sense to introduce even more files that we'll have to convert eventually)
- We encourage the usage of functional components and hooks for all of the new components (among other things, this makes it easier to type stuff)
- We mostly use Material Design specification in our design, so that should be kept in mind when changing the design of components in the "material" folder (namely in this PR, the things like the "Compact dialog" and the way the buttons are displayed are probably not up to spec)
- I'm not too confident on having the "trust all links" option in the first place, but I'll leave that decision to @tec27
- I'm also genuinely not sure what's the best place to keep the trusted links; might even be the server if we want the user's choice to be remembered across machines?
- Do we even need a place in the settings dialog to see the trusted links and be able to "untrust" them? Seems nice I guess, but even Discord doesn't have that afaik :d

tec27:
> * I'm not too confident on having the "trust all links" option in the first place, but I'll leave that decision to @tec27
Yes, I agree that this feature isn't desirable

> I'm also genuinely not sure what's the best place to keep the trusted links; might even be the server if we want the user's choice to be remembered across machines?
Local settings seems fine to me, this is where Discord stores it as well

T1mL3arn:
>`"trust all links"` feature isn't desirable

Lacking of such feature sounds to me as the statement "you are 10 years old child who has no right to open any link without permission" :smile: Having click-link dialog in general is good, but better if a user can turn it off - so here the way to disable the dialog. With Discord as an example - it does not have such feature and some people on their support forums complain about it, that can be summarized as "give us more control" ([\_1\_](https://support.discord.com/hc/en-us/community/posts/360054714992-Add-ability-to-disable-links-are-spoopy-warning-universally-replace-with-tooltip-with-URL), [\_2\_](https://support.discord.com/hc/en-us/community/posts/360033209411--Trust-this-domain-whitelist), [\_3\_](https://support.discord.com/hc/en-us/community/posts/1500001007901-Pop-up-on-mobile-when-clicking-links-is-now-appearing-on-every-link), [\_4\_](https://support.discord.com/hc/en-us/community/posts/4409167153047--Are-you-sure-hold-up-message-for-links-as-an-option-under-privacy-and-safety)). More on that, Discord does not show the dialog for plain links - it asks confirmation for previews only(?) (tested in browser)

>TypeScript should be used

Ok, will update it.

>things like the "Compact dialog" and the way the buttons are displayed are probably not up to spec

This is true but I believe the spec does not suit this case well. Maybe update the spec with sort-of small dialog?

>Do we even need a place in the settings dialog to see the trusted links and be able to "untrust" them? Seems nice I guess, but even Discord doesn't have that afaik

Yes, and some people was needed to untrust domains, so their support forum has an instruction for that (going into browser storage and manually delete the domains)

tec27:
I appreciate the attempt to contribute, but arguing with our decisions, intended UX, etc. isn't really welcome. If you're going to be hostile towards the needs of our project and send PRs that don't do what we asked, it's not really going to be worth our time to review or merge these things.

If you'd like to do what we asked and remove the unwanted parts, we can review that, otherwise we can avoid wasting each others' time further

T1mL3arn:
Need a clarification. Do `trust all links` setting is under `dont argue  with our decisions` condition  and it should be removed ?

2Pacalypse-:
Let me just say, and this is something that we'll definitely need to include in our contributing guide (once we actually write one :d); while we definitely appreciate outside contributions, it's very hard for us to review and approve big changes. Especially when it comes from a newer person who might not yet be familiar with the way we do things.

So there are 2 things that are probably most important when it comes to contributing:
- Keep things small and manageable. Not that we want to discourage big changes completely, but it's much easier to properly review and test PRs when they're smaller and focused on just one thing.
- And perhaps even more important, ask questions! A lot of wasted work can be avoided by just asking questions (either here on GitHub or on Discord), so we can figure out what should be done before actually doing it.

> Need a clarification. Do `trust all links` setting is under `dont argue with our decisions` condition and it should be removed ?

Yeah, "Trust all links" should be removed. While I understand the concerns it's trying to address, I feel like we can deal with it in other ways in future. If people complain it's too spammy, we can show the dialog only for certain links that we detect in some way as sketchy, which seems to be what Discord is doing now.

> This is true but I believe the spec does not suit this case well. Maybe update the spec with sort-of small dialog?

Yeah, perhaps. However I would leave it out of this PR and figure out if there's a need for non-spec compact dialog as a separate issue.

> Yes, and some people was needed to untrust domains, so their support forum has an instruction for that (going into browser storage and manually delete the domains)

While I generally agree with this, the problem is that the current settings dialog doesn't really allow for this to be displayed in an intuitive place. This definitely feels like an advanced type of setting, and here we are displaying it to users as the first thing they see when they open the settings dialog. The idea is to eventually redesign the settings dialog which allows more tabs, and then we could have this setting in a better suited place. But until that happens, I would take out this feature as well out of this PR, and perhaps make a separate PR just for that which we can revisit once we redesign the settings dialog.

T1mL3arn:
Brief summary:

1. <details>
    <summary>How does it look like</summary>

    untrusted link dialog

  </details>

2.  `trust all links` setting is removed
3. Removed ability to manage trusted hosts

Right now it is more-less the same as in Discord.

2Pacalypse- (changes requested):

client/material/checkable-input.tsx
```diff
@@ -0,0 +1,242 @@
+import { rgba } from 'polished'
```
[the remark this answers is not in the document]
T1mL3arn: > it should be implemented with proper styling

Just want to mention, this is a simplest attempt to use existing component as a radio button, making its square box to be a circle with `border-radius: 50%`. Though checkmark remains the same as with checkbox.

client/messaging/external-link.tsx
```diff
@@ -0,0 +1,39 @@
...
+export default function ExternalLink({ href, innerText }: ExternalLinkProps) {
+  const dispatch = useAppDispatch()
+  const localSettings: LocalSettings = useAppSelector(localSettingsSelector)
+
+  return (
+    <a
+      href={href}
+      target='_blank'
+      rel='noopener nofollow'
+      onClick={e => {
```
2Pacalypse-, line 24: I would do this whole thing a bit differently. Instead of having an inlined event handler here that decides whether to open a dialog or not, I would dispatch an action creator that takes the the click event as an argument, e.g.
```ts
onClick={e => dispatch(maybeOpenExternalLinkDialog(e))}
```
And then that action creator would dispatch a thunk that checks trusted hosts and decides whether to open a dialog or not, e.g.
```ts
function maybeOpenExternalLinkDialog(clickEvent: React.MouseEvent): ThunkAction {
  return (dispatch, getState) => {
    const { settings: { local: { trustedHosts } } } = getState()
    
    // ...rest of code
  }
}
```
You can create this action creator in `action-creators.ts` file in `messaging` folder. This would make it unnecessary to have the selector for local settings in this component at all.

2Pacalypse- (changes requested):

client/messaging/external-link-dialog.tsx
```diff
@@ -0,0 +1,82 @@
+import { shell } from 'electron'
```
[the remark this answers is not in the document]
T1mL3arn: >`electron`

Well, looks like it is the reason why I cannot not open web-version :d

client/messaging/external-link.tsx
```diff
@@ -0,0 +1,39 @@
...
+        // assume here `href` will always be a valid url
```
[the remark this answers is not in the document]
T1mL3arn: This assumption goes from the parent component, which in turn matches the text with regexp and extracts (or at least should extract) a valid url. I guess this is related to #773.

client/messaging/external-link-dialog.tsx
```diff
@@ -0,0 +1,82 @@
...
+import { Dialog } from '../material/dialog'
+import { useAppDispatch, useAppSelector } from '../redux-hooks'
+import { mergeLocalSettings } from '../settings/action-creators'
+import { amberA100 } from '../styles/colors'
+
+const LinkAsText = styled.span.attrs((props: { fontWeight?: string }) => ({
+  // TODO(T1mL3arn) make this component handle very very long links ?
+  fontWeight: props.fontWeight || 'normal',
+}))`
+  color: ${amberA100};
```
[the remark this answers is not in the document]
T1mL3arn: Mixed fonts don't look very good(?)

<details>

image
</details>

client/messaging/external-link-dialog.tsx
```diff
@@ -0,0 +1,82 @@
...
+    <Dialog
+      title='External link'
+      showCloseButton={true}
+      onCancel={onCancel}
+      buttons={buttons}
+      dialogRef={props.dialogRef}>
+      <p>
+        You are going to visit <LinkAsText>{href}</LinkAsText> which is outside of ShieldBattery.
+      </p>
+      <CheckBox
```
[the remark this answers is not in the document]
T1mL3arn: Checkbox label shows hostname, not an entire url. And hostname is always shorter than url (though it is possible that someone sends a link with maximum allowed chars in hostname), Also, current implementation uses hostnames, not domains (where "domain" is protocol + hostname) - right now `http://foo.bar` and `https://foo.bar` are the same. So, need clarifications:

1. You still want the label be changed to simple "trust this host/domain"
2. It is fine to use `hostname` or better change it to `domain` (protocol + hostname) ?

client/messaging/external-link.tsx
T1mL3arn: Is it necessary then to create `client/messaging/actions.ts` and import its types in `client/action-types.ts` ?

client/messaging/external-link-dialog.tsx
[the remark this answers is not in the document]
2Pacalypse-: 1. Yes, I believe there's really no need to specify the actual hostname here.
2. I would change it to show domains. Knowing if the link is HTTP or HTTPS seems important enough to me to include it.

client/messaging/external-link.tsx
2Pacalypse-: Yeah, that is how we usually organize stuff now. Maybe it doesn't make much sense to create a separate file for just one action, but it's very likely we'll add more actions related to messaging in the future, so I'd just create the file now.

client/messaging/external-link-dialog.tsx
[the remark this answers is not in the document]
2Pacalypse-: I think it looks ok for now, and we can worry about fixing the design later, if necessary :d

client/messaging/external-link.tsx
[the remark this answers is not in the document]
2Pacalypse-: Yeah, the parent component supplies the URL now, which is detected by a regex which is far from perfect ^^.

But also, there's no guarantees that we won't use this component somewhere else, where we might get the URL in some other way.

client/messaging/action-creators.ts
```diff
@@ -0,0 +1,21 @@
+import { MouseEvent } from 'react'
+import { openDialog } from '../dialogs/action-creators'
+import { DialogType } from '../dialogs/dialog-type'
+import { ThunkAction } from '../dispatch-registry'
+
+export function maybeOpenExternalLinkDialog(e: MouseEvent<HTMLAnchorElement>): ThunkAction {
+  return (dispatch, getState) => {
+    const trustedDomains: string[] = getState().settings.local.trustedDomains
```
2Pacalypse-, line 8: I believe TypeScript can figure out the type here automatically, so you don't have to explicitly type it.

client/messaging/external-link.tsx
```diff
@@ -0,0 +1,22 @@
+import React from 'react'
+import { useAppDispatch } from '../redux-hooks'
+import { maybeOpenExternalLinkDialog } from './action-creators'
+
+interface ExternalLinkProps {
+  href: string
+  innerText: string
+}
+
+export default function ExternalLink({ href, innerText }: ExternalLinkProps) {
```
2Pacalypse-, line 10: Doesn't seem like this component is necessary anymore. You could attach the `onClick` event handler directly on the anchor element in the `TextMessage`.

Only potential issue is that using `useAppDispatch` hook might break tests without doing some kind of mocking. If that's the case, then ignore this comment and just leave this component as it is.

client/messaging/external-link-dialog.tsx
```diff
@@ -0,0 +1,72 @@
...
+import { RaisedButton, TextButton } from '../material/button'
+import { Dialog } from '../material/dialog'
+import { useAppDispatch, useAppSelector } from '../redux-hooks'
+import { mergeLocalSettings } from '../settings/action-creators'
+import { body2 } from '../styles/typography'
+
+const LinkAsText = styled.span`
+  ${body2};
+  overflow-wrap: anywhere;
+  user-select: text;
```
2Pacalypse-, line 14: It's kind of weird to only have this part as selectable. I'd prefer if the whole text was selectable, or none of it.

I think making the link selectable is a good idea, in case the users want to open it in non-default browser / incognito mode, or whatever.

client/messaging/external-link-dialog.tsx
```diff
@@ -0,0 +1,72 @@
...
+`
+
+interface ExternalLinkDialogProps extends CommonDialogProps {
+  domain: string
+  href: string
+}
+
+export default function ExternalLinkDialog(props: ExternalLinkDialogProps) {
+  const { href, domain, onCancel } = props
+  const trustedDomains: string[] = useAppSelector(s => s.settings.local.trustedDomains)
```
2Pacalypse-, line 24: No need to type this explicitly, I believe.

client/messaging/external-link-dialog.tsx
```diff
@@ -0,0 +1,72 @@
...
+      title='External link'
+      showCloseButton={true}
+      onCancel={onCancel}
+      buttons={buttons}
+      dialogRef={props.dialogRef}>
+      <p>
+        You are going to visit <LinkAsText title={href}>{clampString(href, 256)}</LinkAsText> which
+        is outside of ShieldBattery.
+      </p>
+      <br />
```
2Pacalypse-, line 56: I don't think this new line is necessary.

2Pacalypse- (changes requested):

client/messaging/action-creators.ts
T1mL3arn: Not really, tec27 mentioned it [here](https://github.com/ShieldBattery/ShieldBattery/blob/master/client/ladder/action-creators.ts#L11). Moreover, even if `getState()` is typed, `local` does not show me its fields.

client/messaging/external-link.tsx
T1mL3arn: >If that's the case

It is: `could not find react-redux context value; please ensure the component is wrapped in a <Provider>`. So I leave component then.

client/messaging/external-link-dialog.tsx
T1mL3arn: >I think making the link selectable is a good idea

That is the point - only link is selectable, so no accidental selection of adjacent text. Though it can be improved with click-select-copy behavior. What must be improved - long link handling, current implementation (with simple text truncation) just defeats the purpose of the link being selectable.

2Pacalypse- (changes requested):

T1mL3arn:
What is definitely missed - original issue proposes shieldbattery domains as trusted, but this PR does not have such.

2Pacalypse-:
Yes, that can be done as a separate issue. It's not just a matter of trusting shieldbattery domains automatically, but also clicking on them should redirect the user to a correct location in-client, instead of opening a browser.

PR #774, opened 2021-11-09
Fix the mention regex not matching multiple mentions separated by space.

2Pacalypse-:
When having two mentions separated by a space, the "postfix" character
of the first mention would interfer with the "prefix" of the second
mention, resulting in second mention not being mentioned.

This was fixed by making the "postfix" group use the positive lookahead
instead of a named capture group, so it doesn't end up in matched
string.

Also added a regression test for this behaviour.

### DIFF
--- a/common/text/mentions.ts
+++ b/common/text/mentions.ts
@@ -2,7 +2,7 @@ import { USERNAME_ALLOWED_CHARACTERS } from '../constants'
 import { TypedGroupRegExpMatchArray } from '../regex'
 
 const MENTION_PREFIX = String.raw`(?<prefix>\s|^)`
-const MENTION_POSTFIX = String.raw`(?<postfix>\s|$|[,;:?])`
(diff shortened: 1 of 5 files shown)


common/text/mentions.ts
```diff
@@ -2,17 +2,19 @@ import { USERNAME_ALLOWED_CHARACTERS } from '../constants'
...
  * missing from the allowed punctuation list are . and ! since they are allowed username characters
  * as well.
  *
- * The matched user's name is available in the "user" capture group. There's also two additional
- * capture groups, namely "prefix" and "postfix", that contain all the matched characters
- * before/after the username.
+ * The matched user's name is available in the "user" capture group. There's also one additional
+ * named capture group, namely "prefix", that contains all the matched characters before the
+ * username. For characters that are allowed to come after the username, a positive lookahead group
+ * is used. Those character won't end up in a matched string, which could potentially interfer with
```
tec27, line 16: nit: interfere

tec27 (approved):

PR #775, opened 2021-11-10
Fix using of `Record` function

T1mL3arn:
According the docs, `Record` is a factory function and does not use `new` keyword.
This should fix [this problem](https://github.com/ShieldBattery/ShieldBattery/blob/master/client/ladder/action-creators.ts#L11) and more similar.

With `new` keyword ts/vscode is not capable to type things properly, like `getState().settings` in this [PR review](https://github.com/ShieldBattery/ShieldBattery/pull/772/files/2008ecca84e534d6525e79f028a687ed3d4ef562#diff-46dcd510150c14b97f267e0e45e6d269d4577c133ba50d9af14f6df93f8ff5c6):

<details>
<summary>bad typing</summary>

image

</details>

But with this fix all works:

<details>
<summary>correct typing</summary>

image

</details>

Though if this is a correct fix, it should be fixed everywhere. not only in settings reducer.


tec27:
Seems like a fine change to make, this should only be incorrect in JS files (as the TS compiler errors on this, so we've fixed it in files that have been migrated). You could probably find/replace all `new Record(` with `Record(` in `.js` files safely.

T1mL3arn:
>You could probably find/replace all new Record( with Record( in .js files safely.

```cmd
$ tsc --skipLibCheck false

...

Found 54 errors.
```

tec27:
That doesn't tell me very much without being able to see what the actual errors are. This shouldn't really have an effect on TS compilation because you'd only be running find/replace on JS files, which aren't typechecked by the compiler.

T1mL3arn:
All files are OK excepting `client/lobbies/lobby-reducer.js`. Replacing its `new Record` to `Record` produces those 54 errors.

Some of them are:

```cmd
client/lobbies/lobby.tsx:189:41 - error TS2345: Argument of type 'Record<{ name: null; map: null; gameType: null; gameSubType: null; teams: any; host: null; isCountingDown: boolean; countdownTimer: number; isLoading: boolean; }> & Readonly<{ name: null; map: null; ... 6 more ...; isLoading: boolean; }>' is not assignable to parameter of type 'Lobby'.
  Types of property 'has' are incompatible.
    Type '(key: string) => key is "map" | "name" | "host" | "gameType" | "gameSubType" | "teams" | "isCountingDown" | "countdownTimer" | "isLoading"' is not assignable to type '(key: string) => key is "map" | "name" | "host" | "gameType" | "gameSubType" | "teams"'.
      Type predicate 'key is "map" | "name" | "host" | "gameType" | "gameSubType" | "teams" | "isCountingDown" | "countdownTimer" | "isLoading"' is not assignable to 'key is "map" | "name" | "host" | "gameType" | "gameSubType" | "teams"'.
        Type '"map" | "name" | "host" | "gameType" | "gameSubType" | "teams" | "isCountingDown" | "countdownTimer"
| "isLoading"' is not assignable to type '"map" | "name" | "host" | "gameType" | "gameSubType" | "teams"'.
          Type '"isCountingDown"' is not assignable to type '"map" | "name" | "host" | "gameType" | "gameSubType" | "teams"'.

189     const [, , mySlot] = findSlotByName(lobby, user.name)
                                            ~~~~~

client/lobbies/lobby.tsx:190:30 - error TS2531: Object is possibly 'null'.

190     const isHost = mySlot && lobby.host.id === mySlot.id
                                 ~~~~~~~~~~

client/lobbies/lobby.tsx:191:44 - error TS2345: Argument of type 'Record<{ name: null; map: null; gameType: null; gameSubType: null; teams: any; host: null; isCountingDown: boolean; countdownTimer: number; isLoading: boolean; }> & Readonly<{ name: null; map: null; ... 6 more ...; isLoading: boolean; }>' is not assignable to parameter of type 'Lobby'.

191     const canAddObsSlots = canAddObservers(lobby)
                                               ~~~~~
```

So for now all but `lobby-reducer` are fixed.

tec27:
I fixed all the issues in 22277f1e99f6355c1a12fe7f11a6568f88d48bbd

PR #779, opened 2021-11-12
Make channel owners be defined explicitly, instead of implicitly.

2Pacalypse-:
Currently, we were using the "edit_permissions" flag as a makeshift
indicator of channel ownership. However, this is not ideal since more
than one person can have this permission set. Technically we could've
defined a channel owner at the model level, where the owner would be set
as a user with this permission and who joined the channel the earliest.

However, I feel like this would get a bit unwieldy and messy down the
line, especially if we start thinking about features like transferring
channel ownership.

This PR adds another column to the database, which ensures that there
can be only one owner per channel, at the database-level. I guess it is
a little bit weird to have a flag that trumps all other flags, no matter
their values, but I believe this is what Discord does too, so we should
be fine? :d

### DIFF
--- a/server/migrations/20211110231126-add-channel-owners.js
+++ b/server/migrations/20211110231126-add-channel-owners.js
@@ -1,0 +1,26 @@
+exports.up = async function (db) {
+  await db.runSql(`
+    ALTER TABLE joined_channels
+    ADD COLUMN owner
+    BOOLEAN NOT NULL DEFAULT false;
+  `)
+
+  await db.runSql(`
+    CREATE UNIQUE INDEX joined_channels_owner
+    ON joined_channels (channel_name, owner)
+    WHERE owner = true;
+  `)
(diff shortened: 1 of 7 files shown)


server/migrations/20211110231126-add-channel-owners.js
```diff
@@ -0,0 +1,26 @@
+exports.up = async function (db) {
+  await db.runSql(`
```
tec27, line 2: This migration does not seem complete enough, it needs to set proper values for channels that already exist. I realize we haven't shipped multiple channels in production yet, but we do have them locally, and getting our local channels into some state that won't exist on production is going to lead to a bunch of weird errors and misaligned expectations. You can probably use a similar query to the user leaving ones to find the right owner

tec27 (changes requested):

PR #785, opened 2021-11-18
Send user their channel permissions when they join a channel.

2Pacalypse-:
This PR also does a couple of other things:
- cleans up most of the chat event types to only use `userId` instead
  of a whole channel user object if possible
- renames the `joined_channel` table in the DB to `channel_users`

### DIFF
--- a/client/chat/actions.ts
+++ b/client/chat/actions.ts
@@ -1,5 +1,4 @@
 import {
-  ChatEvent,
   ChatInitEvent,
   ChatJoinEvent,
   ChatLeaveEvent,
--- a/client/chat/actions.ts
+++ b/client/chat/actions.ts
@@ -198,18 +198,12 @@ export interface DeactivateChannel {
 }
 
 /**
- * A type that allows using the socket events directly as payload for dispatch actions, by including
- * the channel in the payload, since the channel is not included in the event itself (it is part of
- * the websocket URL).
- */
-type PayloadWithChannel<T extends ChatEvent> = { channel: string } & Omit<T, 'action'>
-
-/**
  * We have joined a channel and the server has sent us some initial data.
  */
 export interface InitChannel {
   type: '@chat/initChannel'
-  payload: PayloadWithChannel<ChatInitEvent>
+  payload: ChatInitEvent
+  meta: { channel: string }
 }
 
 /**
@@ -217,7 +211,8 @@ export interface InitChannel {
  */
 export interface UpdateJoin {
   type: '@chat/updateJoin'
-  payload: PayloadWithChannel<ChatJoinEvent>
+  payload: ChatJoinEvent
+  meta: { channel: string }
 }
 
 /**
@@ -225,7 +220,8 @@ export interface UpdateJoin {
  */
 export interface UpdateLeave {
   type: '@chat/updateLeave'
-  payload: PayloadWithChannel<ChatLeaveEvent>
+  payload: ChatLeaveEvent
+  meta: { channel: string }
 }
 
 /**
@@ -233,7 +229,7 @@ export interface UpdateLeave {
  */
 export interface UpdateLeaveSelf {
   type: '@chat/updateLeaveSelf'
-  payload: { channel: string }
+  meta: { channel: string }
 }
 
 /**
@@ -241,7 +237,8 @@ export interface UpdateLeaveSelf {
  */
 export interface UpdateMessage {
   type: '@chat/updateMessage'
-  payload: PayloadWithChannel<ChatMessageEvent>
+  payload: ChatMessageEvent
+  meta: { channel: string }
 }
 
 /**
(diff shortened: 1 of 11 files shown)


common/chat.ts
```diff
@@ -15,10 +15,20 @@ export enum ClientChatMessageType {
...
+  changeTopic: boolean
+  togglePrivate: boolean
+  editPermissions: boolean
+  owner: boolean
+}
+
 export interface ChatUser {
   id: SbUserId
   name: string
+  joinDate: number
```
tec27, line 30: I'm not a fan of adding these to this structure. If we want to include chat-specific information then the `name` should be removed and this structure should only contain a userId and these chat-specific things, and the structure with id + name should be sent separately from this structure

common/chat.ts
```diff
@@ -15,10 +15,20 @@ export enum ClientChatMessageType {
...
+  permissions: ChannelPermissions
```
tec27, line 31: I do not think it makes sense to send channel permissions about every user to every other user in chat, that seems like a massive information leak to me. I understand you may need this information for some UIs, but those UIs should retrieve it as they need it (and it should be limited to users that should have permission to do so)

tec27 (changes requested):

client/chat/actions.ts
```diff
@@ -193,89 +196,73 @@ export interface DeactivateChannel {
   }
 }
 
+/**
+ * A type that allows using the socket events directly as payload for dispatch actions, by including
+ * the channel in the payload, since the channel is not included in the event itself (it is part of
+ * the websocket URL).
+ */
+type PayloadWithChannel<T = void> = { channel: string } & Omit<T, 'action'>
```
tec27, line 204: This is a very weird type imo. If you're going to omit things from a generic type, it should be limited to types that actually contain the key you're omitting

I would also recommend not using default values for generics unless they are actually useful. The default type here does not seem useful/good, makes it easy to misuse the type, and prevents better type inference in places that might be utilizing this type

server/lib/chat/chat-service.ts
```diff
@@ -515,10 +516,7 @@ export default class ChatService {
     for (const c of channels.values()) {
       this.publisher.publish(getChannelPath(c), {
         action: 'userOffline',
-        user: {
-          id: userId,
-          name: user.name,
-        },
+        userId,
```
tec27, line 519: You're changing the structure of this event, so you need to pass it on a separate path (`action` in this case) or you're going to break existing clients

tec27: Apply throughout this file, there's a bunch of these cases

tec27 (changes requested):

server/lib/chat/chat-models.ts
```diff
@@ -1,77 +1,21 @@
...
-  changeTopic: boolean
-  togglePrivate: boolean
-  editPermissions: boolean
-  owner: boolean
-}
-
-export interface JoinedChannel {
+export interface UserChannelEntry {
   userId: SbUserId
   userName: string
```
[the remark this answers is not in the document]
2Pacalypse-: Currently it's only used when adding a new user to the channel. Maybe we can adjust the return type of that function to something like `UserChannelEntry & { userName: string }`, but not sure if that's better.

server/lib/chat/chat-models.ts
```diff
@@ -88,19 +32,58 @@ function convertJoinedChannelFromDb(props: DbJoinedChannel): JoinedChannel {
...
+    return result.rows
+  } finally {
+    done()
+  }
+}
+
+/**
+ * Gets a user channel entry for a particular user in a particular channel.
+ */
+export async function getChannelUserEntryForUser(
```
tec27, line 75: Is this meant to be `getUserChannelEntryForUser`? As is it's a weird rearrangement from the actual type

server/lib/chat/chat-models.ts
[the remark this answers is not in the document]
tec27: I can see from looking at the code that it's just constructing an `SbUser` from the stuff returned here, you should just stop doing that and request the user out of the DB separately like we do elsewhere. It will be a lot more resilient to us adding things to the user structure that need to be sent down as well

server/lib/chat/chat-service.ts
```diff
@@ -127,8 +132,8 @@ export default class ChatService {
       this.publisher.publish(getChannelPath(originalChannelName), {
         action: 'join2',
         user: {
-          id: result.userId,
-          name: result.userName,
+          id: userInfo.id,
```
tec27, line 135: nit: just return the result from `findUserById` directly, so we don't have to update this every time that struct changes

server/lib/chat/chat-service.ts
```diff
@@ -92,6 +93,10 @@ export default class ChatService {
       throw new ChatServiceError(ChatServiceErrorCode.InvalidJoinAction, 'Already in this channel')
     }
 
+    const userInfo = await findUserById(userId)
```
tec27, line 96: I think as a slight optimization you could await this at the same time as `isUserBannedFromChannel` below, e.g.
`const [userInfo, isBanned] = await Promise.all([
  findUserById(userId),
  isUserBannedFromChannel(originalChannelName, userId),
])`

not really a huge deal but reduces a round trip to the DB

tec27 (approved):

PR #788, opened 2021-12-02
Split user profile overlay into two that open with left and right click.

2Pacalypse-:


2Pacalypse-:
Depends on #785

client/material/portal.tsx
```diff
@@ -52,19 +52,35 @@ export function Portal(props: PortalProps) {
...
+      }
+    },
+    [onDismiss, portalRef],
+  )
+
   useEffect(() => {
     if (open) {
       document.addEventListener('click', onCaptureClick, true /* useCapture */)
       document.addEventListener('click', onBubbleClick)
+      document.addEventListener('contextmenu', onContextMenu, true /* useCapture */)
```
tec27, line 73: Unsure why you don't think we need a bubbling event handler here? This stuff is meant to handle cases where e.g. we have an overlay in a dialog, and they click outside of the dialog to dismiss it. Without that handling, we're gonna dismiss the dialog as well, so please handle it in the same way like we talked about

client/profile/user-profile-overlay.tsx
```diff
@@ -40,70 +35,15 @@ export function ConnectedUserProfileOverlay({
...
-          <MenuItem key='invite' text='Invite to party' onClick={onInviteToPartyClick} />,
-        )
-      }
-    }
-  }
-
   return (
     <UserProfileOverlay user={user} popoverProps={popoverProps}>
-      {actions}
+      {null}
```
tec27, line 46: I don't think actions make sense ever again as of this PR, so you should probably just change the expected props of this component. Are we sharing this with the bottom left thing or something? If so they should probably be split, they aren't the same

client/material/portal.tsx
2Pacalypse-: My belief was that the `markEventAsHandledDismissal(event)` line was in charge of that logic, i.e. not dismissing the dialog if there's another overlay in it. And as for the capture/bubble handlers combo, I thought that was done so when you click on user's entry in channel list, it opens their overlay, and then when you click on their entry again, it should close it, which it does. And this is specifically what we don't want to happen when right-clicking, or at least I thought so.

For what it's worth, I did test pretty much all scenarios I could think of and I thought this is behaviour we'd want (e.g. it doesn't dismiss the dialog when they click outside it with another overlay open). The only thing that I noticed that is perhaps a bit weird is when you right-click on same user entry twice, there is no open/close animation the second time since it closes and opens the same overlay too fast I guess? But it seemed like a minor thing to me.

PR #789, opened 2021-12-07
Allow SB client to be registered for *.rep files

### LINKED ISSUE #769: Allow SB client to be registered for *.rep files
This would register SB in Windows as the default replay file handler (optionally, perhaps we could provide a button to do it in settings and offer to do so on first launch?). When a replay file is opened with SB, it should go through the normal replay launch process (as if it had been selected from the replay overlay).

T1mL3arn:
closes #769

1. launch shb and start a replay when a user opens the replay file
2. start a replay when a user opens the replay file if shb is already launched
3. start a replay if the replay file was dropped into shb window (works only in desktop version)
4. register file association for `.rep` files so they can be opened with shb app with double click

File association registers during installation without asking a user. Also no settings in shb app to change it (my guess it will be needed to play with windows registry to do so).

About testing: pack entire prod app and associate `.rep` files manually with compiled `.exe`. The building takes noticeable amount of time (at least 1.5 minutes for me). Maybe there is an easier way to build and test it. Though at this time it probably does not matter since core functional is here.

### DIFF
--- a/app/single-instance.js
+++ b/app/single-instance.js
@@ -23,8 +23,19 @@ export default function () {
       // successfully connect to the running server). Just send some data to the server, which is
       // running on the first instance, so the main window can be focused there and quit this
       // instance.
-      client.write('focus first instance')
-      app.quit()
+      const event = {
+        type: 'focus first instance',
+      }
+      if (process.argv.length > 1) {
+        const replayFile = process.argv[1]
+
+        if (replayFile.endsWith('.rep')) {
+          event.type = 'openReplay'
+          event.replay = replayFile
+        }
+      }
+
+      client.write(JSON.stringify(event), () => app.quit())
     })
     .on('error', err => {
       if (err.code !== 'ENOENT') throw err
@@ -43,9 +54,16 @@ export default function () {
       getMainWindow = require('./app').getMainWindow
       net
         .createServer(connection => {
-          connection.on('data', () => {
+          connection.on('data', data => {
             const mainWindow = getMainWindow()
             if (mainWindow) {
+              try {
+                const event = JSON.parse(data)
+                if (event.type === 'openReplay') {
+                  mainWindow.webContents.send('openReplay', event.replay)
+                }
+              } catch (e) {}
+
               if (!mainWindow.isVisible()) {
                 mainWindow.show()
                 return
--- a/app/single-instance.js
+++ b/app/single-instance.js
@@ -23,19 +23,15 @@ export default function () {
       // successfully connect to the running server). Just send some data to the server, which is
       // running on the first instance, so the main window can be focused there and quit this
       // instance.
-      const event = {
-        type: 'focus first instance',
-      }
+
+      let event = 'focus first instance'
+
       if (process.argv.length > 1) {
-        const replayFile = process.argv[1]
+        const file = process.argv[1]
 
-        if (replayFile.endsWith('.rep')) {
-          event.type = 'openReplay'
-          event.replay = replayFile
-        }
+        if (file.endsWith('.rep')) event = file
       }
-
-      client.write(JSON.stringify(event), () => app.quit())
+      client.write(event, () => app.quit())
     })
     .on('error', err => {
       if (err.code !== 'ENOENT') throw err
@@ -57,12 +53,10 @@ export default function () {
           connection.on('data', data => {
             const mainWindow = getMainWindow()
             if (mainWindow) {
-              try {
-                const event = JSON.parse(data)
-                if (event.type === 'openReplay') {
-                  mainWindow.webContents.send('openReplay', event.replay)
-                }
-              } catch (e) {}
+              const file = data.toString()
+              if (file.endsWith('.rep')) {
+                mainWindow.webContents.send('openReplay', file)
+              }
 
               if (!mainWindow.isVisible()) {
                 mainWindow.show()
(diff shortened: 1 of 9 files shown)


tec27:
Sorry for the delay on reviewing this, we're in the middle of trying to get a bunch of things done for our next release. I'll get to it as soon as I can

app/single-instance.js
```diff
@@ -23,8 +23,15 @@ export default function () {
       // successfully connect to the running server). Just send some data to the server, which is
       // running on the first instance, so the main window can be focused there and quit this
       // instance.
-      client.write('focus first instance')
-      app.quit()
+
+      let event = 'focus first instance'
+
+      if (process.argv.length > 1) {
```
tec27, line 29: These kind of checks should ideally be shared code with what is happening in the main process, as already even in this PR they are subtly different and have different bugs.

I think a better overall approach here would be to serialize any arguments here and pass them to the main process directly (by sending them in the `client.write`), then the main process can handle them via the same code it handles the initial launch arguments with.

app/single-instance.js
```diff
@@ -43,9 +50,14 @@ export default function () {
       getMainWindow = require('./app').getMainWindow
       net
         .createServer(connection => {
-          connection.on('data', () => {
+          connection.on('data', data => {
             const mainWindow = getMainWindow()
             if (mainWindow) {
+              const file = data.toString()
+              if (file.endsWith('.rep')) {
+                mainWindow.webContents.send('openReplay', file)
```
tec27, line 58: This, too, demonstrates the risk of doing this in a bunch of different places (here you aren't using the type-safe IPC stuff, but in the other place you are), definitely should be shared code

client/replays/action-creators.ts
```diff
@@ -15,6 +17,23 @@ import { makeServerUrl } from '../network/server-url'
 
 const ipcRenderer = new TypedIpcRenderer()
 
+class ReplayFile {
```
tec27, line 20: This really doesn't feel like it should be a class (and the code that follows and uses it is not technically correct, since the JS code that uses this elsewhere is *not* passing instances of this class). What you want is a simple interface:

```ts
interface ReplayFile {
  path: string
  name: string
}
```

No real value from the `fromPath` function here either since we only do that in a single place, you can just inline that implementation there

client/replays/replay-drop.tsx
```diff
@@ -0,0 +1,44 @@
...
+import { useAppDispatch } from '../redux-hooks'
+import { openReplay } from './action-creators'
+
+export default function ReplayDrop() {
+  const dispatch = useAppDispatch()
+
+  const onDrop = useCallback(
+    (e: DragEvent) => {
+      e.preventDefault()
+      e.stopPropagation()
```
tec27, line 11: Calling `stopPropagation` is an anti-pattern and I'd avoid using it unless you *really* need to.

client/replays/replay-drop.tsx
```diff
@@ -0,0 +1,44 @@
...
+      }
+    },
+    [dispatch],
+  )
+
+  const onDragOver = useCallback((e: DragEvent) => {
+    e.preventDefault()
+    e.stopPropagation()
+
+    // NOTE: it is not possible to access files during `dragover` now,
```
tec27, line 25: I'm not sure that this is really worth noting, people in that issue seem to me to be misunderstanding the spec and I very much doubt this behavior will ever change back (it is most certainly a security issue if Chromium allows any random page to read files dragged over them but not actually dropped)

client/replays/replay-drop.tsx
```diff
@@ -0,0 +1,44 @@
...
+export default function ReplayDrop() {
+  const dispatch = useAppDispatch()
+
+  const onDrop = useCallback(
+    (e: DragEvent) => {
+      e.preventDefault()
+      e.stopPropagation()
+
+      const [file] = e.dataTransfer?.files || []
+      if (file?.path.endsWith('.rep')) {
```
tec27, line 14: Same as during application launch, this needs to handle extensions case-insensitively

client/replays/replay-drop.tsx
```diff
@@ -0,0 +1,44 @@
...
+    // NOTE: it is not possible to access files during `dragover` now,
+    // see this electron issue:
+    // https://github.com/electron/electron/issues/9840
+    // Thus, just allow any files here and deal with the garbage
+    // in `drop` handler
+    if (e.dataTransfer) e.dataTransfer.dropEffect = 'link'
+  }, [])
+
+  useEffect(() => {
+    document.addEventListener('drop', onDrop)
```
tec27, line 34: I don't really like that these events are being attached to the document, as I don't think context-less handling for replays is very likely to last that long. We definitely want to be able to e.g. drag map files in to upload and play on, probably drag replay files into the party interface to watch replays with others, etc. To do all that, we're much better served by contextual event handling, rather than one handler at the root.

A better approach that will serve us long-term here would be to make a component that fills its container completely, and attach the drag handlers to that. This would also give us a place to show a UI as you're dragging, which will actually make this discoverable.

prod.yml
```diff
@@ -21,6 +21,11 @@ files:
     filter:
       - '**/*.js'
       - '**/*.node'
+fileAssociations:
+  ext: 'rep'
+  name: 'Starcraft Broodwar Replay'
```
tec27, line 26: 'Broodwar' is not a thing, the trademark is 'StarCraft: Brood War'. I would personally just simplify both the name and description here. From Electron's docs, it seems like the name should be a lot shorter, so probably just:

```yaml
name: 'REP'
description: 'StarCraft Replay File'
```

tec27 (changes requested):

tec27:
This work has been superseded by f02152efe5218a7b221091c98a31668aabcd821f

PR #800, opened 2021-12-25
Add the client-side code for kicking and banning users from chat channels.

2Pacalypse-:
Currently only global moderators and chat owners (person who created the
chat channel) can kick/ban users, since there's no way in the UI yet to
change the channel permissions of other users. That should come soon.


server/lib/chat/chat-api.ts
```diff
@@ -190,20 +198,38 @@ export class ChatApi {
     return await this.chatService.getChannelUsers(channelName, ctx.session!.userId)
   }
 
-  @httpDelete('/:channelName/:targetId')
+  @httpGet('/:channelName/:targetId')
```
tec27, line 201: You should be *really* careful when you add routes like this, as they can collide with previous routes. Yes, using user IDs here minimizes risk, but it's still a risk and I don't think we should be taking it. On top of that, this just isn't really very RESTful. This should be `/:channelName/users/:targetId` (as should all other routes that deal with specific users)

client/chat/channel-ban-user-dialog.tsx
```diff
@@ -0,0 +1,91 @@
...
+import TextField from '../material/text-field'
+import { useAppDispatch } from '../redux-hooks'
+import { openSnackbar } from '../snackbars/action-creators'
+import { moderateUser } from './action-creators'
+
+interface BanUserModel {
+  banReason: string
+}
+
+type ChannelBanUserDialogProps = CommonDialogProps & {
```
tec27, line 18: use an interface and extend, the types will show up a lot better

client/chat/channel-ban-user-dialog.tsx
```diff
@@ -0,0 +1,91 @@
...
+export function ChannelBanUserDialog({
+  dialogRef,
+  onCancel,
+  channel,
+  user,
+}: ChannelBanUserDialogProps) {
+  const dispatch = useAppDispatch()
+  const inputRef = useRef<TextField>(null)
+
+  useEffect(() => {
```
tec27, line 32: Instead of copy-pasting this another time, pull this into a hook that returns a ref to place on the thing that should be focused

client/chat/socket-handlers.ts
```diff
@@ -99,22 +140,39 @@ const eventToAction: EventToActionMap = {
...
   siteSocket.registerRoute('/chat2/:channel', (route, event) => {
     const actionName = event.action as ChatEvent['action']
-    if (!eventToAction[actionName]) return
+    if (!eventToChatAction[actionName]) return
+
+    const action = eventToChatAction[actionName](route.params.channel, event)
+    if (action) dispatch(action)
+  })
+
+  siteSocket.registerRoute('/chat/:channel/:userId', (route, event) => {
```
tec27, line 171: These should also be dispatched on `/chat/:channel/users/:userId` imo

common/chat.ts
```diff
@@ -199,3 +241,32 @@ export interface ModerateChannelUserServerRequest {
    */
   moderationReason?: string
 }
+
+/**
+ * Specific chat information about a user in a particular channel, such as their join date, their
+ * channel role (e.g. are they a moderator), etc.
+ */
+export interface ChatUserProfile {
```
tec27, line 249: imo these structures should usually always include the relevant ID as well (in this case, probably both the user ID and the channel name)

common/chat.ts
```diff
@@ -199,3 +241,32 @@ export interface ModerateChannelUserServerRequest {
...
+   * User is considered a channel moderator if they have one of the following permissions:
+   *  - owner
+   *  - editPermissions
+   *  - ban
+   *  - kick
+   */
+  isModerator: boolean
+}
+
+export type ChatUserProfileJson = Jsonify<ChatUserProfile>
```
tec27, line 261: If you're going to make a Jsonify'd structure, you should basically always *also* be writing the the `toFooJson` method as you do so. Don't make the user of this type write the conversion every time, it's extremely error-prone

common/chat.ts
```diff
@@ -199,3 +241,32 @@ export interface ModerateChannelUserServerRequest {
...
+
+/**
+ * The response returned when fetching a profile of a user in a specific chat channel. The profile
+ * itself can be `null` in case the user has left the channel, but their name is still visible in
+ * old messages.
+ */
+export interface GetChatUserProfileResponse {
+  userId: SbUserId
+  channelName: string
+  profile: ChatUserProfileJson | null
```
tec27, line 271: In general never use `null` (unless it is part of some existing API we don't control). Vastly, vastly, vastly prefer `undefined` (and optional fields when relevant)

server/lib/chat/chat-service.ts
```diff
@@ -217,27 +206,40 @@ export default class ChatService {
         "Can't moderate yourself",
       )
     }
+
+    const { channelPermissions: moderatorChannelPermissions } = moderatorUserChannel
+    const { channelPermissions: targetChannelPermissions } = targetUserChannel
+
+    // TODO(2Pac): Really need tests for these.
+
+    if (targetChannelPermissions.owner && !moderatorPermissions?.moderateChatChannels) {
```
tec27, line 215: imo you should really name these associated conditions by putting them in a variable, and then using those variables in these if statements. These are *absurdly* hard to read and comprehend what the underlying intention is meant to be (and sure, tests would help ensure they don't break in the future, but it would be better to also write code we can actually understand)

server/lib/chat/chat-service.ts
```diff
@@ -217,27 +206,40 @@ export default class ChatService {
...
+      throw new ChatServiceError(
+        ChatServiceErrorCode.ModeratorAccess,
+        'Only global moderators can moderate channel owners',
```
tec27, line 218: I don't think `global moderator` is a good term to standardize on, it seems to imply something in a geographic sense. `Server moderator` or `Service moderator` or `ShieldBattery staff` would be better imo

server/lib/chat/chat-service.ts
```diff
@@ -217,27 +206,40 @@ export default class ChatService {
...
-        targetUserChannel.channelPermissions.ban ||
-        targetUserChannel.channelPermissions.kick
-      ) {
-        throw new ChatServiceError(
-          ChatServiceErrorCode.ModeratorAccess,
-          "Can't moderate users that have moderator access",
-        )
-      }
+      throw new ChatServiceError(
+        ChatServiceErrorCode.ModeratorAccess,
```
tec27, line 229: so here in this function you are returning the same error code for *3* different errors. We *never ever ever* want to do that, error codes are here to help us differentiate what happened. If these are the same error, check them in the same statement and return the same message as well. If these are different errors, give them different codes.

server/lib/chat/chat-service.ts
```diff
@@ -406,6 +400,43 @@ export default class ChatService {
...
 
+  async getChatUserProfile(channelName: string, userId: SbUserId, targetId: SbUserId) {
+    const userSockets = this.getUserSockets(userId)
+    const originalChannelName = await this.getOriginalChannelName(channelName)
+    if (
+      !this.state.users.has(userSockets.userId) ||
+      !this.state.users.get(userSockets.userId)!.has(originalChannelName)
+    ) {
+      throw new ChatServiceError(
+        ChatServiceErrorCode.InvalidGetUsersAction,
```
tec27, line 411: I find this error code naming to be pretty bad, it's basically "generic error for route X". This particular usage would be better named "NotInChannel" or something

server/lib/chat/chat-service.ts
```diff
@@ -443,14 +475,31 @@ export default class ChatService {
     channelName: string,
     userId: SbUserId,
   ): Promise<SbUserId | null> {
-    const { newOwnerId } = await leaveChannel(userId, channelName)
+    const { newOwnerId } = await removeUserFromChannel(userId, channelName)
+
+    // Notify the new owner that their permissions changed, but only if they're online.
```
tec27, line 480: I don't really see the point in checking if they're online, we have a consistent route structure to make it easy to dispatch messages to particular users, just dispatch the message to the route. If they're not online, no sockets will get the message, done. The whole nice part about nydus over e.g. socket.io is this idea of persistent route subscriptions, if you go around checking if sockets are registered every time it sort of defeats the purpose

tec27 (changes requested):

client/chat/channel-ban-user-dialog.tsx
tec27: I needed this for something else, so this is now in `material/auto-focus.ts` and you can just use it here too

server/lib/chat/chat-service.ts
2Pacalypse-: I guess that comment is misleading; my idea with checking if they're online was just as a slight optimization to avoid making a database query fetching their permissions if they're not. If you think the optimization is not worth it, I can just remove it.

client/chat/action-creators.ts
```diff
@@ -15,7 +26,28 @@ export function joinChannel(channel: string): ThunkAction {
     })
     dispatch({
       type: '@chat/joinChannel',
-      payload: fetchJson<void>(apiUrl`chat/${channel}`, { method: 'POST' }),
+      payload: fetchJson<void>(apiUrl`chat/${channel}`, { method: 'POST' }).catch(err => {
```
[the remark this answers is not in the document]
2Pacalypse-: I went with `abortableThunk` approach, but I still wanted to handle the errors in the action-creator, instead of the component itself. The reason for this is that I assume we'll eventually be dispatching this action from multiple places and we'd need to handle these errors in all of them otherwise.

client/chat/action-creators.ts
```diff
@@ -15,7 +26,28 @@ export function joinChannel(channel: string): ThunkAction {
...
+        // TODO(2Pac): Rework how joining channel works. Currently we first navigate to the channel
+        // and then attempt to join it. Which seems weird to me?
+        //
+        // To work around this for now, if the user is banned we redirect them to the index page,
+        // but ideally they wouldn't be navigated to the channel in the first place.
+        replace('/')
```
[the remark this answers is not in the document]
2Pacalypse-: I was thinking of doing this in a separate PR, since I also need to handle the case of someone landing in a channel by directly entering the URL of it.

client/chat/action-creators.ts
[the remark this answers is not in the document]
tec27: I would very much recommend against that, error handling is almost *always* specific to the UI in which the error was generated. If you're trying to join a channel through a dialog, it might make sense to show errors in the dialog. If you're trying to join by clicking a link in another channel, we probably want to show a snackbar or bring up a dialog, etc.

I handled some errors in action creators for matchmaking because it was very hard to avoid at the time, but in practice it's sort of terrible and has led to other problems with hot reloading, etc. because of action creators depending on TSX files.

client/chat/action-creators.ts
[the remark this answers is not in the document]
tec27: Yeah, that's fine

server/lib/chat/chat-service.ts
tec27: That optimization is not worth it

tec27: If the query above is already returning an owner ID, it could likely just return the new permissions as well, if you really want to optimize that. (But generally, this is a query for 1 row from a single table using only the primary key of the table, it's not really every going to be an expensive thing to do)

client/profile/user-context-menu.tsx
```diff
@@ -1,34 +1,79 @@
...
+  const selfUser = useSelfUser()
+  const cancelLoadRef = useRef(new AbortController())
+  const [isLoadingChatUserProfile, setIsLoadingChatUserProfile] = useState<boolean>(false)
+  const [loadingChatUserProfileError, setLoadingChatUserProfileError] = useState<Error>()
+  const selfPermissions = useAppSelector(s => s.auth.permissions)
   const user = useAppSelector(s => s.users.byId.get(userId))
+  const [, routeParams] = useRoute('/chat/:channelName')
+  const channelName = routeParams?.channelName.toLowerCase()
+  const channel = useAppSelector(s => (channelName ? s.chat.byName.get(channelName) : undefined))
+  const chatUserProfile = channel?.userProfiles.get(userId)
```
tec27, line 48: This is a bad way to get this data because now this component needs to re-render any time there is a new message in the channel, someone joins or leaves, someone goes online or offline, etc. The ideal way to return this is inside the `useAppSelector` call, as a single thing. I don't think we need the channel object at all so this should be fairly straightforward

client/profile/user-context-menu.tsx
```diff
@@ -1,34 +1,79 @@
...
 export function ConnectedUserContextMenu({ userId, popoverProps }: ConnectedUserContextMenuProps) {
   const dispatch = useAppDispatch()
-  const selfUser = useAppSelector(s => s.auth.user)
+  const selfUser = useSelfUser()
+  const cancelLoadRef = useRef(new AbortController())
+  const [isLoadingChatUserProfile, setIsLoadingChatUserProfile] = useState<boolean>(false)
+  const [loadingChatUserProfileError, setLoadingChatUserProfileError] = useState<Error>()
+  const selfPermissions = useAppSelector(s => s.auth.permissions)
   const user = useAppSelector(s => s.users.byId.get(userId))
+  const [, routeParams] = useRoute('/chat/:channelName')
```
tec27, line 45: if we're using this in a bunch of places, it should be declared as a hook somewhere. This is gonna be hard to track down if we e.g. change the chat paths at some point in the future

tec27: This also, imo, doesn't really make much sense here. This component is not only used inside chat channels. The intended path for special menu items that depend on context was more like the handling in my shift-click change. I didn't yet build the similar stuff for menu items, but you should do that here instead of checking the route at all.

client/profile/user-context-menu.tsx
tec27: After commenting above, yes this is also not good to be doing in this component, given that it is a general component and not at all specific to chat.

tec27 (changes requested):

client/chat/channel-ban-user-dialog.tsx
```diff
@@ -0,0 +1,87 @@
...
+
+  const onFormSubmit = useCallback(
+    (model: BanUserModel) => {
+      dispatch(
+        moderateUser(
+          channel,
+          user.id,
+          ChannelModerationAction.Ban,
+          {
+            onSuccess: () => dispatch(openSnackbar({ message: `${user.name} was banned` })),
```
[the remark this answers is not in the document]
2Pacalypse-: Hmm, do we care about displaying UI for errors that _shouldn't_ be possible to trigger through UI?

client/profile/user-context-menu.tsx
2Pacalypse-: I'm not sure I quite understand how to do that. I took a look at the commit where you added the shift-click change, but I only ended up even more confused :p

tec27: You need a way of passing in extra menu items, or modifying the list of menu items, or something along those lines. You can do it as props directly, you can use Context for it (as in the shift-click stuff), etc. There's a lot of options. What you *can't* do is what you've done here: add super-specific logic to a very general component

client/network/abortable-thunk.ts
```diff
@@ -17,12 +17,12 @@ export interface RequestHandlingSpec<T = void> {
   /**
    * A function that will be called if the underlying request succeeds.
    */
-  onSuccess: (result: T) => void
+  onSuccess?: (result: T) => void
```
tec27, line 20: I didn't make these optional because it makes it very easy to use this API wrong, would prefer you don't change that

common/chat.ts
```diff
@@ -2,17 +2,17 @@ import { Jsonify } from './json'
...
 export enum ChatServiceErrorCode {
-  UserOffline = 'UserOffline',
-  UserNotFound = 'UserNotFound',
-  InvalidJoinAction = 'InvalidJoinAction',
+  AlreadyJoined = 'AlreadyJoined',
   LeaveShieldBattery = 'LeaveShieldBattery',
-  InvalidLeaveAction = 'InvalidLeaveAction',
-  InvalidModerationAction = 'InvalidModerationAction',
-  ModeratorAccess = 'ModeratorAccess',
+  ModerateChannelOwner = 'ModerateChannelOwner',
```
tec27, line 7: I think these are better when they're phrased as the error, not as what you were trying to do. "CannotModerateYourself" is better than "ModerateYourself", etc.

common/chat.ts
```diff
@@ -260,13 +262,22 @@ export interface ChatUserProfile {
...
+    joinDate: Number(chatUserProfile.joinDate),
+    isModerator: chatUserProfile.isModerator,
+  }
+}
+
 /**
  * The response returned when fetching a profile of a user in a specific chat channel. The profile
- * itself can be `null` in case the user has left the channel, but their name is still visible in
- * old messages.
+ * itself can be `undefined` in case the user has left the channel, but their name is still visible
```
tec27, line 276: nit: If you have a comment about a particular field, it should probably just be a doc comment on that field

server/lib/chat/chat-service.ts
```diff
@@ -477,7 +476,9 @@ export default class ChatService {
   ): Promise<SbUserId | null> {
     const { newOwnerId } = await removeUserFromChannel(userId, channelName)
 
-    // Notify the new owner that their permissions changed, but only if they're online.
+    // As a slight optimization, attempt to find the new owner in the database and notify them that
```
tec27, line 479: So you've added a comment stating the why, but I still don't think this is worth doing, so that doesn't really address my concern. If we wanted to be totally maximally efficient on everything we could do little things like this and add comments everywhere, but the app would become extremely brittle for no real gain. The options here are:

- Utilize the data from a previous query (probably updating the query to return them in addition to what it returns now)
- Just make the new query always (it is *extremely* cheap)

With either of those options, you're always notifying the new owner and letting Nydus deal with no-oping the send if they're offline, which is the right division of responsibilities here.

tec27 (changes requested):

server/lib/chat/chat-service.ts
2Pacalypse-: Oh, I already removed this unnecessary optimization, but it got left behind with my uncommitted changes for some reason. My bad ^^.

client/profile/user-context-menu.tsx
```diff
@@ -1,34 +1,82 @@
-import React, { useCallback } from 'react'
+import React, { useCallback, useEffect, useRef, useState } from 'react'
```
tec27, line 1: Remove all the changes in this file and let's get this merged, this PR has gone on too long and is too big now for me to really keep coming back to it week after week and trying to put all the context back in my head

PR #802, opened 2021-12-25
Generalize the hotkey handling to all buttons.

2Pacalypse-:
This allows any type of button to register a hotkey which will cause the
button to be pressed programmatically.

Also adds back the hotkey Alt+S to open the settings dialog, as well as
a new hotkey Alt+N to open the notifications popup.

### DIFF
--- a/client/activities/activity-button.tsx
+++ b/client/activities/activity-button.tsx
@@ -132,7 +132,7 @@ export const ActivityButton = React.memo(
       const buttonRef = useRef<HTMLButtonElement>()
       const setButtonRef = useCallback(
         (elem: HTMLButtonElement | null) => {
-          buttonRef.current = elem !== null ? elem : undefined
+          buttonRef.current = elem ?? undefined
           if (ref) {
             if (typeof ref === 'function') {
               ref(elem)
--- a/client/activities/activity-button.tsx
+++ b/client/activities/activity-button.tsx
@@ -1,6 +1,6 @@
-import React, { useMemo } from 'react'
+import React, { useCallback, useMemo, useRef } from 'react'
 import styled, { css, keyframes } from 'styled-components'
-import { useButtonHotkey, useButtonRef, useButtonState } from '../material/button'
+import { HotkeyProp, useButtonHotkey, useButtonState } from '../material/button'
 import { buttonReset } from '../material/button-reset'
 import { Ripple } from '../material/ripple'
 import { blue50, colorTextFaint, colorTextSecondary } from '../styles/colors'
@@ -115,12 +115,7 @@ export interface ActivityButtonProps {
    * A hotkey to register for the button. Pressing the specified modifiers and key will result in
    * the button being clicked programmatically.
    */
-  hotkey?: {
-    keyCode: number
-    altKey?: boolean
-    shiftKey?: boolean
-    ctrlKey?: boolean
-  }
+  hotkey: HotkeyProp
 }
 
 export const ActivityButton = React.memo(
@@ -130,7 +125,22 @@ export const ActivityButton = React.memo(
         disabled,
         onClick,
       })
-      const [buttonRef, setButtonRef] = useButtonRef(ref)
+
+      const buttonRef = useRef<HTMLButtonElement>()
+      const setButtonRef = useCallback(
+        (elem: HTMLButtonElement | null) => {
+          buttonRef.current = elem !== null ? elem : undefined
+          if (ref) {
+            if (typeof ref === 'function') {
+              ref(elem)
+            } else {
+              ref.current = elem
+            }
+          }
+        },
+        [ref],
+      )
+
       useButtonHotkey({ ref: buttonRef, disabled, hotkey })
 
       const labelElems = useMemo(() => {
--- a/client/activities/activity-button.tsx
+++ b/client/activities/activity-button.tsx
@@ -1,7 +1,7 @@
-import React, { useCallback, useMemo, useRef } from 'react'
+import React, { useMemo } from 'react'
 import styled, { css, keyframes } from 'styled-components'
 import KeyListener from '../keyboard/key-listener'
-import { useButtonState } from '../material/button'
+import { useButtonState, useHotkeyHandler } from '../material/button'
 import { buttonReset } from '../material/button-reset'
 import { Ripple } from '../material/ripple'
 import { blue50, colorTextFaint, colorTextSecondary } from '../styles/colors'
@@ -127,48 +127,12 @@ export interface ActivityButtonProps {
 export const ActivityButton = React.memo(
   React.forwardRef<HTMLButtonElement, ActivityButtonProps>(
     ({ label, icon, disabled, glowing, count, onClick, hotkey }, ref) => {
-      const localRef = useRef<HTMLButtonElement>()
       const [buttonProps, rippleRef] = useButtonState({
         disabled,
         onClick,
       })
-      const setRefs = useCallback(
-        (elem: HTMLButtonElement | null) => {
-          localRef.current = elem !== null ? elem : undefined
-          if (ref) {
-            if (typeof ref === 'function') {
-              ref(elem)
-            } else {
-              ref.current = elem
-            }
-          }
-        },
-        [ref],
-      )
-      const onKeyDown = useCallback(
-        (event: KeyboardEvent) => {
-          if (disabled || !hotkey) {
-            return false
-          }
+      const [setRefs, hotkeyHandler] = useHotkeyHandler({ ref, disabled, hotkey })
 
-          if (
-            event.keyCode === hotkey.keyCode &&
-            event.altKey === !!hotkey.altKey &&
-            event.shiftKey === !!hotkey.shiftKey &&
-            event.ctrlKey === !!hotkey.ctrlKey
-          ) {
-            if (localRef.current) {
-              localRef.current.click()
-            } else if (onClick) {
-              onClick()
-            }
-            return true
-          }
-
-          return false
-        },
-        [disabled, hotkey, onClick],
-      )
       const labelElems = useMemo(() => {
         if (disabled || !hotkey) {
           return label
@@ -191,7 +155,7 @@ export const ActivityButton = React.memo(
 
       return (
         <Container ref={setRefs} {...buttonProps}>
-          {hotkey ? <KeyListener onKeyDown={onKeyDown} /> : null}
+          {hotkey ? <KeyListener onKeyDown={hotkeyHandler} /> : null}
           {count !== undefined ? <Count>{count}</Count> : null}
           <IconContainer glowing={glowing}>
             {glowing ? icon : null}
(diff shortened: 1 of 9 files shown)


tec27:
Not really a fan of this approach at all, I don't see why buttons need to have hotkey handling code built in? I think having a special component that wrapped buttons before before was also weird. You could write a simple hook that takes a function to call when the hotkey is pressed and the hotkey to bind and there's no need to write some special integration on every button type any more, no need to use refs in this strange way, etc.

tec27:
Since I don't think the above comment will also result in a change that should probably happen: I see 0 reason to keep using `<KeyListener />` as a component now that hooks exist, and you should probably just make a hook that calls `useEffect`

2Pacalypse-:
Well, this hook I wrote here was supposed to be that "simple" hook that detects if a hotkey was pressed :d. I see now that my mistake was in trying to make it work with the `KeyListener` component as well, instead of having the hook be in charge of all the hotkey handling.

However, with that said, there are a couple of complications which I stumbled upon that I'm not sure how to resolve at this moment:
- Assuming this new hook adds a `keydown` listener on the `document` to detect when a hotkey was pressed, we would need some kind of a mechanism to stop detecting the key presses "behind" an overlay or a dialog. `KeyListener` component uses an `exclusive` prop to deal with this, but I'm not sure if that was the best way to solve this problem there either.
- What I liked about the strange refs approach instead of just calling the `onClick` handler directly is that it simulates the button click fully; along with its ripple effect and everything. Also, if we want to send the `event` argument to the `onClick` handler, then we have a type mismatch between `React.KeyboardEvent` and `React.MouseEvent` and I don't know how to reconcile those types properly :d.

tec27:
If you *want* to add handling for hotkeys anywhere in buttons, the right place is in `useButtonState`, not in each individual button class. Individual button classes deal with appearance + style, they do *not* deal with behavior, event handlers, etc, but this PR breaks all of that and moves us straight back to the approach that I moved away from before.

> Assuming this new hook adds a keydown listener on the document to detect when a hotkey was pressed, we would need some kind of a mechanism to stop detecting the key presses "behind" an overlay or a dialog. KeyListener component uses an exclusive prop to deal with this, but I'm not sure if that was the best way to solve this problem there either.

Look at the source of KeyListener: https://github.com/ShieldBattery/ShieldBattery/blob/e081f99c0490834dab909958d92c3747fb301d81/client/keyboard/key-listener.jsx#L51

It doesn't render anything, it exists *only* to call `componentDidMount` and `componentDidUnmount`. This is identical to a `useEffect` hook, there is zero reason to write components like these any more, or to keep using them inside of function components that can use hooks. 

I'm not sure why you're talking about approaching the `exclusive` thing in a different way. The same approach still works, exactly the same as before. You can extremely easily write a `useKeyListener` function in this *same file*, using a `useEffect` to add/remove event handlers that call the *same event handler*. There's no great rethinking necessary, I'm just telling you to start using hooks here so we don't have to add props to every single button type ever to handle behavior that should instead be handled by the behavior hook.

> What I liked about the strange refs approach instead of just calling the onClick handler directly is that it simulates the button click fully; along with its ripple effect and everything. Also, if we want to send the event argument to the onClick handler, then we have a type mismatch between React.KeyboardEvent and React.MouseEvent and I don't know how to reconcile those types properly :d.

Fine, you can add a `ref` parameter to `useButtonState`, and return a `ref` on its output props that is a multiplexing ref, but you *need to be sure* you have checked every usage of button and moved any refs on the element into the `useButtonState` call. I think this is a bit of a footgun and will be very easy to screw up and break hotkeys when someone accidentally adds a ref on the element, but you can make this decision

2Pacalypse-:
> One way to mitigate the problems would be to add an optional prop to each button whose key is a symbol in the button file, which is only set on the return value of `useButtonState`. You could also tag the ref return value with something (e.g. set a private symbol on it). Then, you could write a simple function that, in dev mode only, takes the value of that prop + the input `ref`. If the prop is true, and the symbol is not defined on the input ref, throw an error as it means they called `useButtonState` but set a different ref on the element (overwriting the `useButtonState` ref). That would at least let us catch these problems at render time in development

I went to do this, but then I wondered is this really necessary now? From what I can see, all buttons forward their refs to the `useButtonState` hook, so I guess I'm failing to see how could someone overwrite the ref we use in the `useButtonState` hook. Can you take a look at my new approach and see if I'm missing something (which is very likely :d).

tec27:
> EDIT: Nvm, I see what you mean now after I saw the failing tests :p. I guess Typescript is also something that would catch this bug for us (since it complains about setting the same prop on an element twice). So your suggested mitigation would only be necessary for JS code (I think?); or we can just manually inspect all the usages of the button now and expect all new code to be Typescript? :d

If TS errors on it then that is sufficient protection to me, any remaining JS code should be considered deprecated at this point and moved to TS as soon as possible. I just don't want it to be something you can silently do and not realize that you're doing the wrong thing.

But yes, you should inspect all existing usages in JS if you haven't already.

client/material/button.tsx
```diff
@@ -27,17 +28,26 @@ export const Label = styled.span`
...
   onBlur?: (event: React.FocusEvent) => void
   onFocus?: (event: React.FocusEvent) => void
   onClick?: (event: React.MouseEvent) => void
   onDoubleClick?: (event: React.MouseEvent) => void
   onMouseDown?: (event: React.MouseEvent) => void
   onMouseEnter?: (event: React.MouseEvent) => void
   onMouseLeave?: (event: React.MouseEvent) => void
-  onKeyDown?: (event: React.KeyboardEvent) => void
-  onKeyUp?: (event: React.KeyboardEvent) => void
+  onKeyDown?: (event: KeyboardEvent) => void
```
tec27, line 49: Honestly I think I would probably leave this as-is and continue attaching these handlers directly on the button. If you register these for the keylistener handler instead, the event is going to be handled in a different spot and thus things like `event.currentTarget` will not be the same. You're likely to break things that depend on that being a particular value (or just cause a lot of headaches for future people that expect the event to be handled at the button level)

client/material/button.tsx
```diff
@@ -171,47 +197,73 @@ export function useButtonState({
...
+                rippleRef.current?.onActivate(event)
+                keyDownActivatedRef.current = true
+              }
+            })
           }
-        })
-      }
+        }
+
+        if (
```
tec27, line 222: As mentioned above, split the hotkey handling back out of this, into a handler that it *only* passed to the keylistener, and leave the previous keyboard handling intact as a React event handler attached to the button. No real reason to do both in the same handler.

client/notifications/activity-bar-entry.tsx
```diff
@@ -84,6 +87,7 @@ export function NotificationsButton() {
           icon={<NotificationsIcon />}
           title={hasUnread ? 'Notifications (unread)' : 'Notifications'}
           onClick={onClick}
+          hotkey={{ keyCode: KEY_N, altKey: true }}
```
tec27, line 90: probably better to pull the entire hotkey config into a file-level constant tbh, I think done this way it will result in it changing every render (and isn't really providing value, since you can just replace the `KEY_N` constant above with something like: `const ALT_N: HotkeyProp = { keyCode: keycode('n'), altKey: true }`

Probably apply that everywhere these are used

tec27 (changes requested):

client/material/button.tsx
```diff
@@ -171,47 +197,73 @@ export function useButtonState({
...
-        keyDownActivatedRef.current = true
-      } else {
-        keyDownActivatedRef.current = false
-
-        requestAnimationFrame(() => {
-          // Sometimes browsers don't active the element until after the event has been processed,
-          // so we check again on the next frame
-          if (activationTarget.matches(':active')) {
+  useKeyListener({
+    onKeyDown: useCallback(
```
tec27, line 201: Also there's no value in `useCallback`ing stuff passed to `useKeyListener`, it just stores it in refs anyway. You're just doing needless bookkeeping if you do this

client/material/button.tsx
```diff
@@ -28,26 +28,42 @@ export const Label = styled.span`
   white-space: nowrap;
 `
 
-export interface HotkeyProp {
-  keyCode: number
-  altKey?: boolean
-  shiftKey?: boolean
-  ctrlKey?: boolean
+export function useButtonRef(
```
tec27, line 31: I find this function pretty strange to be here and to be named this way. It really has nothing to do with buttons at all, it's just a multiplexing ref? If you want that, I guess fine, but it should probably live elsewhere and be named as such (and take a generic param instead of being hardcoded to `HTMLButtonElement`). If you're having trouble with a name, I'd probably do `useMultiRef`

client/material/button.tsx
```diff
@@ -272,12 +246,54 @@ export function useButtonState({
...
+export interface HotkeyProp {
+  keyCode: number
+  altKey?: boolean
+  shiftKey?: boolean
+  ctrlKey?: boolean
+}
+
+export interface ButtonHotkeyProps {
+  /** The reference to the button that should be pressed programmatically. */
+  ref?: React.MutableRefObject<HTMLButtonElement | undefined>
```
tec27, line 266: This being optional doesn't make any sense. Hotkeys don't work if you don't have a ref?

client/material/button.tsx
```diff
@@ -272,12 +246,54 @@ export function useButtonState({
...
+  /** Whether the button is disabled (hotkey will do nothing). */
+  disabled?: boolean
+  /**
+   * A hotkey to register for the button. Pressing the specified modifiers and key will result in
+   * the button being clicked programmatically.
+   */
+  hotkey?: HotkeyProp
```
tec27, line 273: I don't really get why this is optional either, what is the point of using this if you don't have a hotkey?

tec27: Okay I looked lower and I see now. This is not a good structure, this makes it so every single button will register a global key listener, which is potentially expanding the number of key listeners drastically. `useButtonHotkey` operates only on a ref, I don't see why the button types need any understanding of hotkeys whatsoever once you've added this hook. So you should make these fields required and remove the usage from the button types below.

tec27 (changes requested):

2Pacalypse-:
Depends on #812

client/activities/activity-button.tsx
```diff
@@ -116,25 +115,24 @@ export interface ActivityButtonProps {
...
         disabled,
         onClick,
       })
-      const setRefs = useCallback(
+
+      const buttonRef = useRef<HTMLButtonElement>()
+      const setButtonRef = useCallback(
         (elem: HTMLButtonElement | null) => {
-          localRef.current = elem !== null ? elem : undefined
+          buttonRef.current = elem !== null ? elem : undefined
```
tec27, line 135: Think the safer thing to do now is probably just `elem ?? undefined` (this current logic is risky if this ref was passed into another multiplexing ref that also did this conversion)

tec27 (approved):

PR #803, opened 2021-12-25
Generalize the hotkey label to all buttons as well.

2Pacalypse-:
Well, except the IconButton. IconButtons should probably display their
hotkeys in a tooltip or something.

### DIFF
--- a/client/material/button.tsx
+++ b/client/material/button.tsx
@@ -1,4 +1,4 @@
-import React, { useCallback, useMemo, useRef, useState } from 'react'
+import React, { useCallback, useRef, useState } from 'react'
 import styled from 'styled-components'
 import { assertUnreachable } from '../../common/assert-unreachable'
 import { useKeyListener } from '../keyboard/key-listener'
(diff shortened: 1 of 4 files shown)


2Pacalypse-:
Depends on #802

client/material/button.tsx
```diff
@@ -307,6 +307,42 @@ export function useHotkeyHandler({
...
+export interface HotkeyedLabelProps {
+  /** Whether the button is disabled (label will remain unchanged). */
+  disabled?: boolean
+  /** A hotkey that is registered for the button. The label will underline the hotkeyed letter. */
+  hotkey?: HotkeyProp
+  /** The given label of the button. If not a string, it will remain unchanged. */
+  label: string | React.ReactNode
+}
+
+export function useHotkeyedLabel({ disabled, hotkey, label }: HotkeyedLabelProps) {
```
tec27, line 319: I'd prefer this was implemented as a component rather than a hook. As a general rule, try to avoid using JSX or returning elements in hooks. So phrased differently, if the thing needs to return an Element, it should be a component (or split between a hook + a component, but in this case no hook really needs to exist)

You can probably just memoize the component as well in this case, instead of doing `useMemo` inside of it

tec27 (approved):

PR #804, opened 2021-12-26
Add hotkeys to tab items as well.

2Pacalypse-:
This PR adds chrome-style hotkeys to all tab items. Mainly Ctrl+1 to
Ctrl+9; with Ctrl+9 being a special hotkey that always focuses the last
tab in the list, regardless of how many tabs there are. Additionally,
Ctrl-PageUp / Ctrl-PageDown hotkeys can be used to cycle through
tabs as well.

This PR also adds support for all buttons to register multiple hotkeys
for a single button.

### DIFF
--- a/client/main-layout.tsx
+++ b/client/main-layout.tsx
@@ -161,7 +161,7 @@ export function MainLayout() {
 
   const searchingMatchButtonRef = useRef<HTMLButtonElement>(null)
   const settingsButtonRef = useRef<HTMLButtonElement>(null)
-  useButtonHotkey({ ref: settingsButtonRef, hotkeys: [ALT_S] })
+  useButtonHotkey({ ref: settingsButtonRef, hotkey: ALT_S })
 
   useEffect(() => {
     dispatch(openChangelogIfNecessary())
@@ -280,7 +280,7 @@ export function MainLayout() {
       label='Find match'
       onClick={onFindMatchClick}
       disabled={inGameplayActivity}
-      hotkeys={[ALT_F]}
+      hotkey={ALT_F}
     />
   ) : (
     <ActivityButton
@@ -290,7 +290,7 @@ export function MainLayout() {
       glowing={true}
       label='Finding…'
       onClick={() => setSearchingMatchOverlayOpen(true)}
-      hotkeys={[ALT_F]}
+      hotkey={ALT_F}
     />
   )
   const activityButtons = IS_ELECTRON
@@ -302,14 +302,14 @@ export function MainLayout() {
           label='Create'
           onClick={onCreateLobbyClick}
           disabled={inGameplayActivity}
-          hotkeys={[ALT_C]}
+          hotkey={ALT_C}
         />,
         <ActivityButton
           key='join-game'
           icon={<JoinGameIcon />}
           label='Join'
           onClick={onJoinLobbyClick}
-          hotkeys={[ALT_J]}
+          hotkey={ALT_J}
           count={lobbyCount > 0 ? lobbyCount : undefined}
         />,
         <ActivityButton
@@ -317,21 +317,21 @@ export function MainLayout() {
           icon={<StyledMapsIcon />}
           label='Maps'
           onClick={onMapsClick}
-          hotkeys={[ALT_M]}
+          hotkey={ALT_M}
         />,
         <ActivityButton
           key='replays'
           icon={<ReplaysIcon />}
           label='Replays'
           onClick={onReplaysClick}
-          hotkeys={[ALT_R]}
+          hotkey={ALT_R}
         />,
         <ActivityButton
           key='ladder'
           icon={<LadderIcon />}
           label='Ladder'
           onClick={() => navigateToLadder()}
-          hotkeys={[ALT_D]}
+          hotkey={ALT_D}
         />,
         <ActivitySpacer key='spacer' />,
       ]
@@ -341,14 +341,14 @@ export function MainLayout() {
           icon={<DownloadIcon />}
           label='Download'
           onClick={() => dispatch(openDialog(DialogType.Download))}
-          hotkeys={[ALT_O]}
+          hotkey={ALT_O}
         />,
         <ActivityButton
           key='ladder'
           icon={<LadderIcon />}
           label='Ladder'
           onClick={() => navigateToLadder()}
-          hotkeys={[ALT_D]}
+          hotkey={ALT_D}
         />,
         <ActivitySpacer key='spacer' />,
       ]
--- a/client/main-layout.tsx
+++ b/client/main-layout.tsx
@@ -161,7 +161,7 @@ export function MainLayout() {
 
   const searchingMatchButtonRef = useRef<HTMLButtonElement>(null)
   const settingsButtonRef = useRef<HTMLButtonElement>(null)
-  useButtonHotkey({ ref: settingsButtonRef, hotkey: ALT_S })
+  useButtonHotkey({ ref: settingsButtonRef, hotkeys: [ALT_S] })
 
   useEffect(() => {
     dispatch(openChangelogIfNecessary())
@@ -280,7 +280,7 @@ export function MainLayout() {
       label='Find match'
       onClick={onFindMatchClick}
       disabled={inGameplayActivity}
-      hotkey={ALT_F}
+      hotkeys={[ALT_F]}
     />
   ) : (
     <ActivityButton
@@ -290,7 +290,7 @@ export function MainLayout() {
       glowing={true}
       label='Finding…'
       onClick={() => setSearchingMatchOverlayOpen(true)}
-      hotkey={ALT_F}
+      hotkeys={[ALT_F]}
     />
   )
   const activityButtons = IS_ELECTRON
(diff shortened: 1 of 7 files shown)


2Pacalypse-:
Depends on #803

tec27:
Doesn't seem like this PR got updated for the later stuff so I'll hold off on reviewing it for now

client/main-layout.tsx
```diff
@@ -341,14 +341,14 @@ export function MainLayout() {
...
-          hotkey={ALT_O}
+          hotkeys={[ALT_O]}
         />,
         <ActivityButton
           key='ladder'
           icon={<LadderIcon />}
           label='Ladder'
           onClick={() => navigateToLadder()}
-          hotkey={ALT_D}
+          hotkeys={[ALT_D]}
```
tec27, line 351: If you do this you're creating a new array every time and thus the whole reason for pulling these out (that they memoize properly) is lost

client/material/button.tsx
```diff
@@ -285,12 +290,12 @@ export interface HotkeyedTextProps {
...
  * it will be used so we leave it here.
  */
-export const HotkeyedText = React.memo(({ hotkey, text, disabled }: HotkeyedTextProps) => {
+export const HotkeyedText = React.memo(({ hotkeys, text, disabled }: HotkeyedTextProps) => {
   if (disabled) {
     return <>{text}</>
   }
 
-  const hotkeyString = String.fromCharCode(hotkey.keyCode).toLowerCase()
+  const hotkeyString = String.fromCharCode(hotkeys[0].keyCode).toLowerCase()
```
tec27, line 298: It doesn't really make sense to me to do this here, just accept a single hotkey and let the component using it decide which hotkey that is

client/material/tabs.tsx
```diff
@@ -119,10 +151,17 @@ export function Tabs<T>({ children, activeTab, onChange, className }: TabsProps<
         return child
       }
 
+      const hotkeys = [{ keyCode: KEY_NUMBERS[i], ctrlKey: true }]
```
tec27, line 154: Since you added `hotkeys` to TabItemProps, isn't it possible someone sets that now? Shouldn't this forward the existing value if it's set?

tec27 (changes requested):

client/state-hooks.ts
```diff
@@ -68,3 +68,28 @@ export function useForceUpdate(): () => void {
...
+ * local ref in a component that already forwards a ref to another component below it.
+ */
+export function useMultiRef<T>(
+  ref: React.ForwardedRef<T>,
+): [elemRef: React.MutableRefObject<T | undefined>, setElemRef: (elem: T | null) => void] {
+  const elemRef = useRef<T>()
+  const setElemRef = useCallback(
+    (elem: T | null) => {
+      elemRef.current = elem ?? undefined
+      if (ref) {
```
tec27, line 83: If we're dealing with `ForwardedRef` isn't this always true? (If you want to accept non-forwarded refs as well, I'd probably make the function param optional + use a different type)

client/material/tabs.tsx
```diff
@@ -151,13 +151,22 @@ export function Tabs<T>({ children, activeTab, onChange, className }: TabsProps<
         return child
       }
 
-      const hotkeys = [{ keyCode: KEY_NUMBERS[i], ctrlKey: true }]
+      const childHotkeys = (child.props as TabItemProps<T>).hotkeys
+      const hotkeys: HotkeyProp[] = []
+
+      if (Array.isArray(childHotkeys)) {
```
tec27, line 147: This is a weird way to check if something is undefined :)

tec27: An easier way to do this is just change the above like to `const hotkeys = (childHotkeys ?? []).slice()`

client/material/tabs.tsx
```diff
@@ -174,5 +177,52 @@ export function Tabs<T>({ children, activeTab, onChange, className }: TabsProps<
...
+        if (enabledChildren.length < 2) {
+          return false
+        }
+
+        let previousOrNextTab: T | undefined
+        React.Children.forEach(enabledChildren, (child, i) => {
+          let previousOrNextChild
+          if (event.keyCode === PAGEUP) {
+            previousOrNextChild =
+              i === enabledChildren.length - 1 ? enabledChildren[0] : enabledChildren[i + 1]
```
tec27, line 199: The logic on this line seems backwards to me (they hit page up, but you're going to the next tab?). You can also use modulus to do this without a ternary:

previous: `(i - 1 + enabledChildren.length) % enabledChildren.length`
next: `(i + 1) % enabledChildren.length`

client/material/tabs.tsx
```diff
@@ -174,5 +177,52 @@ export function Tabs<T>({ children, activeTab, onChange, className }: TabsProps<
...
+          } else {
+            previousOrNextChild =
+              i === 0 ? enabledChildren[enabledChildren.length - 1] : enabledChildren[i - 1]
+          }
+
+          if (
+            child &&
+            previousOrNextChild &&
+            (previousOrNextChild.props as TabItemProps<T>).value === activeTab
```
tec27, line 208: I can't really understand what this check is meant to do? Can you explain what you're trying to accomplish here?

To me this reads as "if the tab we're going to is already the active one, send an onChange event for it" which seems... backwards

client/material/tabs.tsx
```diff
@@ -174,5 +177,52 @@ export function Tabs<T>({ children, activeTab, onChange, className }: TabsProps<
...
+      if ((event.keyCode === PAGEUP || event.keyCode === PAGEDOWN) && event.ctrlKey === true) {
+        const enabledChildren: React.ReactElement[] = []
+        React.Children.forEach(children, (child, i) => {
+          if (child && !(child.props as TabItemProps<T>).disabled) {
+            enabledChildren.push(child)
+          }
+        })
+
+        if (enabledChildren.length < 2) {
+          return false
```
tec27, line 191: Feel like this should probably return true in this case: The hotkey was right but we're not in a state where it would do anything, so we're exiting early

tec27 (changes requested):

client/material/tabs.tsx
2Pacalypse-: It is very likely that I made this code more convoluted than it needs to be, but my thinking was that in order to find the previous/next tab to switch to, I have to iterate through all the (enabled) tabs and check if the active tab is the one in front/behind of the one I'm looking for. This also explains why the logic above seems backwards. For example, to find the _previous_ tab, I have to check if the _next_ tab is the active one.

client/state-hooks.ts
2Pacalypse-: Typescript says that it might be `null`. And while I do think there's value in expanding this hook to include non-forwarded refs, I'll leave that as an exercise to a future reader if/when it's needed since I can't be arsed to fiddle with these types anymore :).

To be more serious, I'll do it if you think it should be done now.

client/material/tabs.tsx
2Pacalypse-: I did think (or rather found a stackoverflow post :p) of using the modulus operator, but to be perfectly honest I just find that a lot less readable/understandable.

client/state-hooks.ts
tec27: Well, in the future the easier answer to my question was to look at the types to ForwardedRef, which I didn't do before because I didn't have an editor open :P

```ts
type ForwardedRef<T> = ((instance: T | null) => void) | MutableRefObject<T | null> | null;
```

So yeah, it can (apparently) be null.

I don't think any further modifications are necessary to take non-forwarded refs, given that typing (although the named type does feel a bit strange)

client/material/tabs.tsx
tec27: We should step back and rethink this whole thing then, I don't think this is really maintainable as-is and is very likely to be a source of bugs going forward

tec27: So the goal here is, "call onChange with the value of the tab that should be newly selected". So first thing's first, this function can exit early if onChange isn't set, there's nothing useful we can do in that case.

Then, when you loop through the children to make an `enabledChildren` list, *always* include the currently selected tab whether it is enabled or not. Store the index of that tab when you find it.

Next, we made an array of all enabled tabs above. There's no need to loop through things again, we have all that we need in that array:

```ts
if (event.keyCode === PAGEUP) {
  const previousIndex = (selectedIndex - 1 + enabledChildren.length) % enabledChildren.length
  const tab = enabledChildren[previousIndex]
  onChange(tab.props.value)
} else if (event.keycode === PAGEDOWN) {
  const nextIndex = selectedIndex + 1 % enabledChildren.length
  const tab = enabledChildren[nextIndex]
  onChange(tab.props.value)
}
```

Much simpler

client/material/tabs.tsx
```diff
@@ -139,5 +171,52 @@ export function Tabs<T>({ children, activeTab, onChange, className }: TabsProps<
...
+            enabledChildren.push(child)
+          }
+        })
+
+        if (enabledChildren.length < 2) {
+          return true
+        }
+
+        let previousOrNextTab: T | undefined
+        React.Children.forEach(enabledChildren, (child, i) => {
```
tec27, line 189: Just for future reference, this is not something you should be doing. `React.Children.forEach` is meant for looping over the children of an element (as given to you by React), you should never be using it to loop over arrays that you yourself created, even if they contain elements.

PR #809, opened 2022-01-15
Add a Tooltip component and use it to display full dates in user profile

2Pacalypse-:
overlay join date and chat message timestamp.

There are a few issues remaining that I haven't figured out how best to
resolve yet:
- Tooltip open state tracking needs to be reworked. Using simple
  mouseenter/mouseleave events to track the Tooltip open state is not
  good enough. For example, it's possible for Tooltip to be displayed
  above its target element (in case of popover clamping at the edge of a
  screen for example), in which case the 'mouselave' event of the target
  element is triggered if the Tooltip is shown right below the mouse
  cursor.

  Also, there's some weirdness when Tooltip is opened above one element
  that also has a Tooltip and then moving the mouse over that element
  quickly can show/hide its Tooltip immediately. This is most easily
  reproducible on hovering over chat message timestamps by going from a
  below message to the one right above it.
- The transition used to open/close the Tooltip is copy-pasted from
  Popovers, with a delay of 800ms. This is probably not a correct thing
  for Tooltips, but then again, I'm not sure what is.
- I've used our lightest background color for Tooltips, since they can
  be shown over all other content, including Popovers which are fairly
  light themselves. However, I'm not sure if this was the best choice,
  but I wasn't sure what else to choose.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -217,7 +217,7 @@
     "no-this-before-super": 2,
     "no-throw-literal": 2,
     "no-trailing-spaces": 2,
-    "no-undef": 2,
+    "no-undef": 0,
     "no-undef-init": 2,
     "no-undefined": 0,
     "no-underscore-dangle": 0,
--- a/.eslintrc
+++ b/.eslintrc
@@ -59,6 +59,7 @@
 
   "globals": {
     "document": "readonly",
+    "JSX": "readonly",
     "navigator": "readonly",
     "window": "readonly",
     "IS_ELECTRON": "readonly",
--- a/.eslintrc
+++ b/.eslintrc
@@ -59,7 +59,6 @@
 
   "globals": {
     "document": "readonly",
-    "JSX": "readonly",
     "navigator": "readonly",
     "window": "readonly",
     "IS_ELECTRON": "readonly",
--- a/client/material/tooltip.tsx
+++ b/client/material/tooltip.tsx
@@ -49,7 +49,7 @@ const arrowStyle: Record<TooltipPosition, FlattenSimpleInterpolation> = {
   `,
 }
 
-const TooltipContent = styled.div<{ $position?: TooltipPosition }>`
+const TooltipContent = styled.div<{ $position: TooltipPosition }>`
   ${caption};
   ${shadow2dp};
 
@@ -75,7 +75,7 @@ const TooltipContent = styled.div<{ $position?: TooltipPosition }>`
     background-color: inherit;
     border: 1px solid rgba(255, 255, 255, 0.36);
 
-    ${props => arrowStyle[props.$position!]};
+    ${props => arrowStyle[props.$position]};
   }
 `
 
@@ -101,10 +101,10 @@ interface TooltipProps {
   /**
    * A custom component that will be used instead of the default Tooltip content element. Can be
    * used if you wish to customize the Tooltip style. Will get injected with the following props:
-   *  - $position: TooltipPosition
-   *  - children: string (the Tooltip's text)
+   *  - $position: the `TooltipPosition` of the content
+   *  - children: the Tooltip's text
    * */
-  contentComponent?: React.ReactNode
+  ContentComponent?: React.ComponentType<{ $position: TooltipPosition; children: React.ReactNode }>
 }
 
 /**
@@ -119,7 +119,7 @@ export function Tooltip({
   className,
   ariaLabel = text,
   position = 'bottom',
-  contentComponent = <TooltipContent />,
+  ContentComponent = TooltipContent,
 }: TooltipProps) {
   const [open, setOpen] = useState(false)
   const [anchorElem, setAnchorElem] = useState<HTMLElement>()
@@ -186,11 +186,6 @@ export function Tooltip({
     originY = 'center'
   }
 
-  const tooltipContent = React.cloneElement(contentComponent as JSX.Element, {
-    $position: position,
-    children: text,
-  })
-
   return (
     <>
       <div
@@ -210,7 +205,7 @@ export function Tooltip({
                 originX={originX}
                 originY={originY}
                 styles={styles}>
-                {tooltipContent}
+                <ContentComponent $position={position}>{text}</ContentComponent>
               </NoPointerPopoverContent>
             </NoPointerPortal>
           ),
--- a/client/material/tooltip.tsx
+++ b/client/material/tooltip.tsx
@@ -1,0 +1,121 @@
+import React, { useState } from 'react'
+import { useTransition, UseTransitionProps } from 'react-spring'
+import styled from 'styled-components'
+import { background100 } from '../styles/colors'
+import { caption } from '../styles/typography'
+import { OriginX, OriginY, PopoverContent, useAnchorPosition } from './popover'
+import { Portal } from './portal'
+import { defaultSpring } from './springs'
+
+const isDev = __WEBPACK_ENV.NODE_ENV !== 'production'
+
+const Container = styled.div`
+  ${caption};
+  display: flex;
+  justify-content: center;
+  align-items: center;
+
+  min-height: 24px;
+  padding: 4px 8px;
+  border-radius: 4px;
+  background-color: ${background100};
+`
+
+export type TooltipPosition = 'left' | 'right' | 'top' | 'bottom'
+
+interface TooltipProps {
+  /** The text that should be displayed in the Tooltip. */
+  text: string
+  /** The child element that the Tooltip should be linked to. Should only be a single element. */
+  children: React.ReactNode
+  /** One of the four sides that can be used to position the Tooltip. Defaults to 'bottom'. */
+  position?: TooltipPosition
+  /** Class name applied to the root container of the Tooltip (that has the background, etc.). */
+  className?: string
+}
+
+/**
+ * A component that displays some content in a floated UI element when a user hovers over a target
+ * element. Should generally be used over native `title` attribute on elements.
+ *
+ * Utilizes popovers to deal with positioning, but with a much simpler API than that of a Popover.
+ */
+export function Tooltip({ text, children, position = 'bottom', className }: TooltipProps) {
+  const [anchorElem, setAnchorElem] = useState<HTMLElement>()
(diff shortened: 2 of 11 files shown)


client/material/tooltip.tsx
```diff
@@ -0,0 +1,121 @@
...
+const Container = styled.div`
+  ${caption};
+  display: flex;
+  justify-content: center;
+  align-items: center;
+
+  min-height: 24px;
+  padding: 4px 8px;
+  border-radius: 4px;
+  background-color: ${background100};
```
tec27, line 21: So first, your comment is a lie about this being the "lightest", there is a `background50` :)

But, primary text color doesn't come anywhere close to passing contrast on anything lighter than `background400` so this is a no-go. You can pass contrast by flipping the text to invert (e.g. black 87%) but it feels pretty odd to have black text in general imo.

I would try for some other visual treatment, darker colors with a light border maybe? And I would recommend getting the Contrast plugin in Figma and using it to check these things so that I don't have to check everything myself (Chrome can also tell you in devtools, but it's a bit hard for this particular element since it disappears)

client/material/tooltip.tsx
```diff
@@ -0,0 +1,121 @@
...
+  )
+
+  const onMouseEnter = (event: React.MouseEvent) => {
+    setAnchorElem(event.currentTarget as HTMLElement)
+  }
+  const onMouseLeave = (event: React.MouseEvent) => {
+    setAnchorElem(undefined)
+  }
+
+  const childrenCount = React.Children.count(children)
```
tec27, line 69: So *if* you wanted to do this, React already has a thing for this, don't really need to build your own: https://reactjs.org/docs/react-api.html#reactchildrenonly

However I think this is a bad idea. I don't see why you'd ever need to do this, and it's not type-safe for you to do what you're doing with the props here (you have no restrictions on the passed in children to ensure they can take these props).

A better way imo is to wrap the children and attach these event handlers to that element. I would apply this components `className` to that wrapper and allow a custom `PopoverContent` replacement to be passed in if they need to customize the tooltip style (but honestly that should be extremely rare, these things are probably very consistent)

tec27: Another alternative to this approach, if you want to avoid the wrapper, is to split the component into two parts: the UI and a hook. The UI is mostly as this is, the hook returns the necessary event handlers/props to attach to the element being wrapped.

For this case though, since our content is *almost always* block content (rather than e.g. spans), the wrapper seems a lot easier. Looking at Discord's implementation (you can use React dev tools in the browser one), it seems like they allow the wrapper behavior to be turned on or off, not quite sure why they need that complexity but 🤷

client/messaging/message-layout.tsx
```diff
@@ -41,11 +42,13 @@ interface MessageTimestampProps {
 }
 
 export const MessageTimestamp = (props: MessageTimestampProps) => (
-  <Timestamp title={longTimestamp.format(props.time)}>
-    <Separator>[</Separator>
-    {shortTimestamp.format(props.time)}
-    <Separator>] </Separator>
-  </Timestamp>
+  <Tooltip text={longTimestamp.format(props.time)} position='top'>
+    <Timestamp>
```
tec27, line 46: I feel like whenever we use this thing, we should probably also be including the tooltip text in an aria attribute (e.g. `aria-label`). Tooltips aren't particular accessible to screenreaders

tec27: (With a wrapper element you could probably do this as part of the Tooltip component actually, although you may want to make it customizable for things that have more content than just what is in the tooltip. Or maybe there's a better ARIA attribute, I dunno)

tec27 (changes requested):

.eslintrc
```diff
@@ -59,6 +59,7 @@
 
   "globals": {
     "document": "readonly",
+    "JSX": "readonly",
```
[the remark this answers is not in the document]
2Pacalypse-: It was always necessary IIRC, we just haven't used it so far. Granted, after addressing the PR comments we still don't need it, so I'll remove it.

But after looking into this a bit more deeply, it seems like we should probably turn off the `no-undef` rule for Typescript files instead at least?
https://typescript-eslint.io/docs/linting/troubleshooting#i-get-errors-from-the-no-undef-rule-about-global-variables-not-being-defined-even-though-there-are-no-typescript-errors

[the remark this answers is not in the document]
tec27: Yeah, you can turn it off for everything (TS checks JS files at a basic level too I think? At the very least in the editor it does)

client/material/tooltip.tsx
```diff
@@ -0,0 +1,255 @@
...
+  /** Event handler that will be called when a cursor leaves the Tooltip target area. */
+  onMouseLeave: (event: React.MouseEvent) => void
+}
+
+/**
+ * A helper component that wraps the children that were sent to the Tooltip component, thus ensuring
+ * that the Tooltip component itself always has a single child on which to attach the mouse event
+ * listeners.
+ */
+function TooltipChildrenWrapper({
```
tec27, line 109: I am extremely unsure of what value this adds or why it exists at all? Or why we declared props for it? It seems like you started implementing my comments backwards (e.g. making it possible to replace the wrapper div instead of making it possible to replace the content div), realized that halfway through and did it the correct way in the component below, but then left this component here anyway?

Just inline this in the component below?

client/material/tooltip.tsx
```diff
@@ -0,0 +1,255 @@
...
+  text,
+  children,
+  className,
+  onMouseEnter,
+  onMouseLeave,
+}: TooltipChildrenWrapperProps) {
+  return (
+    <div
+      className={className}
+      aria-label={text}
```
tec27, line 119: Like I mentioned in the preview review, this should probably be customizable, it won't *always* make sense to label the tooltip anchor the same as the tooltip I don't think. You can default it to the tooltip text by doing `ariaLabel = text` in the function params

client/material/tooltip.tsx
```diff
@@ -0,0 +1,255 @@
...
+          open && (
+            <NoPointerPortal open={open}>
+              <NoPointerPopoverContent
+                anchorX={anchorX}
+                anchorY={anchorY}
+                originX={originX}
+                originY={originY}
+                styles={styles}>
+                {tooltipContainer ? (
+                  tooltipContainer
```
tec27, line 245: I don't think it makes sense for this to work this way, it should be treated as a replacement for TooltipContent and receive the same props (`$position`, children). If the replacer doesn't care about those things they can ignore them, if they do it prevents them having to re-implement that stuff themselves.

tec27 (changes requested):

client/material/tooltip.tsx
2Pacalypse-: That's exactly what happened :)

Now I wish I haven't spent all this time writing JSDoc for it, but oh well...

client/material/tooltip.tsx
```diff
@@ -0,0 +1,220 @@
...
+  &::before {
+    content: '';
+    position: absolute;
+    width: 8px;
+    height: 8px;
+
+    background-color: inherit;
+    border: 1px solid rgba(255, 255, 255, 0.36);
+
+    ${props => arrowStyle[props.$position!]};
```
tec27, line 78: If the intention is that this is always specified, you should just make it that way on the props. Otherwise, this should probably be `props.$position ? arrowStyle[props.$position] : ''` or something. (Since this isn't exported, I'd generally prefer just making the prop required)

client/material/tooltip.tsx
```diff
@@ -0,0 +1,220 @@
...
+    originY = 'bottom'
+    anchorY = anchorY - 8
+  } else if (anchorOriginY === 'bottom') {
+    originY = 'top'
+    anchorY = anchorY + 8
+  } else {
+    originY = 'center'
+  }
+
+  const tooltipContent = React.cloneElement(contentComponent as JSX.Element, {
```
tec27, line 189: Nah, this was not the intention. The goal is that someone passes in a *component class/function*, not a whole element. You would then just use it as `<contentComponent $position={position}>{text}</contentComponent>` in the code below. In TS this will be a `React.ComponentType<{ $position, children }>`.

In general if you're using `React.cloneElement` I think you're probably doing the wrong thing, just for future reference. Extremely hard to ever manage types around those safely and they're just a pretty weird design

client/profile/user-profile-overlay.tsx
```diff
@@ -236,9 +237,9 @@ function OverlayContents({ userId, onDismiss }: { userId: SbUserId; onDismiss: (
         <>
           <div>
             <SectionHeader>Info</SectionHeader>
-            <Body1 title={longTimestamp.format(profile.created)}>
-              Joined {joinDateFormat.format(profile.created)}
-            </Body1>
+            <Tooltip text={longTimestamp.format(profile.created)}>
```
tec27, line 240: imo the aria-label here should be something more like `Joined ${longFormattedDate}`, I believe once you add that attribute it's going to ignore the children of this element so it should reproduce the full thing if possible

tec27 (changes requested):

client/profile/user-profile-overlay.tsx
2Pacalypse-: Can you clarify what you mean by "once you add that attribute it's going to ignore the children of this element"?

client/material/tooltip.tsx
2Pacalypse-: Apparently TypeScript complains if a component doesn't start with a capitalized letter so I named the prop `ContentComponent`.

client/profile/user-profile-overlay.tsx
tec27: As in, the things using the ARIA attributes (e.g. screenreaders) will read out the `aria-label` rather than the text content of the children

tec27: Probably I guess we should be marking the Portal content as `aria-hidden` as well, since it doesn't follow the normal document hierarchy

tec27: http://whatsock.com/training/matrices/#:~:text=3.1.11%3A%20Tooltip%20Role maybe this is a better solution tbh, I may just try to implement this so we can get this checked in without further back-and-forth

PR #812, opened 2022-01-29
TS-ify the MainLayout and move it to a functional component.

2Pacalypse-:


client/main-layout.tsx
```diff
@@ -0,0 +1,401 @@
...
+      <Content>
+        <Switch>
+          <ConditionalRoute
+            path='/admin/:rest*'
+            filters={[IsAdminFilter]}
+            component={LoadableAdminPanel}
+          />
+          <Route path='/chat' component={ChatList} />
+          <Route path='/chat/:channel' component={ChatChannel} />
+          <Route path='/ladder/:matchmakingLabel?'>
```
tec27, line 349: 1) Why is this "matchmakingLabel" and not "matchmakingType"?
2) imo this isn't a great approach for these because it makes it hard for these components to add or change the parameters they accept. Ideally they would be the maintainers of all routes underneath their respective root path, and not have to come back up and modify main-layout every time they make a change (thus, this should simply be `/ladder/:rest*` and pass all the parameters into a component in the ladder package). Apply throughout this file

tec27 (changes requested):
This is extremely hard to review. For big moves like this you need to make *absolutely certain* that git registers it as a rename, and not a delete/add. I can't really tell what has changed in this file as it is

client/main-layout.tsx
```diff
@@ -0,0 +1,401 @@
...
+      dispatch(addTermsOfServiceNotificationIfNeeded())
+      dispatch(addAcceptableUseNotificationIfNeeded())
+    }
+
+    return () => {
+      firstLoggedIn = true
+    }
+  }, [isEmailVerified, dispatch])
+
+  const onFindMatchClick = () => {
```
tec27, line 158: So I know we talked about how we're probably overusing `useCallback`, *but*, when you're updating existing code, especially the code that renders basically everything else in the app, you should probably try to maintain the existing semantics as much as possible. 

If you can absolutely verify that having stable references for these isn't necessary, sure, you can make that change. But it doesn't really seem like that's the case here, so we should probably leave them as-is (that is, wrap these in `useCallback`).

For new code this decision often has less consequences, but for stuff like this I'd focus on changing as little as possible unless you know you have covered all the permutations (ideally with tests, but at the very least with manual testing)

client/main-layout.tsx
```diff
@@ -0,0 +1,401 @@
...
+
+  const onRemoveMap = (map: MapInfoJson) => {
+    dispatch(removeMap(map))
+  }
+
+  const onRegenMapImage = (map: MapInfoJson) => {
+    dispatch(regenMapImage(map))
+  }
+
+  const serverMapsProps = {
```
tec27, line 208: This is a circular dependency, and once you useCallback onUploadMap that becomes very clear:

onUploadMap uses serverMapProps which uses onUpoadMap

It's also declared *after* it's being used, which is something you should generally avoid. I don't think this really needs to exist as an object and you should probably pass these props directly instead of doing stuff like this (this is also bad from a typing perspective because it is effectively `any`)

client/main-layout.tsx
```diff
@@ -115,315 +118,284 @@ const FadedSettingsIcon = styled(SettingsIcon)`
...
+    onRegenMapImage,
   }
 
-  onMapsClick = () => {
-    if (!isShieldBatteryHealthy(this.props)) {
-      this.props.dispatch(openDialog(DialogType.ShieldBatteryHealth))
-    } else if (!isStarcraftHealthy(this.props)) {
-      this.props.dispatch(openDialog(DialogType.StarcraftHealth))
+  const onMapsClick = () => {
+    if (!isShieldBatteryHealthy({ starcraft })) {
```
tec27, line 218: it is incredibly weird to me that we require a valid installation to browse server maps so... yeah. I'm gonna leave it as is for now because that's how it was but I mean, whatever reason we have for doing this here, we should fix that

client/activities/activity-button.tsx
```diff
@@ -125,7 +125,7 @@ export interface ActivityButtonProps {
 }
 
 export const ActivityButton = React.memo(
-  React.forwardRef<HTMLButtonElement, ActivityButtonProps>(
+  React.forwardRef<HTMLButtonElement | undefined, ActivityButtonProps>(
```
tec27, line 128: This is the wrong change to make. React uses a `null` initial ref value to indicate that it will completely control the ref (and that the ref is not mutable by outside APIs). The types also encode this.

The right fix here is to initialize the refs you're passing into this with `null`

PR #821, opened 2022-02-19
Remove `editAllChannels` permission.

2Pacalypse-:
I have no idea what was the idea behind this permission when I first
added it, but I'm fairly certain that it's way too general and ambiguous
for it to make sense to keep.

### DIFF
--- a/client/auth/auth-records.ts
+++ b/client/auth/auth-records.ts
@@ -16,7 +16,6 @@ export class PermissionsRecord
   extends Record({
     editPermissions: false,
     debug: false,
-    editAllChannels: false,
     banUsers: false,
     manageMaps: false,
     manageMapPools: false,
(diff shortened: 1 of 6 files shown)


tec27:
Is the intention that `moderateChatChannels` serves the same purpose? What is the permission that controls editing channels that you do not own?

2Pacalypse-:
Yeah, the `moderateChatChannels` permission was added for specific purpose of allowing users to moderate all chat channels. The problem with `editChatChannels` permission is that it was added with no specific purpose in mind. "Editing" is a pretty ambiguous term, and my thinking is that if we need to allow server admins to perform additional things on all chat channels, then we can add more specific permissions for those actions, e.g. `removeServerAvatar`.

PR #822, opened 2022-02-19
Add a way to mention other users from user's context menu.

2Pacalypse-:


tec27:
This is the same sort of issue as the other things you tried to add to this menu, it can't be done this way. This component is used outside of chat, we don't necessarily have a chat context here. This menu item needs to be added dynamically in the places that need it, just like ban, etc.

2Pacalypse-:
Oh, I thought the mention functionality was safe to add here since we allow mentioning users in all chat contexts, but I see what you mean now with `chatContext` possibly not being defined. I'll close this PR then until I manage to figure out how exactly to make menu items contextual.

PR #823, opened 2022-02-19
Sync the server's assets (/public folder) to DO spaces.

2Pacalypse-:


server/update_server.sh
```diff
@@ -16,4 +16,21 @@ for f in server/redis-migrations/*.js; do
   node -r "./babel-register" -r "core-js/proposals/reflect-metadata" -r "dotenv/config" "$f" || exit 1
 done
 
+echo "Syncing public assets to the cloud"
```
tec27, line 19: This should probably only run if Spaces is configured, otherwise this can fail in weird ways

server/update_server.sh
```diff
@@ -16,4 +16,21 @@ for f in server/redis-migrations/*.js; do
...
+HOST_BUCKET="%(bucket)s.${HOST_BASE}"
+SPACE_NAME=$(echo $SB_FILE_STORE | jq '.doSpaces.bucket')
+
+s3cmd sync \
+  --access_key=${ACCESS_KEY} \
+  --secret_key=${SECRET_KEY} \
+  --host=${HOST_BASE} \
+  --host-bucket=${HOST_BUCKET} \
+  --recursive \
+  --delete-removed \
```
tec27, line 32: I don't think deleting removed stuff is particularly good, ideally we want static assets to be available forever but we don't really want to have to keep them all in the repo forever

server/update_server.sh
```diff
@@ -16,4 +16,21 @@ for f in server/redis-migrations/*.js; do
...
+  --skip-existing \
```
tec27, line 33: I don't know what this argument does, but it sure sounds like "don't bother doing anything with stuff that already exists, even if the content is different", which seems like not something we want. Is that the case?

tec27 (changes requested):

server/update_server.sh
2Pacalypse-: Yeah, I might have misread the description of this argument. I thought it would only skip files whose md5 checksum and size is the same since `sync` command checks those by default. But in the sync guide it says:

_Sometimes you don’t want to compare checksums and sizes of the remote vs local files and only want to upload those that are new. For that use the `—skip-existing` option._

So yeah, we probably don't want this either.

tec27:
These changes look good, think this PR should also include the other parts that we discussed though (actually using these resources for static assets in the app)

client/index.jsx
```diff
@@ -117,7 +121,10 @@ Promise.all([rootElemPromise])
 
     store.dispatch(action)
     try {
-      await sessionPromise
+      const [config] = await Promise.all([configPromise, sessionPromise])
+      const serverOrigin = getServerOrigin().toLowerCase()
+      const publicAssetsUrl = new JsonLocalStorageValue(`${serverOrigin}:publicAssetsUrl`)
```
tec27, line 126: Small thing, but I think I would store the entire config here probably, rather than just this one piece of it. Can't really imagine a config value we wouldn't want to cache

client/network/server-url.ts
```diff
@@ -27,3 +29,17 @@ export function getServerOrigin() {
...
   }
 }
+
+/**
+ * Returns an URL for a path that is supposed to be in a public assets directory. In case it's not
+ * set (should be very rare), we default to our production server CDN.
+ */
+export function makePublicAssetUrl(path: string) {
+  const serverOrigin = getServerOrigin().toLowerCase()
+  const publicAssetsUrl = new JsonLocalStorageValue<string>(`${serverOrigin}:publicAssetsUrl`)
```
tec27, line 39: The goal with this class is that you can construct it once and then use it as an accessor/writer, so you probably don't want to be constructing it on every call to this method. Make it a global variable instead (and maybe put it somewhere common that both the thing loading the config + this can use).

client/network/server-url.ts
```diff
@@ -27,3 +29,17 @@ export function getServerOrigin() {
...
+  if (publicAssetsUrl) {
```
tec27, line 40: This should handle leading slashes the same way that `makeServerUrl` does (line 17), otherwise this is fraught with issues around whether or not the user provided a leading slash

tec27: The logic here also doesn't really makes sense, `publicAssetsUrl` is always set. It seems like what you actually wanted was `if (publicAssetsUrl.getValue())`, but you should probably also retrieve that only once to avoid paying for the JSON parse cost twice

server/routes.ts
```diff
@@ -55,10 +55,23 @@ export default function applyRoutes(app: Koa, websocketServer: WebsocketServer)
...
+  let publicAssetsUrl: string
+  if (process.env.SB_FILE_STORE) {
+    const fileStoreSettings = JSON.parse(process.env.SB_FILE_STORE)
+    if (fileStoreSettings.doSpaces) {
+      const settings = fileStoreSettings.doSpaces
+      publicAssetsUrl = settings.cdnHost
+        ? `https://${settings.cdnHost}/public`
+        : `https://${settings.bucket}.${settings.endpoint}/public`
+    } else {
+      publicAssetsUrl = process.env.SB_CANONICAL_HOST!
```
tec27, line 67: imo you are better off just specifying a relative URL here (`/`)

server/routes.ts
```diff
@@ -55,10 +55,23 @@ export default function applyRoutes(app: Koa, websocketServer: WebsocketServer)
...
+    }
+  }
+
   router.get('/config', async ctx => {
-    ctx.body = {}
+    ctx.body = {
+      publicAssetsUrl,
```
tec27, line 73: I would create a common type for this config and use it everywhere, having this be `any` is extremely prone to breaking things with changes

server/routes.ts
tec27: And then I would also make the Spaces ones include a trailing slash for consistency, will make the handling on the client a bit easier

client/index.jsx
```diff
@@ -100,6 +103,7 @@ Promise.all([rootElemPromise])
     return { elem, store }
   })
   .then(async ({ elem, store }) => {
+    const configPromise = fetchJson('/config', { method: 'get' })
```
tec27, line 106: imo this should be included in `_sbInitData` in the web version, we want to avoid as many extra requests as we can

tec27 (changes requested):

server/routes.ts
tec27: Ah sorry, forgot the app wouldn't be on the right origin for that. Doing what you had (including the canonical host) is fine in that case, I think tacking it on in the client is more trouble than it's worth.

client/index.jsx
```diff
@@ -125,7 +125,9 @@ Promise.all([rootElemPromise])
       const config = await configPromise
       serverConfig.setValue(config)
     } catch (err) {
-      // TODO(2Pac): Do something with the error
+      // Ignoring the error here shouldn't be that big of a deal since the config is usually cached
```
tec27, line 128: I'd probably just log the error at least (as a warning probably)

Generally wouldn't want to swallow these errors completely as it can make it hard to spot weird issues

client/network/server-url.ts
```diff
@@ -12,6 +9,14 @@ export function makeServerUrl(path: string) {
     return path
   }
 
+  // This can't be a top-level variable since we're importing a module that's using this function
```
tec27, line 12: I very much do not understand this comment, nor the semantics that you think are at play here. If we're importing a file that is using this file, we have fucked up, and we should fix that instead of doing this. Circular dependenices are bad

client/server-config-storage.ts
```diff
@@ -0,0 +1,6 @@
+import { ServerConfig } from '../common/server-config'
+import { JsonLocalStorageValue } from './local-storage'
+import { getServerOrigin } from './network/server-url'
+
+const serverOrigin = getServerOrigin().toLowerCase()
```
tec27, line 5: So this I guess is the problematic call? I would just pull the `baseUrl` thing into its own file and import that in both files, and use that as the key for localStorage (in Electron, in the browser you don't really need a unique key for this at all since each origin will get its own localStorage, so an empty string or any arbitrary string is fine there)

tec27 (changes requested):

PR #825, opened 2022-03-07
Move the channel owners state to the channel table itself.

2Pacalypse-:
Even though it makes retrieving some information a bit harder, having
the owner state in the channels table itself just makes a lot more sense
than having it in channel_users table. Also, this ensures that there is
only one owner per channel, and makes changing user permissions a bit
safer.

### DIFF
--- a/server/lib/chat/chat-models.ts
+++ b/server/lib/chat/chat-models.ts
@@ -262,7 +262,7 @@ export async function removeUserFromChannel(
     if (deleteChannelResult.rowCount > 0) {
       // Channel was deleted; meaning there is no one left in it so there is no one to transfer the
       // ownership to
-      return { newOwnerId: undefined } as LeaveChannelResult
+      return {}
     }
 
     const currentOwnerResult = await client.query<{ owner_id: SbUserId }>(sql`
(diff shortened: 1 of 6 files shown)


server/lib/chat/chat-models.ts
```diff
@@ -247,80 +237,96 @@ export async function getMessagesForChannel(
...
         NOT EXISTS (SELECT 1 FROM channel_users WHERE channel_name = ${channelName})
-      RETURNING name`)
-    if (result.rowCount > 0) {
+      RETURNING name;
+    `)
+    if (deleteChannelResult.rowCount > 0) {
       // Channel was deleted; meaning there is no one left in it so there is no one to transfer the
       // ownership to
-      return { newOwnerId: null }
+      return { newOwnerId: undefined } as LeaveChannelResult
```
tec27, line 265: Seems like these casts would be unnecessary if you just left the key off? (e.g. `return {}`)

server/lib/chat/chat-models.ts
```diff
@@ -247,80 +237,96 @@ export async function getMessagesForChannel(
...
-      UPDATE channel_users
-      SET kick = true, ban = true, change_topic = true, toggle_private = true,
-        edit_permissions = true, owner = true
-      WHERE user_id = ${result.rows[0].user_id} AND channel_name = ${channelName}`)
-    return { newOwnerId: result.rows[0].user_id }
+      UPDATE channels
+      SET owner_id = ${earliestUserResult.rows[0].user_id}
+      WHERE name = ${channelName};
+    `)
+    return { newOwnerId: earliestUserResult.rows[0].user_id } as LeaveChannelResult
```
tec27, line 329: Very much don't understand why this would need a cast, is transact just not inferring types properly here? If so we should fix that instead of doing these casts on every return

tec27 (approved):

PR #829, opened 2022-03-20
Add race statistics tracking for ladder games.

2Pacalypse-:
So far we only tracked race statistics for all games a user has played
in aggregate. Now we're tracking them for only their ladder games as
well, partitioned by each matchmaking type.

Note that we're currently not displaying this information anywhere on
the client. That will come in a separate PR.


common/ladder/index.ts
```diff
@@ -1,17 +1,18 @@
+import { RaceStats } from '../races'
 import { SbUser, SbUserId } from '../users/sb-user'
 
 /**
  * A ranked player for a particular matchmaking ladder. Contains information about their play
  * history and current rating.
  */
-export interface LadderPlayer {
+export type LadderPlayer = {
```
tec27, line 8: Would generally always prefer using an `interface` and `extends`, type unions should be saved for usage in most cases (e.g. "I need this function to accept the combination of these 2 types") not for normal type declaration

common/users/user-stats.ts
```diff
@@ -1,25 +1,8 @@
...
-  rLosses: number
-
-  rPWins: number
-  rPLosses: number
-  rTWins: number
-  rTLosses: number
-  rZWins: number
-  rZLosses: number
-}
+export type UserStats = { userId: SbUserId } & RaceStats
```
tec27, line 8: Same as previously, use `extends`

server/lib/matchmaking/models.ts
```diff
@@ -4,11 +4,12 @@ import {
   MatchmakingResult,
   MatchmakingType,
 } from '../../../common/matchmaking'
+import { RaceStats } from '../../../common/races'
 import { SbUserId } from '../../../common/users/sb-user'
 import db, { DbClient } from '../db'
 import { Dbify } from '../db/types'
 
-export interface MatchmakingRating {
+export type MatchmakingRating = {
```
tec27, line 12: Same as previously, `extends`

server/lib/matchmaking/models.ts
```diff
@@ -162,13 +195,27 @@ export async function updateMatchmakingRating(
...
+       r_t_losses = ${mmr.rTLosses},
+       r_z_wins = ${mmr.rZWins},
+       r_z_losses = ${mmr.rZLosses}
     WHERE user_id = ${mmr.userId} AND matchmaking_type = ${mmr.matchmakingType};
   `)
 }
 
 // TODO(tec27): Remove username from this and get user data in another query
-export interface GetRankingsResult {
+export type GetRankingsResult = {
```
tec27, line 218: `extends`

server/migrations/20220316164353-add-race-stats-to-matchmaking-ratings.js
```diff
@@ -0,0 +1,180 @@
...
+    await db.runSql(`
+      UPDATE matchmaking_ratings AS s
+      SET r_${race}_wins = r.wins
+      FROM (
+        SELECT user_id, g.config->'gameSourceExtra'->>'type' as matchmaking_type, COUNT(*) as wins
+        FROM games_users gu JOIN games g
+          ON gu.game_id = g.id
+        WHERE
+          g.config->>'gameSource' = 'MATCHMAKING' AND
+          gu.assigned_race = '${race}' AND
```
tec27, line 73: I think you copied this from the previous migration, but this logic doesn't seem to be correct (and we may need to correct the previous migration as well). `assigned_race` is populated regardless of whether random was selected, and I don't see anything in this query that ensures it only applies to random, so adding e.g. `z_wins` and `r_z_wins` double-counts them

tec27 (changes requested):

server/migrations/20220316164353-add-race-stats-to-matchmaking-ratings.js
tec27: If only I had caught it when I wrote the original code :P

PR #830, opened 2022-03-24
Display the most played race in the ladder table.

2Pacalypse-:


client/ladder/ladder.tsx
```diff
@@ -239,6 +241,13 @@ const PlayerCell = styled(BaseCell)`
   align-items: center;
 `
 
+const RaceCell = styled(BaseCell)`
+  width: 48px;
+  display: flex;
+  justify-content: flex-end;
```
tec27, line 247: end alignment is reserved for numeric-ish data, I don't think races really qualify as that

client/ladder/ladder.tsx
```diff
@@ -433,13 +448,25 @@ const Row = React.memo(({ style, isEven, player, username, curTime, onSelected }
...
   const [buttonProps, rippleRef] = useButtonState({ onClick })
 
+  const raceStats = new Map<number, RaceChar>([
+    [player.pWins + player.pLosses, 'p'],
+    [player.tWins + player.tLosses, 't'],
+    [player.zWins + player.zLosses, 'z'],
+    [player.rWins + player.rLosses, 'r'],
+  ])
+  const mostPlayedGames = [...raceStats.keys()].sort((a, b) => b - a)[0]
+  const mostPlayedRace = raceStats.get(mostPlayedGames)!
```
tec27, line 458: This is too much cleverness imo, you could simplify this by sorting the tuples you're passing to the Map constructor instead (and there isn't really value in doing the Map stuff anyway, since you're passing a custom sort function anyway):

```ts
const raceStats: Array<[number, RaceChar]> = [
  [player.pWins + player.pLosses, 'p'],
  [player.tWins + player.tLosses, 't'],
  [player.zWins + player.zLosses, 'z'],
  [player.rWins + player.rLosses, 'r'],
]
raceStats.sort((a, b) => b[0] - a[0])
const mostPlayedRace = raceStats[0][1]
```

tec27 (approved):

PR #832, opened 2022-04-02
Add an ability to search for players by their name on the ladder page.

2Pacalypse-:


client/ladder/action-creators.ts
```diff
@@ -3,33 +3,52 @@ import { MatchmakingType } from '../../common/matchmaking'
...
     ) {
       return
     }
 
-    dispatch({ type: '@ladder/getRankingsBegin', payload: { matchmakingType, fetchTime } })
+    lastFetchTimeByMatchmakingType.set(matchmakingType, fetchTime)
+    lastSearchQueryByMatchmakingType.set(matchmakingType, searchQuery)
+
+    const result = await fetchJson<GetRankingsResponse>(
+      apiUrl`ladder/${matchmakingType}`.concat(searchQuery ? urlPath`?q=${searchQuery}` : ''),
```
tec27, line 35: Just use `+` (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/concat#performance)

client/ladder/action-creators.ts
```diff
@@ -3,33 +3,52 @@ import { MatchmakingType } from '../../common/matchmaking'
...
+      {
+        signal: spec.signal,
+      },
+    )
+
+    if (fetchTime < lastFetchTime) {
```
tec27, line 41: So what you're trying to do here is good, but this implementation doesn't actually accomplish it. What you want is "check the current lastFetchTime value to make sure it is the same as when we started the fetch". What you have here is "check the value of the local variable we assigned at the start of the function to make sure it hasn't changed". This function is the only owner of that variable so we can guarantee it will *never* change in this code, thus this code is dead.

What you actually want is:
```ts
if (fetchTime < lastFetchTimeByMatchmakingType.get(matchmakingType))
```

I think this would also be more readable if you reversed the condition and put the dispatch inside of that, rather than using an early return:

```ts
if (fetchTime >= lastFetchTimeByMatchmakingType.get(matchmakingType)) {
  dispatch({
    type: '@ladder/getRankings',
    payload: result,
    meta: { matchmakingType },
  })
}
```

client/ladder/action-creators.ts
```diff
@@ -3,33 +3,52 @@ import { MatchmakingType } from '../../common/matchmaking'
...
 import { ThunkAction } from '../dispatch-registry'
 import { push } from '../navigation/routing'
+import { abortableThunk, RequestHandlingSpec } from '../network/abortable-thunk'
 import { fetchJson } from '../network/fetch'
 
 const LADDER_RANKINGS_CACHE_TIME_MS = 60 * 1000
 
-export function getRankings(matchmakingType: MatchmakingType): ThunkAction {
-  return (dispatch, getState) => {
+const lastFetchTimeByMatchmakingType = new Map<MatchmakingType, Date>()
```
tec27, line 11: Using Dates for anything like this is a mistake and will not function correctly across time changes. Use a monotonic counter (`performance.now()`)

tec27 (changes requested):
Files with comments are the only ones I've really looked at so far, but the way the reducer works as of this change doesn't really seem fully baked to me. It treats the entire ladder rankings as existing only in the context of a particular query, which doesn't really make sense, and has a lot of the same issues that we have seen with the old implementation of the maps system.

imo queries should return results particular to that query and you can store those results in the reducer, but they shouldn't overwrite the *entire* rankings list. They should be stored as separate data that is used only for those searches. As you have it implemented it is very hard to deal with out of sync data, the reducer has zero indication what the data it's holding is even queried from or for

client/ladder/action-creators.ts
```diff
@@ -17,37 +18,46 @@ export function getRankings(
   spec: RequestHandlingSpec<void>,
 ): ThunkAction {
   return abortableThunk(spec, async dispatch => {
-    const fetchTime = new Date()
-    const lastFetchTime = lastFetchTimeByMatchmakingType.get(matchmakingType) ?? new Date(0)
+    // `performance.now()` can return value that's less than our cache time, so we floor it to that.
```
tec27, line 21: This isn't really a floor, if you want that then you should use `Math.max`. I don't really understand this code though, the constant is used to check for the difference between the current time and the last time. This feels like a hacky workaround to avoid dealing with the fact that the last time can be undefined, and the code would be easier to understand if you just dealt with that directly instead.

client/ladder/action-creators.ts
```diff
@@ -17,37 +18,46 @@ export function getRankings(
...
+    const lastFetchTime = lastFetchTimeByMatchmakingType.get(matchmakingType) ?? 0
+    const lastSearchTime = lastFetchTimeByMatchmakingType.get(matchmakingType) ?? 0
     const lastSearchQuery = lastSearchQueryByMatchmakingType.get(matchmakingType) ?? ''
 
-    if (
-      Number(fetchTime) - Number(lastFetchTime) < LADDER_RANKINGS_CACHE_TIME_MS &&
-      lastSearchQuery === searchQuery
-    ) {
-      return
+    if (searchQuery) {
```
tec27, line 27: `searchQuery` isn't specified as being possibly `undefined`, so this feels incredibly weird to me. imo you should just split this into 2 functions: one with a searchQuery and one without, since they are almost entirely separate code paths anyway.

They should probably also be separate redux actions

client/ladder/ladder.tsx
```diff
@@ -162,6 +172,35 @@ export function Ladder({ matchmakingType: routeType }: LadderProps) {
     }
   }, [routeType])
 
+  let content = <LoadingDotsArea />
+  if (searchQuery && searchResults) {
+    content = (
+      <LadderTable
```
tec27, line 178: imo, given how many props this has, you'd be better off just setting an object to pull `lastUpdated`, `totalCount`, and `players` from rather than setting an entire element. How you have it now is extremely prone to causing problems with future updates

tec27 (changes requested):

client/ladder/action-creators.ts
2Pacalypse-: I used "floor" here in the sense that this is the lowest value that the fetch time can be now.

The reason why I did this is because `performance.now()` returns the time since the beginning of the current document's lifetime (unlike `Date.now()` which returns time since 1971 or whatever). This means that `performance.now()` can return, for example, 20 seconds, while our cache time is 1 minute, which throws a wrench in our condition below where we assume that all requests made during initial cache time period are already cached.

I could rewrite the condition to treat any request made during initial cache time (e.g. `fetchTime < LADDER_RANKINGS_CACHE_TIME_MS`) as a way to always make the request, but then we're losing benefits of caching during that period. It's also possible that I'm overly complicating this and am missing something obvious :d

client/ladder/action-creators.ts
2Pacalypse-: Yeah, I'm basically relying here on an empty string being a falsy value, but I wasn't really happy with that myself. I also debated with myself about separating this into 2 functions and redux actions, which in hindsight I should've done ^^

client/ladder/action-creators.ts
tec27: I get why you're doing it, it's not a correct thing to do. The correct thing to do is to properly handle the case that it has not been fetched, and not use 0 as a magic value to indicate this.

client/router-hooks.ts
```diff
@@ -0,0 +1,38 @@
...
+ *   const [page, setPage] = useLocationSearchParam('page')
+ *   console.log(page) // '3'
+ *   return <button onClick={() => setPage('4')}>Go to page 4</button>
+ * }
+ */
+export const useLocationSearchParam = (
+  name: string,
+): [value: string, setValue: (value: string) => void] => {
+  const [location, setLocation] = useLocation()
+  const searchParams = new URLSearchParams(window.location.search)
```
tec27, line 23: imo this should be memoized. And if you're going to use `URLSearchParams` you may as well just use `URL` on `location` instead (which then avoids accessing this global value randomly): https://developer.mozilla.org/en-US/docs/Web/API/URL/searchParams

client/router-hooks.ts
```diff
@@ -0,0 +1,38 @@
...
+  const searchValue = searchParams.get(name) ?? ''
+
+  const setLocationSearch = (value: string) => {
+    if (value) {
```
tec27, line 27: if/when you memoize the above value you'll want to create a copy of the params here before mutating (honestly, you should have done this even when things weren't memoized imo). Think you can do that via:
```ts
const params = new URLSearchParams(Array.from(searchParams.entries()))
```

client/search/search-input.tsx
```diff
@@ -0,0 +1,81 @@
...
+`
+
+interface SearchInputProps {
+  searchQuery: string
+  onSearchChange: (value: string) => void
+  className?: string
+}
+
+export function SearchInput({ searchQuery, onSearchChange, className }: SearchInputProps) {
+  const [inputValue, setInputValue] = useState('')
```
tec27, line 24: this should be initialized to `searchQuery`, otherwise you're needlessly double-rendering every time this component starts with a query

client/search/search-input.tsx
```diff
@@ -0,0 +1,81 @@
...
+  const [inputFocused, setInputFocused] = useState(false)
+  const inputFocusedRef = useValueAsRef(inputFocused)
+  const inputRef = useRef<HTMLInputElement>(null)
+
+  useEffect(() => {
+    // In case the `searchQuery` (the value that was actually searched for) gets out of sync with
+    // the input value (which can happen if network requests return out of order for example), we
+    // update the input value to what was actually searched for so it reflects the search results
+    // accurately.
+    setInputValue(searchQuery)
```
tec27, line 34: If you initialize it properly as I said above, this shouldn't ever be necessary. What was happening in your testing wasn't "network requests returned out of order", it's that you weren't initialized the state to the actual value, so until a change event was delivered by the input it never got updated

client/search/search-input.tsx
```diff
@@ -0,0 +1,81 @@
...
+import { useKeyListener } from '../keyboard/key-listener'
+import { fastOutSlowInShort } from '../material/curves'
+import TextField from '../material/text-field'
+import { useValueAsRef } from '../state-hooks'
+
+const ESCAPE = 'Escape'
+const F = 'KeyF'
+
+const TextFieldContainer = styled(TextField)`
+  width: ${props => (props.isFocused ? '250px' : '200px')};
```
tec27, line 13: this is a weird value to bake into a component that you seemingly intend to be generic across a bunch of interfaces. Would prefer that this is a percentage of some sort, or that it is at least customizable through CSS custom properties

server/lib/matchmaking/models.ts
```diff
@@ -257,22 +257,38 @@ function fromDbGetRankingsResult(r: DbGetRankingsResult) {
...
+        r.p_wins, r.p_losses, r.t_wins, r.t_losses, r.z_wins, r.z_losses, r.r_wins, r.r_losses,
+        r.r_p_wins, r.r_p_losses, r.r_t_wins, r.r_t_losses, r.r_z_wins, r.r_z_losses,
+        r.last_played_date
       FROM ranked_matchmaking_ratings_view r JOIN users u
         ON r.user_id = u.id
       WHERE r.matchmaking_type = ${matchmakingType}
+    `
+
+    if (searchStr) {
+      const escapedStr = `%${searchStr.replace(/[_%\\]/g, '\\$&')}%`
```
tec27, line 280: imo pull the sanitization here into a separate common spot (or find a library to use for this, I'm not sure I trust us to have thought of all the possible vectors for abuse here)

tec27 (changes requested):

client/search/search-input.tsx
tec27: This absolutely should be handled, but I'm not sure why you'd be handling it here or in this way. That particular ordering should already be handled in the action creator, which only dispatches actions for the most recent request's response

client/router-hooks.ts
2Pacalypse-: Hmm, I actually tried doing that initially (and now again after completely forgetting about it :p), but `wouter` doesn't return the query parameters in their `useLocation` hook. There's an open issue about it: https://github.com/molefrog/wouter/issues/58

tec27: `useLocation` just gives you a string. I wasn't suggesting using it directly, I was telling you to use it to construct a `URL`, which will parse the string and give you parameters. I think you may have misread what I wrote, this is all I was suggesting:

```ts
const [location, setLocation] = useLocation()
const searchParams = useMemo(() => {
  const url = new URL(location)
  return location.searchParams
}, [location])
```

tec27: Ah, I clicked through to the issue and I see what you mean now. That is extremely unexpected to me and I think we should probably write a hook that handles this properly, at the very least. I think there are other parts of the codebase that have bugs now because they expect query string changes to re-render

2Pacalypse-: As far as I can see in their source code, they do take query string into account when deciding when to re-render, so we should be safe there. They just don't return the query string from the `useLocation` hook for whatever reason.

We could provide our own hook that will be used for `useLocation`, but we'd basically have to reimplement everything they already did just to add `search` variable in the return array here: https://github.com/molefrog/wouter/blob/master/use-location.js#L60

Unless I'm missing something and there's a deeper issue here.

tec27: Their implementation calls `dispatchEvent`, thus you can just listen on `window` for these events and not reimplement anything. There are examples in the linked issue of exactly how to do this

client/ladder/ladder.tsx
```diff
@@ -141,15 +218,15 @@ export function Ladder({ matchmakingType: routeType }: LadderProps) {
...
-            lastUpdated={rankings.lastUpdated}
-            totalCount={rankings.totalCount}
-            players={rankings.players}
+            {...rankingsData}
             usersById={usersById}
-            isLoading={rankings.isLoading}
-            lastError={rankings.lastError}
-            curTime={Number(rankings.fetchTime)}
+            lastError={lastError}
+            curTime={CURRENT_TIME}
```
tec27, line 226: You can't do this, you've built this such that this is the value when this file is loaded, not when the rankings are fetched. It won't ever change. You should probably just add this fetchTime value to both sides of the reducer, like it was previously

client/ladder/ladder.tsx
```diff
@@ -342,7 +428,7 @@ export function LadderTable(props: LadderTableProps) {
 
   const renderRow = useCallback(
     (index: number) => {
-      const player = playersRef.current?.get(index - 1)
+      const player = playersRef.current?.[index]
```
tec27, line 431: huh, didn't know you could do this with that syntax, weird. (You can also do it with function calls, in case you didn't know)

client/ladder/ladder.tsx
```diff
@@ -379,7 +472,7 @@ export function LadderTable(props: LadderTableProps) {
             TableRow,
             FillerRow,
           }}
-          totalCount={props.totalCount}
+          data={playersRef.current}
```
tec27, line 475: Prefer the actual value here, I see no reason to use the ref here

client/search/search-input.tsx
```diff
@@ -0,0 +1,95 @@
...
+import { useKeyListener } from '../keyboard/key-listener'
+import { fastOutSlowInShort } from '../material/curves'
+import TextField from '../material/text-field'
+import { useValueAsRef } from '../state-hooks'
+
+const ESCAPE = 'Escape'
+const F = 'KeyF'
+
+const TextFieldContainer = styled(TextField)`
+  ${props =>
```
tec27, line 13: Always prefer using the variable only on the value side of the CSS *unless it cannot be avoided*. So this should be:

```css
width: ${props.isFocused ? 'var(--sb-search-input-focused-width, 250px)' : 'var(--sb-search-input-width, 200px)' };
```

client/search/search-input.tsx
```diff
@@ -0,0 +1,95 @@
...
+          return true
+        }
+
+        return false
+      },
+    })
+
+    return (
+      <TextFieldContainer
+        clasName={className}
```
tec27, line 81: You typo'd this prop name, I don't know why typescript didn't fail this but probably we should look into that as well

client/router-hooks.ts
```diff
@@ -0,0 +1,66 @@
+import { useCallback, useMemo, useSyncExternalStore } from 'react'
```
tec27, line 1: Put this file in `client/navigation/` imo

tec27 (changes requested):

client/search/search-input.tsx
2Pacalypse-: Probably because `TextField` is still in JavaScript :d

PR #833, opened 2022-04-10
Implement dialog stacking.

2Pacalypse-:
This makes it possible to open one dialog from another, without losing
the context of first dialog. This should be used very sparingly, since
opening dialogs from dialogs can make for a very confusing user
experience.

This PR also does the following:
- Moves dialogs reducer to immer
- Types the payload of opening dialogs, along with its `initData`
- Moves dialog opening/closing animation to react-spring

### DIFF
--- a/client/dialogs/connected-dialog-overlay.tsx
+++ b/client/dialogs/connected-dialog-overlay.tsx
@@ -52,6 +52,8 @@ const Scrim = styled(animated.div)`
 const INVISIBLE_SCRIM_COLOR = rgba(dialogScrim, 0)
 const VISIBLE_SCRIM_COLOR = rgba(dialogScrim, 0.42)
 
+const noop = () => {}
+
 export interface DialogContextValue {
   styles: SpringValues
 }
@@ -124,9 +126,9 @@ export function ConnectedDialogOverlay() {
 
   const isTopDialogModal = topDialog ? getDialog(topDialog.type, starcraft).modal : false
   const onCancel = useCallback(
-    (modal: boolean, event?: React.MouseEvent) => {
-      if (!modal && (!event || !isHandledDismissalEvent(event.nativeEvent))) {
-        dispatch(closeDialog())
+    (dialogType: DialogType | 'all', event?: React.MouseEvent) => {
+      if (!event || !isHandledDismissalEvent(event.nativeEvent)) {
+        dispatch(closeDialog(dialogType))
       }
     },
     [dispatch],
@@ -165,7 +167,14 @@ export function ConnectedDialogOverlay() {
     <>
       {scrimTransition(
         (styles, open) =>
-          open && <Scrim style={styles} onClick={event => onCancel(isTopDialogModal, event)} />,
+          open && (
+            <Scrim
+              style={styles}
+              onClick={event =>
+                isTopDialogModal ? noop() : onCancel(topDialog?.type ?? 'all', event)
+              }
+            />
+          ),
       )}
       {dialogTransition((styles, dialogState) => {
         const { component: DialogComponent, modal } = getDialog(dialogState.type, starcraft)
@@ -179,7 +188,9 @@ export function ConnectedDialogOverlay() {
               <DialogContext.Provider value={{ styles }}>
                 <DialogComponent
                   dialogRef={dialogRef}
-                  onCancel={(event: React.MouseEvent) => onCancel(modal, event)}
+                  onCancel={(event: React.MouseEvent) =>
+                    modal ? noop() : onCancel(dialogState.type, event)
+                  }
                   {...dialogState.initData}
                 />
               </DialogContext.Provider>
--- a/client/dialogs/connected-dialog-overlay.tsx
+++ b/client/dialogs/connected-dialog-overlay.tsx
@@ -1,8 +1,7 @@
 import { rgba } from 'polished'
 import React, { useCallback, useRef } from 'react'
 import ReactDOM from 'react-dom'
-import { animated, useTransition } from 'react-spring'
-import { CSSTransition, TransitionGroup } from 'react-transition-group'
+import { animated, SpringValues, useTransition, UseTransitionProps } from 'react-spring'
 import styled from 'styled-components'
 import { assertUnreachable } from '../../common/assert-unreachable'
 import EditAccount from '../auth/edit-account'
@@ -34,6 +33,7 @@ import StarcraftHealthCheckupDialog from '../starcraft/starcraft-health'
 import { dialogScrim } from '../styles/colors'
 import CreateWhisperSessionDialog from '../whispers/create-whisper'
 import { closeDialog } from './action-creators'
+import { DialogState } from './dialog-reducer'
 import { DialogType } from './dialog-type'
 import { SimpleDialog } from './simple-dialog'
 
@@ -52,14 +52,12 @@ const Scrim = styled(animated.div)`
 const INVISIBLE_SCRIM_COLOR = rgba(dialogScrim, 0)
 const VISIBLE_SCRIM_COLOR = rgba(dialogScrim, 0.42)
 
-const transitionNames = {
-  appear: 'enter',
-  appearActive: 'enterActive',
-  enter: 'enter',
-  enterActive: 'enterActive',
-  exit: 'exit',
-  exitActive: 'exitActive',
+export interface DialogContextValue {
+  styles: SpringValues
 }
+export const DialogContext = React.createContext<DialogContextValue>({
+  styles: {},
+})
 
 function getDialog(
   dialogType: DialogType,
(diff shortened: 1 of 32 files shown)


client/dialogs/connected-dialog-overlay.tsx
```diff
@@ -115,32 +113,30 @@ function getDialog(
...
-  const { dialogType } = dialogState
-  const { component: DialogComponent, modal } = dialogType
-    ? getDialog(dialogType, starcraft)
-    : { component: null, modal: false }
-
+  const isTopDialogModal = topDialog ? getDialog(topDialog.type, starcraft).modal : false
   const onCancel = useCallback(
-    (event?: React.MouseEvent) => {
-      if (dialogType && !modal && (!event || !isHandledDismissalEvent(event.nativeEvent))) {
+    (modal: boolean, event?: React.MouseEvent) => {
```
tec27, line 127: This seems like a strictly worse way to structure this, I think you'd be better off adding `modal` to the dependencies array and just passing this function to the `DialogComponent` directly. There's extremely little reason to `useCallback` this function otherwise, since you're creating a new instance on every render to call it.

You'll have to move the `getDialog` call up here but I don't see anything that would prevent doing that

client/dialogs/dialog-reducer.ts
```diff
@@ -1,23 +1,28 @@
...
-      initData: Map(Object.entries(action.payload.initData)),
-    })
+    const { type, initData } = action.payload
+
+    state.history.push({ type, initData })
   },
 
   ['@dialogs/close'](state, action) {
-    return new DialogState()
+    state.history.pop()
```
tec27, line 26: I feel like if we're going to allow stacking, the `close` calls should probably be more specific to a particular dialog type, or this gets pretty dangerous. I'd add a dialog type parameter to this + allow them to specify `all` as well. Specifying a particular type will pop the stack until it removes a dialog of that type (or do nothing if no dialog of that type was open?). Specifying `all` clears the stack completely.

client/dialogs/dialog-type.ts
```diff
@@ -20,3 +20,125 @@ export enum DialogType {
   TermsOfService = 'termsOfService',
   Whispers = 'whispers',
 }
+
+export interface BaseDialogPayload {
+  type: DialogType
+  initData?: Record<string, unknown>
```
tec27, line 26: honestly feels weird to have this, I think you could just not include this an a type union would work out fine? And once you remove this I'm not sure having this base type is really worthwhile either, just feels like a lot of extra code boilerplate for something that will be enforced by the code using it anyway

client/maps/map-preview.tsx
```diff
@@ -7,7 +7,9 @@ import { background400 } from '../styles/colors'
 import { batchGetMapInfo } from './action-creators'
 import MapImage from './map-image'
 
-const StyledDialog = styled(Dialog)`
+const StyledDialog = styled(Dialog)<{ $mapWidth?: number; $mapHeight?: number }>`
+  --sb-map-width: ${props => props.$mapWidth};
```
tec27, line 11: This is not an equivalent change. When we use CSS variables in `style` attributes it's to prevent re-rendering the style every time they change. If you put them into the CSS in this way, you've reverted to adding a new class to the body for every new value, and thus CSS variables are not very useful. Please revert

tec27 (changes requested):

client/dialogs/dialog-type.ts
2Pacalypse-: Hmm, I'm not sure if I'm understanding TypeScript correctly, but I thought that we need _something_ to discriminate `initData` on for different Dialog types. And I thought my only options to do that were to either add `dialogType` property to the `initData` itself, or to type the whole payload like this since it already should have dialog type.

client/dialogs/connected-dialog-overlay.tsx
2Pacalypse-: I think I tried doing that initially, but not sure how that would work exactly. Unless you mean I should call `getDialog` for all currently open dialogs here to get their `modal` status, and then generate `onCancel` for each one? But then I would need to create `onCancel` in a loop or something, which I don't think can be done with hooks. Maybe I could use `useMemo`?

client/maps/map-preview.tsx
2Pacalypse-: Yeah, I thought about that. But the problem I have now is that we need to apply `styles` generated by the `react-spring` to the dialog as well. I guess it should be possible to combine those styles with these ones, but I was too lazy to figure it out :d. I'll take a second look.

client/dialogs/dialog-type.ts
tec27: The only reason I can see this base type being useful is for use in a reducer that has no knowledge of the underlying dialog type. I didn't think about that before, so I guess it can stay if that's your reason for using it. (But no, this is not a necessary thing to have for TS to figure out what the type of `initData` is, TS will use whatever unique properties exist in the types in the union to figure out which specific type you are referring to)

If it is not possible to have any other fields (and that seems to be the case, to me), then you'd probably be better off declaring this as a generic type with 2 parameters: a DialogType and an `initData` type, and just using that for all the payloads. That would at least be properly restrictive.

client/maps/map-preview.tsx
tec27: It's just an object, you can spread them both into a single object

client/dialogs/connected-dialog-overlay.tsx
tec27: If the only difference is the `modal` flag then a different option would be to generate a modal and a non-modal handler always, and pick between them at render time. Since the modal handling is always to just ignore the cancel, you could really just make that version a constant no-op function outside the component.

client/dialogs/dialog-reducer.ts
```diff
@@ -23,6 +23,18 @@ export default immerKeyedReducer(DEFAULT_DIALOG_HISTORY_STATE, {
...
+      state.history = []
+      return
+    }
+
+    const dialogIndex = state.history.findIndex(h => h.type === dialogType)
+    if (dialogIndex < 0) {
+      return
+    }
+
+    state.history.splice(dialogIndex)
```
tec27, line 38: No, I said when you close a dialog of a particular type, you close *all the dialogs up to that point*. What you've done here is *only close that particular dialog*. So if we have a settings dialog and it opens a "find your starcraft path dialog", then something happens that closes settings, your code here will leave the starcraft path dialog up despite its owner being closed. Seems not great, please implement as specified

client/dialogs/dialog-reducer.ts
```diff
@@ -23,6 +23,18 @@ export default immerKeyedReducer(DEFAULT_DIALOG_HISTORY_STATE, {
...
   ['@dialogs/close'](state, action) {
-    state.history.pop()
+    const { dialogType } = action.payload
+
+    if (dialogType === 'all') {
+      state.history = []
+      return
+    }
+
+    const dialogIndex = state.history.findIndex(h => h.type === dialogType)
```
tec27, line 33: `findIndex` finds the first index that matches, but the way you're stacking dialogs is in ascending order of time opened, that is, the oldest dialogs are first. When a dialog of a particular type is closed, we should be closing the *newest* dialog of that type, not the *oldest*. You want `findLastIndex`, which doesn't exist, so you'll need to implement this behavior yourself.

tec27 (changes requested):

client/dialogs/dialog-reducer.ts
2Pacalypse-: Oh, I thought the `Array.splice()` method would remove _all_ the elements from the given index to the end of array if the second argument is omitted. Or was that `Array.slice()`, I always confuse those two :d

tec27: Yeah, `splice` is "remove some number of elements, optionally inserting new ones in their place". `slice` also wouldn't be what you want here (at least with a single argument), since that gives you a new array with the elements *starting* at that position.

2Pacalypse-: Wait, I'm a bit confused then. Isn't `splice` what we want then? If we have a history of [Dialog1, Dialog2, Dialog3] and something closes Dialog2, we'll be left with [Dialog1].

tec27: No, you probably want to do `slice(0, index)`

2Pacalypse-: I'm still a bit confused I'm afraid. If we take the following hypothetical scenario:
- User opens the settings dialog
- User opens the "find your starcraft path" dialog within it

The `state.history` array should be [Dialog.Settings, Dialog.FindStarCraftPath]. Then if something closes the `Dialog.Settings`, `state.history.splice(index)` will basically be translated to `state.history.splice(0)`, which will remove all elements from the array and `state.history` should be equal to `[]`, which is what we want. I feel like I'm missing something rather obvious, but I'm just not seeing it atm ^^

common/arrays.test.ts
```diff
@@ -0,0 +1,14 @@
+import { findLastIndex } from './arrays'
+
+const makeCallbackFn = (searchValue: number) => (value: number) => value === searchValue
+
+describe('common/arrays', () => {
+  test('findLastIndex', () => {
+    expect(findLastIndex([], makeCallbackFn(1))).toBe(-1)
```
tec27, line 7: a useful test to have would be with an array that contains 2 of the same value, to ensure it's actually picking the last index and not just any matching one

tec27 (approved):

PR #834, opened 2022-04-18
Move the FileBrowser to TypeScript and functional components.

2Pacalypse-:
This PR also allows us to finally get rid of the react-virtualized.
There are multiple bugs that were introduced in the last couple of years
that I fixed here, either automatically by moving to hooks, or
explicitly. Some of those are:
- File browser displaying phantom folder due to trailing slash after the
  root folder
- "Up one directory" entry showing at the root folder sometimes
- Uploading a map doesn't redirect you to the "My Maps" tab where the
  uploaded map is visible (I already had someone asking me why map
  upload doesn't work, but now that I think about it, they were most
  likely in the "Official Maps" tab when uploading and then they didn't
  see the uploaded map)


client/file-browser/file-browser-types.ts
```diff
@@ -0,0 +1,48 @@
...
+}
+
+export interface FileBrowserRootFolder {
+  id: FileBrowserRootFolderId
+  name: string
+  path: string
+}
+
+export interface FileBrowserEntryConfig {
+  icon: React.ReactElement
```
tec27, line 46: seems like `React` needs to be imported here? not sure why it'd be globally available...

tec27 (approved):

PR #837, opened 2022-05-05
Add an option to all text fields to have a button to clear text.

2Pacalypse-:
And use it in search input.
This PR also moves TextField and related components to TS and
functional components.


client/messaging/message-input.tsx
```diff
@@ -108,11 +108,10 @@ export const MessageInput = React.forwardRef<MessageInputHandle, MessageInputPro
         className={props.className}
         label='Send a message'
         value={message}
-        maxLength={500}
         floatingLabel={false}
         allowErrors={false}
         showDivider={props.showDivider}
-        inputProps={{ autoComplete: 'off' }}
+        inputProps={{ autoComplete: 'off', maxLength: 500 }}
```
tec27, line 114: Honestly I'm not really a fan of adding this, even to stuff that has a max length. I prefer a validation that doesn't actively limit the amount of text that can be put into the field. If you e.g. paste text longer than 500 characters into this field with this attribute, it will be abruptly cut off at that point. If you paste it into a field that merely validates that length, the full text will get pasted and you have the opportunity to edit it to get it below the length, instead of just having it truncated at an arbitrary spot

client/search/search-input.tsx
```diff
@@ -63,7 +44,7 @@ export const SearchInput = React.forwardRef<SearchInputHandle, SearchInputProps>
           inputRef.current?.focus()
           inputRef.current?.select()
           return true
-        } else if (event.code === ESCAPE && inputFocusedRef.current) {
+        } else if (event.code === ESCAPE) {
```
tec27, line 47: Doesn't this make all escape presses get eaten here, regardless of where their focus is on the page? I'm not sure that's really a great thing to be doing

client/search/search-input.tsx
```diff
@@ -1,54 +1,35 @@
...
 const ESCAPE = 'Escape'
 const F = 'KeyF'
 
 const TextFieldContainer = styled(TextField)`
-  width: ${props =>
-    props.isFocused
-      ? 'var(--sb-search-input-focused-width, 250px)'
-      : 'var(--sb-search-input-width, 200px)'};
-
+  width: var(--sb-search-input-width, 200px);
```
tec27, line 13: What I was saying in Discord was, you don't need to have these custom CSS props at all any more, just put this styling in the places that are using this thing and use the static width value there. (I don't honestly know that this component needs to exist at all either, and having a generic component that exists to eat particular keypresses on the page feels a bit odd to me. That feels like it's more the concern of the page itself and would be better handled at that level, but, up to you I guess)

tec27 (changes requested):

client/search/search-input.tsx
tec27: And while we're here, 250 is not divisible by 4 and thus isn't a valid width, you probably want 256 or something

client/material/text-field.tsx
```diff
@@ -0,0 +1,416 @@
...
+`
+
+const ClearButton = styled(IconButton)`
+  width: 32px;
+  min-height: 32px;
+`
+
+export type TextSelectionDirection = 'forward' | 'backward' | 'none'
+
+export interface TextFieldHandle {
```
tec27, line 138: This should most definitely not exist, just pass back a ref to the input. Otherwise we have to forward every single method/value/whatever that we think is vaguely useful. The only thing in here that isn't a direct forward is `clear`, and `clear` makes no sense to have here (this component is controlled, if you want to clear it, you should be setting the new value via props)

tec27 (changes requested):

client/messaging/message-input.tsx
2Pacalypse-: Yeah, that makes sense. I'll just remove the length limit on the client for now, but will add a proper validation in a later PR.

client/search/search-input.tsx
2Pacalypse-: Yeah, my initial thinking was to have this component to encapsulate keyboard handling (Ctrl + F to focus search, Escape to unfocus) and potentially show a spinner somewhere next to it. But without a spinner, all this component does is handle these two hotkeys then. I could stop eating the Escape hotkey by tracking the focus state again as I was doing and only handle it if the input is focused. But that still leaves the Ctrl + F hotkey which, as you said, is probably the concern of the page itself since there could be more than one search inputs on a page. Hmm.

I'll remove it :d

client/material/text-field.tsx
```diff
@@ -0,0 +1,373 @@
...
+      })
+
+    const clearAndFocusInput = useStableCallback(() => {
+      if (!inputRef.current) {
+        return
+      }
+
+      // React overrides the input value setter so we have to do shenanigans to set a value
+      // manually. See this answer for more info: https://stackoverflow.com/a/46012210/398302
+      const nativeInputValue = Object.getOwnPropertyDescriptor(
```
tec27, line 288: I still think this is a quite odd thing to be doing at all. This input is controlled, i.e. its value should always be driven by props. You're doing things exactly backwards here. The more sensible approach is to deliver a change event to the normal handler (that is, just call `onChange`) saying that the value is now empty, and then focus the input, letting the parent change the props as needed.

client/material/text-field.tsx
```diff
@@ -0,0 +1,373 @@
...
+  onEnterKeyDown?: (event: React.KeyboardEvent<HTMLInputElement>) => void
+  onFocus?: (event: React.FocusEvent<HTMLInputElement>) => void
+  onKeyDown?: (event: React.KeyboardEvent<HTMLInputElement>) => void
+  rows?: number
+  trailingIcons?: React.ReactElement[]
+  type?: string
+  value: string
+}
+
+// A Material text field component with single-line, multi-line and text area variants, supporting
```
tec27, line 164: Use JSDoc-style comments

client/material/text-field.tsx
```diff
@@ -0,0 +1,373 @@
...
+          $dense={dense}
+          $multiline={multiline}>
+          <ClearTooltip text='Clear'>
+            <ClearButton icon={<ClearIcon />} onClick={clearAndFocusInput} />
+          </ClearTooltip>
+        </TrailingIcon>,
+      )
+    }
+
+    const renderLabel = () => {
```
tec27, line 312: would prefer that this is just assigned to a variable rather than creating a function that we call and immediately throw away every time

tec27 (approved):

client/material/text-field.tsx
2Pacalypse-: That's what I initially tried to do (just call the `onChange`), but then I realized the `event` parameter would be wrong, so I would have to change the signature of the `onChange` handler. Initially I tried making it so it accepts only a `value`, but then I saw that we're using some additional properties from the `change` event in forms (i.e., name), and then the only thing left was to make the parameter some weird thing. In the end I decided to just imperatively trigger the `change` event, since this functionality is internal to the TextField component itself.

PR #838, opened 2022-05-10
Make the user context menu, well... contextual.

2Pacalypse-:
So this is probably like my third attempt at achieving this. My thinking
in this PR was to make the following:
- Make a `UserContextMenuProvider` component which wraps the whole main
  application (basically everywhere we might want to display a user
  context menu), and holds all the state needed to handle and display
  user context menu.
- This means that the `UserContextMenu` component holds the menu items
  for each user, and exposes methods to modify them; for now, 3 methods
  are exposed, one to replace all of the menu items (useful for
  initializing the context menu), one to append menu items, and one to
  prepend).
- Additionally, `UserContextMenuProvider` holds the status of the menu
  itself (whether it's open, and a method to close it). This allows us
  to use it in places that might not have this information readily
  available, e.g. text messages.

The way I imagined this working is by using our current
`ConnectedUserContextMenu` component to "initialize" the provider with
common menu items, as well as keep the menu status updated. And then, in
all the other places, we should have all the information to use this
context to dynamically append and/or prepend menu items to the common
ones.

However, I'm having trouble with the timings of the effects that are
run, mainly the first two; one that updates the menu status, and one
that sets the common menu items that runs right after. The menu is still
technically closed when we attempt to add common items, so they're never
added.

### DIFF
--- a/client/app.jsx
+++ b/client/app.jsx
@@ -22,6 +22,7 @@ import {
   PrivacyPolicyPage,
   TermsOfServicePage,
 } from './policies/policy-displays'
+import { UserContextMenuProvider } from './profile/user-context-menu'
 import LoadingIndicator from './progress/dots'
 import { useAppSelector } from './redux-hooks'
 import { RootErrorBoundary } from './root-error-boundary'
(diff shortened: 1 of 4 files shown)


2Pacalypse-:
Closed in favour of https://github.com/ShieldBattery/ShieldBattery/pull/844

PR #839, opened 2022-05-21
Fix the Starcraft Path dialog closing issues now that the dialogs stack.

2Pacalypse-:
For some reason it didn't cross my mind to use the only place where we
actually have dialog stacking to test it when I added dialog history,
and instead I tested it by creating a dummy dialog stacking in Join
Channel dialog just for testing purposes. Oops.

### DIFF
--- a/client/dialogs/action-creators.ts
+++ b/client/dialogs/action-creators.ts
@@ -1,11 +1,26 @@
 import React from 'react'
+import { ThunkAction } from '../dispatch-registry'
+import { isStarcraftHealthy } from '../starcraft/is-starcraft-healthy'
 import { CloseDialog, OpenDialog } from './actions'
 import { DialogPayload, DialogType } from './dialog-type'
 
-export function openDialog(payload: DialogPayload): OpenDialog {
-  return {
-    type: '@dialogs/open',
-    payload,
+export function openDialog(payload: DialogPayload): ThunkAction {
+  return (dispatch, getState) => {
+    const { starcraft } = getState()
+
+    if (payload.type === DialogType.Settings && !isStarcraftHealthy({ starcraft })) {
+      dispatch({
+        type: '@dialogs/open',
+        payload: {
+          type: DialogType.StarcraftPath,
+        },
+      })
+    } else {
+      dispatch({
+        type: '@dialogs/open',
+        payload,
+      })
+    }
   }
 }
 
@@ -14,10 +29,13 @@ export function openSimpleDialog(
   simpleContent: React.ReactNode,
   hasButton = true,
 ): OpenDialog {
-  return openDialog({
-    type: DialogType.Simple,
-    initData: { simpleTitle, simpleContent, hasButton },
-  })
+  return {
+    type: '@dialogs/open',
+    payload: {
+      type: DialogType.Simple,
+      initData: { simpleTitle, simpleContent, hasButton },
+    },
+  }
 }
 
 export function closeDialog(dialogType: DialogType | 'all'): CloseDialog {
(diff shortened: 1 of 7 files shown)


2Pacalypse-:
One issue that I noticed now with dialog stacking is that the dialogs behind the one on top are still clickable. But I'm not sure if we can address this unless we make the `z-index` of dialog scrim the same as the `z-index` of dialog itself, and just rely on position in DOM to have dialogs on top of scrim.

tec27:
That could be fixed with CSS alone (apply `pointer-events: none` to any dialog that isn't the top one). Putting that on some base container should apply it to all descendants as well

client/dialogs/action-creators.ts
```diff
@@ -1,11 +1,26 @@
...
 
-export function openDialog(payload: DialogPayload): OpenDialog {
-  return {
-    type: '@dialogs/open',
-    payload,
+export function openDialog(payload: DialogPayload): ThunkAction {
+  return (dispatch, getState) => {
+    const { starcraft } = getState()
+
+    if (payload.type === DialogType.Settings && !isStarcraftHealthy({ starcraft })) {
```
tec27, line 11: This is not a good change imo. This action creator should never be opening *different* dialogs than what you asked it to, and doing this here means we will almost certainly end up building a ton of special behavior into this function in the future so, please don't

tec27: A cleaner solution is to write an action creator in the settings package to open the correct dialog, and use that wherever we open the settings. (I also generally think this approach is unnecessary now since we only support a single StarCraft version and the settings available are therefore always the same, but that's probably a larger change)

tec27 (changes requested):

client/dialogs/action-creators.ts
tec27: Another alternative (that would be more similar to what was in the code previously) would be to make these 2 dialogs part of the same dialog, and just pick which one to render based on state at the time. This is probably also a larger change though, since you need to track when people specifically request to change the path as well

2Pacalypse-:
I disabled the pointer-events on the non-top dialogs, but it still looks kind of bad :d

image

When the scrim is displayed over the non-top dialogs it looks kind of better imo (but it requires dialog scrim z-index to be the same as dialog z-index)

image

tec27:
Yeah, I think having the scrim is definitely necessary. I guess an alternative approach would just be to hide the background dialogs entirely (with `visibility: none` probably)? Not sure which would be preferable, maybe not great to hide the background dialog without an animation

I don't see any particular issues with having the scrim be the same z-index as the dialog itself though, since the same component controls both their DOM ordering is pretty guaranteed

client/dialogs/connected-dialog-overlay.tsx
```diff
@@ -172,16 +172,18 @@ export function ConnectedDialogOverlay() {
...
 
         // Dialog content implementations should focus *something* when mounted, so that our focus
         // traps have the proper effect of keeping focus in the dialog
         return (
           <>
             <span tabIndex={0} onFocus={onFocusTrap} />
             <span ref={focusableRef} tabIndex={-1}>
-              <DialogContext.Provider value={{ styles }}>
+              <DialogContext.Provider
+                value={{ styles: { ...styles, pointerEvents: isTopDialog ? 'auto' : 'none' } }}>
```
tec27, line 186: Probably better to prefer `undefined` over `'auto'` here

tec27 (approved):

2Pacalypse-:
>I don't see any particular issues with having the scrim be the same z-index as the dialog itself though, since the same component controls both their DOM ordering is pretty guaranteed

Yeah, this was my thinking as well. I left them as two separate variables, but with the same value for now.

PR #841, opened 2022-05-28
Refactor file-browser to not use redux.

2Pacalypse-:
It's unlikely we would've ever used the file-browser state outside the
file-browser itself, so it was not worth the complexity and boilerplate
of redux. This simplifies the file-browser quite a bit, which is always
a good thing. Now it's actually pleasurable to work on this component,
hooray!

A few additional changes in this PR:
  - Separate the file-browser breadcrumbs and entries into their own
    files
  - Remove a leftover "externalError" message that was not used anymore
  - Add props to allow any file-browser type (e.g. maps, replays) to
    sort the folders and files in any way they want, defaulting to
    sorting both by name. Note that this is different than what we had
    so far, where we sorted the replay files by date. I feel like that
    caused more confusion than it helped and I'm planning on adding a
    way for users to specify how they want their files sorted anyway
    soon(tm).


client/file-browser/file-browser.tsx
```diff
@@ -600,59 +406,55 @@ export function FileBrowser({
...
   })
 
-  const noRowsRenderer = useCallback(() => {
-    if (isLoadingFiles) {
-      return <LoadingDotsArea />
-    } else if (loadFilesError) {
-      return <ErrorText>{loadFilesError.message}</ErrorText>
-    } else {
-      return <EmptyText>Nothing to see here</EmptyText>
+  const renderRow = useStableCallback((index: number) => {
```
tec27, line 409: Does react-virtuoso call this during render? If so, it won't be safe to use this function (we have no way of enforcing this, but React's implementation will)

tec27 (approved):

client/file-browser/file-browser.tsx
tec27: I added some logging where we do this in Ladder and it *seems* like the layout effect for the callback always runs before the function is called, so I think this is okay

PR #843, opened 2022-06-16
Refactor the menu component to not render the popover part.

2Pacalypse-:
Now the menu component only renders its contents and it's left to the
users of this component to render it inside a popover, if they wish so.
The reason for this separation is so we can have menus that need to
run certain hooks (e.g. connect to a store) to only do so if they're open.

This commit also does the following:
  - change the handling of the selectable menu items
  - fix the scrollable menu (the menu items would shrink and not show
    the scrollbar)
  - fix the focused state of the menu items by using the button's
    native focus state instead of faking it
  - remove the TAB hotkey (not sure what this hotkey should be doing in
    the first place)
  - add the SPACE hotkey to click the menu item (along with ENTER)
  - stop scrolling the menu to the selected item when first opened; i'm
    not sure if this was ever a desired behaviour and feels like it
    would be hard to make work if we ever allow multiple selections
  - a few other changes (like moving to a functional component for
    menus, etc.)

### DIFF
--- a/client/material/menu/menu.tsx
+++ b/client/material/menu/menu.tsx
@@ -1,7 +1,7 @@
-import PropTypes from 'prop-types'
-import React from 'react'
+import React, { useMemo, useRef, useState } from 'react'
 import styled from 'styled-components'
-import KeyListener from '../../keyboard/key-listener'
+import { useKeyListener } from '../../keyboard/key-listener'
+import { useStableCallback } from '../../state-hooks'
 import { CardLayer } from '../../styles/colors'
 import { body1, subtitle1 } from '../../styles/typography'
 import { Popover, PopoverProps } from '../popover'
@@ -10,8 +10,7 @@ import { isMenuItem } from './menu-item-symbol'
 
 const ENTER = 'Enter'
 const ENTER_NUMPAD = 'NumpadEnter'
-const TAB = 'Tab'
-const ESCAPE = 'Escape'
+const SPACE = 'Space'
 const UP = 'ArrowUp'
 const DOWN = 'ArrowDown'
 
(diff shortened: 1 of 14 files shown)


client/material/menu/menu.tsx
```diff
@@ -53,175 +52,122 @@ export const Overlay = styled(CardLayer)<{ $dense?: boolean }>`
...
+}
+
+/**
+ * A wrapper component around Popover that can be used to quickly write menus.
+ *
+ * This component just renders the `Popover` and expects it to only render the `MenuList` when the
+ * menu is actually open. This makes sure that all the logic of the menu (e.g. keyboard handlers) is
+ * only run when the menu is open.
+ */
+export function Menu({ popoverProps, ...menuListProps }: MenuProps) {
```
tec27, line 167: I really do not see the value in this existing, it just adds a component that can almost never be skipped during rendering (because its props contain an object whose reference will nearly always change), and doesn't do anything better than components just using these two components themselves

client/material/select/select.tsx
```diff
@@ -244,25 +245,27 @@ export const Select = React.forwardRef<SelectRef, SelectProps>(
...
     )
 
-    const [displayValue, selectedIndex] = useMemo(() => {
-      if (value === undefined) {
-        return [undefined, undefined]
-      }
+    const [displayValue, options] = useMemo(() => {
+      let displayText: string | undefined
+      const options = React.Children.map(children, (child, index) => {
+        if (!isSelectedMenuItem(child)) return child
```
tec27, line 251: What exactly is the point of `SelectedMenuItem`? Is it that the item always appears as selected? Is it that item *could* be selected? I'm really not sure because the naming implies the former, but this code implies the latter

tec27 (changes requested):
I'm not sure you fully understood how this worked previously and why it needed to be changed. We already didn't apply keyboard handlers, render DOM, etc. unless the menu was open.

The reason we need `MenuList` is because we need to be able to write components that *just* render the content part (not the popover part) so that we can run certain hooks only when the menu is open.

Any places that render a menu whose contents depends on hooks (e.g. redux selectors) should be changed here to split that code into a component that returns a `MenuList`.

Before:

```ts
function MyPage() {
  const data = useAppSelector(s => /* ... */) // always runs, doesn't matter if menu is open
  return <Menu>/* ... do something with data ... */</Menu>
}
```

After:
```ts
function MyPageMenu() {
  const data = useAppSelector(s => /* ... */)  // only runs when menu is open
  return <MenuList>/* ... do something with data ... */</MenuList>
}

function MyPage() {
  return <Popover><MyPageMenu /></Popover>
}
```

tec27 (approved):

PR #844, opened 2022-06-19
Add a way to dynamically add/modify menu items on UserContextMenu. 

2Pacalypse-:
And utilize it by adding mention menu items at various places.


client/messaging/mention-hooks.tsx
```diff
@@ -0,0 +1,61 @@
...
+      onMenuClose()
+    },
+    [chatContext],
+  )
+  return useCallback(
+    (userId: SbUserId, items: React.ReactNode[], onMenuClose: (event?: MouseEvent) => void) => {
+      // TODO(2Pac): Make the `items` a map of "menu item category" -> "items" so we can add this
+      // action to the end of a "general" category, but before the "party" category. Or something
+      // like that.
+      items.push(
```
tec27, line 50: I think this works *okay* for this commit but I don't think this works with even 1 more level of this work (e.g. kick/ban). You'd be better off making a function that calls through to something on `ChatContext` to add extra menu items imo. Would make it a lot easier to add certain items only in specific contexts (a channel vs a lobby), and prevent needing to write a new hook for every new menu item

tec27 (approved):

common/js/maps.ts
```diff
@@ -0,0 +1,37 @@
+// Utility functions for use on Maps
+
+/**
+ * Takes a map and appends a new item to an array at a given key. If the key doesn't exist, a new
+ * array with the item is created and inserted at the key.
+ *
+ * Mutates the input map instead of returning a new map.
+ */
+export function appendToMapArray<K, T>(map: Map<K, T[]>, key: K, item: T) {
```
tec27, line 9: nit: to me, "MapArray" implies "an array of Maps". I think a better term for this is "MultiMap" (that's how C++ and Java refer to it)

common/js/maps.ts
```diff
@@ -0,0 +1,37 @@
+// Utility functions for use on Maps
```
tec27, line 1: Not a fan of the "js" folder. First because these are TS, not JS, and secondly because this is an huge catch-all: every piece of code we write that isn't Rust could seemingly match. These are all dealing with data structure implementations, so `data-structures` might make sense if you really don't want them to be top-level in `common`

tec27 (approved):

common/js/maps.ts
tec27: https://en.wikipedia.org/wiki/Multimap It has a wiki entry, I didn't make it up :P

PR #846, opened 2022-07-02
Add user context menu items to kick and ban users from chat channels.

2Pacalypse-:


client/chat/channel.tsx
```diff
@@ -322,45 +335,42 @@ function renderMessage(msg: Message) {
   }
 }
 
-type UserEntry = [userId: SbUserId, username: string | undefined]
+type UserEntry = Map<SbUserId, string | undefined>
 
 function useUserEntriesSelector(userIds: ReadonlySet<SbUserId> | undefined) {
   return useCallback(
-    (state: RootState): ReadonlyArray<UserEntry> => {
+    (state: RootState): UserEntry => {
```
tec27, line 342: The thing you return here should be readonly (probably deeply, so you can just use `Immutable<>`)

client/chat/channel.tsx
```diff
@@ -322,45 +335,42 @@ function renderMessage(msg: Message) {
...
+function areUserEntriesEqual(a: UserEntry, b: UserEntry): boolean {
+  if (a.size !== b.size) {
     return false
   }
 
-  for (let i = 0; i < a.length; i++) {
-    const [aId, aName] = a[i]
-    const [bId, bName] = b[i]
-    if (aId !== bId || aName !== bName) {
+  for (const [userId, aName] of a) {
```
tec27, line 358: I'm really not sure that this change you've made is worth the cost, imo you should probably just revert to returning tuples. If you really need a `Map` version elsewhere it is quite use to `useMemo` a call to `new Map(tuples)`, but this amount of lookups is considerably more expensive than just saying "do these 2 arrays match at every index"

I'll note that I'm not really sure why you need a Map version of these things at all. You can write the actions to be thunks such that they look up the necessary username from the state when they need it, instead of doing it how you did below. I also don't know that including the username is particularly necessary in the action text, but if you really want to, you could easily write a tiny component to do the lookup when it's rendered and you don't need to pull it from the state for the *entire channel*.

tec27 (changes requested):

client/chat/channel.tsx
tec27: I'll also just note that I think these changes you've made come from thinking about this stuff in a problematic way. When you need data in these things you should try to push that data retrieval as deep as possible in the tree. The data we retrieve for the chat channel as a whole is not intended to be data used by every single component underneath of it, it is *just* what we need to decide what to render at the top-level.

The data we had here previously was what we needed to sort the user IDs, nothing more. When we render a user in e.g. the user list, those rows retrieve their *own* data, they retrieve the rest of what the need to show. By doing things that way, we avoid re-rendering *everything* in the channel every time a small piece of information changes.

The entire point of the design of the context menu addition stuff was the function that is called there *can contain hooks*, it is effectively a component inside the menu rendering function. The point was that when you need that data, you could retrieve it at that much-deeper point, that we wouldn't be structuring components very shallow in the tree to constantly retrieve/format a bunch of data they didn't actually need (and even more crucially, that these hooks/data retrieval would only be called while the menu was actually showing).

So please, utilize the benefits of the design that we actually put there. Making this change should require *zero* changes to the data being retrieved by the chat channel

2Pacalypse-: I'm not really sure how to do this.

The `modifyMenuItems` function that is passed to the Chat component can't really have hooks inside it, since that function is called inside a `useMemo` hook in the Chat component, giving me this error:

image

But I also get linter errors for trying to call hooks inside a function that is neither a React component (doesn't start with a capital letter), or another hook (doesn't start with `use`).

I don't know if I'm completely misunderstanding your point though. I also tried looking if I can push the addition of channel menu items anywhere deeper, but I don't really see it.

client/chat/channel-menu-items.tsx
```diff
@@ -0,0 +1,131 @@
...
+import { getChatUserProfile, moderateUser } from './action-creators'
+
+const DestructiveMenuItem = styled(MenuItem)`
+  color: ${colorError};
+
+  --sb-ripple-color: ${colorError};
+`
+
+/**
+ * Even though this function is technically not a React component, nor a custom hook, we still treat
```
tec27, line 24: nit: This isn't really appropriate for being a doc comment. Doc comments are mainly for people *using* the function, this can just be a note above it or whatever

client/chat/channel.tsx
```diff
@@ -515,6 +516,7 @@ export default function Channel(props: ChatChannelProps) {
               offline={sortedOfflineUsers}
             />
           }
+          modifyMenuItems={(...args) => addChannelMenuItems(...args, channelName)}
```
tec27, line 519: nit: I think you actually could `useCallback` this safely now (since the memoization will be around creating the function reference, not the actual call to the function created)

client/chat/channel-menu-items.tsx
tec27: We could also honestly consider this to be a hook and just rename the parameter to like `useModifiedMenuItems`, which would make the linter happy as well (provided you also renamed this to match). May be easier than trying to fight it

tec27 (approved):

client/chat/channel-menu-items.tsx
2Pacalypse-: The reason why I didn't name it as a hook is that we're calling it inside another hook, which not only triggers another linter error, but makes it pretty confusing to read as well :d

tec27: Calling it inside a hook should be fine, tons of hooks are just forwarding to other hooks (e.g. `useState`, `useRef`, etc.), so I think you may be confused about the real reason for the lint error. I don't really care very much whether it is named that way or we just disable the lint though, so whatever

PR #852, opened 2022-08-24
Add channel IDs and use them instead of name for channel identification.

2Pacalypse-:
Also completely rework the channel joining and include a new
channel-status-card component that is for now only used to display
the info for a channel that user hasn't joined yet, but will probably
be used in other places too, like mentioning the channel in chat.


client/chat/connected-channel-name.tsx
```diff
@@ -0,0 +1,32 @@
...
+  /** The channel to show the corresponding name for. */
+  channelId: SbChannelId
+}
+
+/**
+ * A component which, given a channel ID, displays a channel name. Pretty barebones at the moment,
+ * but will be extended with further functionality soon.
+ */
+export function ConnectedChannelName({ className, channelId }: ConnectedChannelNameProps) {
+  const channel = useAppSelector(s => s.chat.byId.get(channelId))
```
tec27, line 24: This gets a bit weird, because potentially someone might want to use this component to display a channel name for a channel that someone hasn't joined, but you've split the reducers for joined channels and unjoined ones. I think a better structure would be to make a reducer for messages/user lists, and one for storing information about the channel (name, etc.), or make a single reducer that combines all of those things. But having 2 reducers that can both feasibly be used to look up a channel name is gonna lead to a lot of weird bugs. In general try to only have *1* place to store and look up the same information

tec27: It also isn't really safe imo to display this information without also triggering retrieval of it if it doesn't exist, should ideally do stuff similar to what we do for users with microtasks + batch retrieval. (Not very nice to users to tell them we're loading something that we never make an attempt to load :) )

client/chat/nav-entry.tsx
```diff
@@ -0,0 +1,47 @@
...
+}: ChatNavEntryProps) {
+  const onLeaveClick = useStableCallback(() => {
+    onLeave(channelId)
+  })
+
+  const button = <LeaveButton icon={<CloseIcon />} title='Leave channel' onClick={onLeaveClick} />
+
+  return (
+    <Entry
+      link={`/chat/${channelId}/${encodeURIComponent(channelName)}`}
```
tec27, line 40: Use `urlPath` instead of encoding manually

client/chat/channel-status-card.tsx
```diff
@@ -0,0 +1,200 @@
...
+  channelId,
+  channelName,
+}: ConnectedChannelStatusCardProps) {
+  const dispatch = useAppDispatch()
+  const channel = useAppSelector(s => s.chat.byId.get(channelId))
+  const channelStatus = useAppSelector(s => s.channelStatus.byId.get(channelId))
+
+  const [findChannelError, setFindChannelError] = useState<Error>()
+  const [joinChannelError, setJoinChannelError] = useState<Error>()
+  const cancelJoinRef = useRef(new AbortController())
```
tec27, line 149: imo this should probably be undefined unless you actually make a join request, but I'll admit I'm not really sure why you'd want to cancel an in-flight join request if the leave the page? If I click a join button and then navigate to a different channel I don't think I'm signifying that I don't want that to happen any more?

client/chat/channel-status-card.tsx
```diff
@@ -0,0 +1,200 @@
...
+      abortController.abort()
+    }
+  }, [])
+
+  const onViewClick = useStableCallback(() => {
+    navigateToChannel(channelId, channelName)
+  })
+
+  const onJoinClick = useStableCallback(() => {
+    dispatch(
```
tec27, line 180: This should maybe set some kind of flag to indicate a join is in progress an disable whatever button

client/chat/channel-status-card.tsx
```diff
@@ -0,0 +1,200 @@
...
+      }),
+    )
+  })
+
+  return (
+    <ChannelStatusCard
+      channelName={channelName}
+      channelStatus={channelStatus}
+      isUserInChannel={!!channel}
+      findChannelError={findChannelError}
```
tec27, line 194: This gets right back into the same sort of structure I dislike with storing errors in Redux stores, except you've built it on top of the new stuff that is supposed to fix that. Please stop building this pattern :)

If you want to write a component that doesn't touch the store at all for testing, you could pass them a function to call for the `findChannel` action, and in production (this component) you can just pass the real `findChannel`. In the unconnected component you can even type it as `typeof findChannel` to keep things in sync. Similar solution for `joinChannel`

tec27 (changes requested):

server/lib/chat/chat-service.ts
```diff
@@ -147,56 +185,58 @@ export default class ChatService {
...
+
+    return channel!
   }
 
-  async leaveChannel(channelName: string, userId: SbUserId): Promise<void> {
+  async leaveChannel(channelId: SbChannelId, userId: SbUserId): Promise<void> {
     const userSockets = this.getUserSockets(userId)
-    const originalChannelName = await this.getOriginalChannelName(channelName)
-    if (originalChannelName === 'ShieldBattery') {
+    if (channelId === 1) {
```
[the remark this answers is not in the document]
2Pacalypse-: By feature flag, do you mean something like this added to the `flags.ts`:
```ts
export const CAN_LEAVE_SHIELDBATTERY_CHANNEL = (channelId: SbChannelId) => channelId === 1 && OFF()
```

server/lib/chat/chat-service.ts
```diff
@@ -324,18 +362,48 @@ export default class ChatService {
...
+    channelId: SbChannelId,
     channelName: string,
     userId: SbUserId,
+  ): Promise<ChannelStatus> {
+    const channelInfo: ChannelInfo | undefined = (await getChannelInfo([channelId]))[0]
+
+    if (!channelInfo) {
+      // If the channel with the same name exists, it means it was closed at some point (by everyone
+      // in it leaving), and then created again by someone else, in which case it got a new ID.
+      const channelClosed = await findChannelByName(channelName)
```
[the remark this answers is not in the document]
2Pacalypse-: The problem with this is that I'm currently showing the "Create" button in the channel-status-card if the channel doesn't exist, but I'm not showing it if the channel was recreated (which is what this check _actually_ is). If we don't have a way to distinguish between these two errors, then I have to handle them the same in the channel-status-card, which might be fine?

server/lib/chat/chat-service.ts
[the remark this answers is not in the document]
tec27: no, you just add `CAN_LEAVE_SHIELDBATTERY_CHANNEL` to flags, which is a normal flag, and then you make this check `if (channel !== 1 || CAN_LEAVE_SHIELDBATTERY_CHANNEL)`

server/lib/chat/chat-service.ts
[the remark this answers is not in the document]
tec27: I think that seems fine, there really is no difference between a "recreation" and a "creation", the channel is going to have the same name, be empty, etc. either way

client/chat/channel-status-card.tsx
```diff
@@ -0,0 +1,200 @@
...
+  max-width: 440px;
+  min-width: 220px;
+  padding: 16px;
+
+  display: flex;
+  flex-direction: row;
+  align-items: center;
+`
+
+const StyledLoadingDotsArea = styled(LoadingDotsArea)`
```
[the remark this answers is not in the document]
2Pacalypse-: Hmm, the `LoadingDotsArea` component doesn't actually have the `width: 100%` set. Maybe it should?
As it is right now, it gets left-aligned in its parent.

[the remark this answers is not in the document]
tec27: Okay, then you can keep this, except I don't get why the max-width is `960px`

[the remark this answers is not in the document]
2Pacalypse-: That was just a leftover from when I first wrote this over a *month* ago, and it wasn't inside a status card back then :d. Will remove it.

client/chat/connected-channel-name.tsx
2Pacalypse-: I left the second part of your comment (triggering the retrieval of channel info) for a next PR if you don't mind. I only made a barebones version of this component for, but I will expand it significantly once I add channel mentions, which should be soon.

tec27: You should file an issue for it, but sure

client/chat/channel-info-card.tsx
```diff
@@ -0,0 +1,197 @@
...
+  const isChannelNotFound =
+    isFetchError(findChannelError) && findChannelError.code === ChatServiceErrorCode.ChannelNotFound
+  const isUserBanned =
+    isFetchError(joinChannelError) && joinChannelError.code === ChatServiceErrorCode.UserBanned
+
+  return (
+    <ChannelInfoCard
+      channelName={channelName}
+      channelInfo={channelInfo}
+      isLoading={!channelInfo && !findChannelError}
```
tec27, line 188: This is really not a good structure for this, and will be a source of many bugs and  a lot of confusion, just as it has been in the matchmaking reducer. These are not separate states, they are all just potential points in the graph of states, but you're passing them piecemeal. A user cannot be banned from a channel that does not exist, the user cannot be in a channel that does not exist, etc.

I'd still prefer that this kind of state is managed within the component itself (like I said before, replacing getChannelInfo/joinChannel in the test page is almost certainly a better solution that will result in more 'real' behavior being tested), but if you must pass it this way, you should be passing that state on a single prop, not 5 props

client/chat/join-channel.tsx
```diff
@@ -0,0 +1,107 @@
...
+          }
+
+          setJoinChannelErrorMessage(message)
+        },
+      }),
+    )
+  })
+
+  useEffect(() => {
+    const abortController = cancelJoinRef.current
```
tec27, line 70: I remain confused about why we would abort a join attempt just because the dialog was unmounted

server/lib/chat/chat-service.ts
```diff
@@ -118,85 +156,128 @@ export default class ChatService {
...
+          }
+        })
+      } catch (err) {
+        if (!(err instanceof RetryableError)) {
+          throw err
+        }
       }
+    } while (!succeeded && attempts < MAX_JOIN_ATTEMPTS)
+
+    const message = await addMessageToChannel(userId, channel!.id, {
```
tec27, line 230: This isn't safe, you never checked that the preceding operation actually succeeded? If it did not, you need to throw some other error before trying to do all this other stuff

tec27: Honestly this should almost certainly be done in the transaction (try/catches) as well. I guess you were trying to pull it out to avoid duplicating code but like, duplicated but correct code is a lot better than de-duplicated but incorrect code.

server/lib/chat/chat-service.ts
```diff
@@ -253,23 +334,23 @@ export default class ChatService {
       )
     }
 
-    const originalChannelName = await this.getOriginalChannelName(channelName)
-    if (originalChannelName === 'ShieldBattery') {
+    if (channelId === 1 && !CAN_LEAVE_SHIELDBATTERY_CHANNEL) {
+      // TODO(2Pac): Remove this restriction once we have a news/home page
```
tec27, line 338: Just put this todo around the feature flag and not everywhere that you reference the feature flag

tec27 (changes requested):

client/chat/channel-info-card.tsx
```diff
@@ -0,0 +1,157 @@
...
+      <StyledChannelIcon as={ChannelIcon} />
+    ) : (
+      <ErrorChannelIcon as={WarningIcon} />
+    )
+
+  let subtitle
+  if (isChannelNotFound) {
+    subtitle = (
+      <ErrorText>
+        We couldn't find this channel, it might not exist or it has been re-created by someone else
```
tec27, line 119: nit: This channel could not be found. It might not exist, or it may have been re-created by someone else.

client/chat/channel-info-card.tsx
```diff
@@ -0,0 +1,157 @@
...
+      </ErrorText>
+    )
+  } else if (channelInfo?.private && !isUserInChannel) {
+    subtitle = <ErrorText>This channel is private and requires an invite to join</ErrorText>
```
tec27, line 123: nit: add a period

client/chat/channel-info-card.tsx
```diff
@@ -0,0 +1,157 @@
...
+  } else if (isUserBanned) {
+    subtitle = <ErrorText>You are banned from this channel</ErrorText>
```
tec27, line 125: nit: add a period

client/chat/channel.tsx
```diff
@@ -412,82 +413,29 @@ export default function Channel(props: ChatChannelProps) {
...
-            signal: abortController.signal,
-            onSuccess: () => setJoinChannelError(undefined),
-            onError: err => setJoinChannelError(err),
-          }),
-        )
-      } else {
-        push('/')
-      }
+      dispatch(retrieveUserList(channelId))
+      dispatch(activateChannel(channelId) as any)
```
tec27, line 419: these casts can be removed now, should probably look through this PR and remove any other ones as well

tec27 (approved):

Issue #786, opened 2021-12-01
Map details editor doesn't allow empty descriptions, but empty descriptions are valid


tec27:
You can currently upload a map with an empty description, and everything works fine, except:

1) View map details for the map
2) Edit the name
3) Save the name

Expected: Saving the name works
Actual: 400 Bad request, "map description can't be empty"

claytondicks:
If empty descriptions are allowed, is there any reason for the server to throw the error when description is empty?

2Pacalypse-:
No, that restriction needs to be removed (and made sure that it doesn't break anything else).

2Pacalypse-:
Done in #855

PR #855, opened 2022-08-29
Removing bad request error for empty map description

claytondicks:
For #786 

Just removed the offending if statement. 

Test suite runs successfully and I've been able to edit / update a map without description without issue.  

Was able to create a game with that map and play with no issues.


2Pacalypse- (approved):
Thanks!

Issue #756, opened 2021-09-10
Standardize icon filenames


tec27:
The files under `client/icons/material/` are all over the place on naming scheme right now, we should standardize on something so it's easy to find things. The name for the underlying icon (minus and sizing, color, prefixes, etc.) should match what is on https://fonts.google.com/icons

I think a reasonable format for things now would be:
`[icon_name]-[size].svg`

If there is some special modifier for that icon only (I'm not sure we have any of those right now, but maybe?) then we could do something like:
`[icon_name]-[size]-[modifier].svg`

Reasoning behind this format is:
- `ic_` prefix isn't really useful for anything and was a holdover from Android naming schemes in the icon set
- `_` in the icon name and `-` to separate our sections makes it easier to pick out the `icon_name` part
- colors in the icon name aren't really useful since we recolor icons anyway, and potentially leads to people copying multiple versions of the same icon in (bad)

2Pacalypse-:
One additional thing I think we should standardize on here is the `[size]` itself. I noticed when downloading icons from google, their size uses `dp` suffix, while I think we should use `px` suffix, maybe. So e.g., `24dp` becomes `24px`.

tec27:
> One additional thing I think we should standardize on here is the `[size]` itself. I noticed when downloading icons from google, their size uses `dp` suffix, while I think we should use `px` suffix, maybe. So e.g., `24dp` becomes `24px`.

Could have sworn I mentioned that here but yes I agree :)

tec27:
No, it refers to an icon set, it's just from someone copying a file out of the icon set without renaming it

2Pacalypse-:
Done in #858

PR #858, opened 2022-09-02
Standardizing all the icons

claytondicks:
For #756 

Went through and renamed things that didn't match the new standard and updated all references.  

Found one duplicate image and just removed it

Tests came back clean and local copy seems to work without issue.


2Pacalypse- (approved):
Thanks!

PR #860, opened 2022-09-05
Sanitize map name in the replay browser.

### LINKED ISSUE #859: Replay browser details should sanitize map names (to remove color codes, etc.)
This should be the same thing we do on the server during map upload, as is it displays a bunch of tofu characters for a lot of maps.

2Pacalypse-:
Closes #859

### DIFF
--- a/common/maps.ts
+++ b/common/maps.ts
@@ -201,7 +201,7 @@ export function getTeamNames(
  * Filters out unprintable characters used for color codes in BW (we don't utilize these in our
  * client, and they just show up as tofu).
  */
-export function filterColorCodes(str: string) {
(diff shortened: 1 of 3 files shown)


common/maps.ts
```diff
@@ -197,6 +197,25 @@ export function getTeamNames(
   }
 }
 
+/**
+ * Filters out unprintable characters used for color codes in BW (we don't utilize these in our
+ * client, and they just show up as tofu).
+ */
+export function filterColorCodes(str: string) {
```
tec27, line 204: nit: note the return type here

tec27 (approved):

PR #863, opened 2022-09-08
Augment the channel bans to use the client identifiers.

2Pacalypse-:


server/lib/chat/chat-models.ts
```diff
@@ -381,13 +381,55 @@ export async function updateUserPermissions(
   }
 }
 
+export async function countBannedChannelIdentifiers(
```
tec27, line 384: nit: name is a bit confusing to me because it sounds like the identifiers are for the channel, not the user. Maybe `countBannedIdentifiersForChannel`?

server/lib/chat/chat-models.ts
```diff
@@ -398,14 +440,60 @@ export async function banUserFromChannel(
   }
 }
 
+export async function banAllChannelIdentifiers(
```
tec27, line 443: `banAllIdentifiersFromChannel`?

server/lib/chat/chat-service.ts
```diff
@@ -161,6 +164,21 @@ export default class ChatService {
...
   }
 
+  private async banUserFromChannelIfNeeded(
+    channelId: SbChannelId,
+    targetId: SbUserId,
+    client: DbClient,
+  ): Promise<boolean> {
+    const count = await countBannedChannelIdentifiers({ channelId, targetId }, client)
+    if (count >= MIN_IDENTIFIER_MATCHES) {
+      await banUserFromChannel({ channelId, targetId, reason: 'ban evasion' }, client)
```
tec27, line 174: This was fine for being exposed to admins, I would not do this for regular users. imo copy the previous ban reason (any of the previous ban reasons is fine, since there's no guarantee they'll all be the same, but perhaps picking the one with the most occurrences is best. This should be doable with a modification to the existing query), add a column to mark this as an automated enforcement (and mark automated enforcements in some way in the UI, so people aren't confused why they have users in the list they did not manually ban)

tec27 (changes requested):

server/lib/chat/chat-models.ts
```diff
@@ -381,18 +381,131 @@ export async function updateUserPermissions(
...
-      INSERT INTO channel_bans (user_id, channel_id, ban_time, banned_by, reason)
-      VALUES (${targetId}, ${channelId}, ${new Date()}, ${moderatorId}, ${reason});
+    if (automated) {
+      await client.query(sql`
+        WITH rc AS (
+          SELECT reason, COUNT(*) AS reason_count
+          FROM channel_bans
+          WHERE channel_id = ${channelId} AND user_id = ANY(${connectedUsers})
+          GROUP BY reason
+          ORDER BY reason_count DESC
```
tec27, line 450: probably want a secondary order here too so that ties are at least deterministic (ban date most likely)

server/lib/chat/chat-models.ts
```diff
@@ -381,18 +381,131 @@ export async function updateUserPermissions(
...
+    moderatorId,
+    targetId,
+    reason,
+    automated = false,
+    connectedUsers = [],
+  }: {
+    channelId: SbChannelId
+    moderatorId?: SbUserId
+    targetId: SbUserId
+    reason?: string
```
tec27, line 435: If reason and automated are exclusive, the type should specify things that way. For example:
```ts
{ channelId: SbChannelId, /*...*/, reason: string } | { channelId: SbChannelId, /*...*/, automated: true, connectedUsers: SbUserId[] }
```

server/lib/chat/chat-models.ts
```diff
@@ -381,18 +381,131 @@ export async function updateUserPermissions(
...
+  },
+  withClient?: DbClient,
 ): Promise<void> {
-  const { client, done } = await db()
+  const { client, done } = await db(withClient)
   try {
-    await client.query(sql`
-      INSERT INTO channel_bans (user_id, channel_id, ban_time, banned_by, reason)
-      VALUES (${targetId}, ${channelId}, ${new Date()}, ${moderatorId}, ${reason});
+    if (automated) {
```
tec27, line 443: Probably want a sanity check (before the `db()` call) that `connectedUsers` has `length > 0`

server/lib/chat/chat-service.ts
```diff
@@ -176,25 +201,34 @@ export default class ChatService {
...
+    let error: ChatServiceError | undefined
     do {
       attempts += 1
       try {
         await transact(async client => {
           channel = await findChannelByName(channelName, client)
           if (channel) {
             if (this.state.users.has(userId) && this.state.users.get(userId)!.has(channel.id)) {
-              throw new ChatServiceError(
+              error = new ChatServiceError(
```
tec27, line 212: This is an incorrect change, you *want* to throw this here, you *want* this transaction to be rolled back. The fix was to pass back a flag (either by setting something from the closure, or by returning a value (`transact` will return whatever you return from the transaction function), and checking that to see if the user was banned. It was *not* to avoid throwing *every* error

server/lib/chat/chat-service.ts
```diff
@@ -176,25 +201,34 @@ export default class ChatService {
...
               )
+              return
             }
 
             const isBanned = await isUserBannedFromChannel(channel.id, userId, client)
             if (isBanned) {
-              throw new ChatServiceError(
+              error = new ChatServiceError(
+                ChatServiceErrorCode.UserBanned,
+                'This user has been banned from this chat channel',
```
tec27, line 223: These messages are potentially user-visible and should be phrased as such, they are not for our log purposes

server/lib/chat/chat-service.ts
```diff
@@ -176,25 +201,34 @@ export default class ChatService {
...
+              )
+              return
+            } else if (await this.banUserFromChannelIfNeeded(channel.id, userId, client)) {
```
tec27, line 226: You can avoid duplicating the exact same error code by just combining these conditions:
```ts
if (isBanned || await this.banUserFromChannelIfNeeded(channel.id, userId, client)) { /* ... */ }
```

tec27 (changes requested):

server/lib/chat/chat-service.ts
tec27: So thinking about this further, I think handling things in this way is really a mistake to begin with. Generally you should try to do things idempotently: 2 requests for the same thing coming in should deliver the same result. If I as a user request to join a channel twice, and the 2nd request hits at a point where I've already joined that channel, why would I care that I'm already in the channel? Better for the server to tell me "sure, you're in that channel". Programming it to be an error makes it very easy to do bad things on the client (e.g. show the user an error dialog instead of just... taking them to the channel they wanted to be in).

So, don't write things this way, this is not an error.

Your "channel exists" code here should have 3 cases:
1) The user is already in the channel: cool, we're done, mark this as successful and return
2) The user is banned or becomes banned: cool, we're done, mark that they are banned and return (throw an error corresponding to being banned outside of the transaction)
3) The user is not in the channel: do the things necessary to have them join, return

tec27: I'll also note that I don't really think checking `this.state` to determine if the user is in the channel is a great idea here. That state is not in sync with what is in the DB, it is only updated *after* DB operations complete, with some non-determinant amount of time between their completion and the update. Would almost certainly be better to actually query the DB for this information here.

(Aside from the out-of-sync stuff, I think `this.state.users` only contains users that currently have a connected websocket? So it's even worse to use that for this purpose)

server/lib/chat/chat-models.ts
2Pacalypse-: Hmm, adding ban date to the `ORDER BY` clause requires me to add it to the `GROUP BY` clause as well, I think, and I'm not sure if that's desirable?

tec27: That doesn't seem to be the case given that `reason_count` is not in the `GROUP BY`, if you're getting an error here it's likely that it needs to appear in the `SELECT`. Since the goal is to pick the value with the highest `reason_count` and earliest date, you could likely add `MIN(date)` to the select and use that here

server/lib/chat/chat-service.ts
```diff
@@ -201,33 +201,22 @@ export default class ChatService {
     let channel: ChannelInfo | undefined
     let userChannelEntry: UserChannelEntry | undefined
     let message: ChatMessage
-    let error: ChatServiceError | undefined
+    let isUserBanned: boolean | undefined
```
tec27, line 204: imo this it's better to just set this to `false`, no need for type annotations that way

tec27 (approved):

Issue #861, opened 2022-09-05
Adjust postgres random_page_cost


tec27:
Currently our postgres server just uses the default conf from the docker container, would be good to adjust at least `random_page_cost` since its default value is a bit off from reality on modern servers

PR #865, opened 2022-09-12
Updated random_page_cost value for postgres

claytondicks:
For #861 

I've just updated the command in the docker-compose file to update this setting to 1.1 which is typical when running public clouds.  Can be adjusted to 1.5 if the storeage is on the slower side

Also, let me know if you'd prefer to just add a full custom conf file vs the one command and I can update PR

### DIFF
--- a/deployment/appserver/db/postgresql.conf
+++ b/deployment/appserver/db/postgresql.conf
@@ -1,0 +1,751 @@
+# -----------------------------
+# PostgreSQL configuration file
+# -----------------------------
+#
+# This file consists of lines of the form:
+#
+#   name = value
+#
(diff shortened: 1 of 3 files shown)


tec27:
I do think we'd prefer to have this in a config file rather than in the launch arguments (and if possible, be able to share that config between our production setup [`./deployment/appserver/docker-compose.yml`] and our development setup [`./docker-compose.yml`], although that isn't a dealbreaker for me)

claytondicks:
Sounds good.  I've gone ahead and updated it to reflect the new config file.  Also added it to the dev setup docker-compose as well

tec27:
Awesome, thank you!

PR #868, opened 2022-09-17
Rename the "high_traffic" column to "official" in the channels table.

2Pacalypse-:
Official channels have the following characteristics:
- they're created by staff members only
- they don't have owners (so only server moderators can moderate them)
- they're not deleted when everyone leaves (thus keeping the message history in perpetuity)
- TODO MAYBE: join/leave messages are not shown in them
- TODO MAYBE: new accounts are automatically joined in all (some?) of them

TODO: Currently, every channel that is created is a non-official channel (except ShieldBattery which was created as an official channel in a migration), so once we make it possible to update channel settings, we gotta make sure to set the channel owner to NULL if the channel is being made into an official channel. I don't think it's that big of a deal even if we don't, since only staff will most likely be able to make official channels, but it'll make things easier in case of de-modding or something.

### DIFF
--- a/server/lib/chat/chat-models.ts
+++ b/server/lib/chat/chat-models.ts
@@ -314,32 +314,32 @@ export async function removeUserFromChannel(
     //   - `toggle_private`
     //   - `change_topic`
     // If there's no such user, then the user who has joined the channel earliest is chosen.
-    const newOwnerResult = await client.query<DbUserChannelEntry>(sql`
-      SELECT user_id
-      FROM channel_users
-      WHERE channel_id = ${channelId}
-      ORDER BY
-        edit_permissions DESC,
-        ban DESC,
-        kick DESC,
-        toggle_private DESC,
-        change_topic DESC,
-        join_date
-      LIMIT 1;
+    const newOwnerResult = await client.query<{ owner_id: SbUserId }>(sql`
+      WITH own AS (
+        SELECT user_id
+        FROM channel_users
+        WHERE channel_id = ${channelId}
+        ORDER BY
+          edit_permissions DESC,
+          ban DESC,
+          kick DESC,
+          toggle_private DESC,
+          change_topic DESC,
+          join_date
+        LIMIT 1
+      )
+      UPDATE channels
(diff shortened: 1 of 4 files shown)


tec27:
> * TODO MAYBE: new accounts are automatically joined in all (some?) of them

don't think this is anything we'd actually build, auto-join is a special case for particular channels and related to other features (e.g. the "main" channel [account creation], rank channels [matchmaking/ranked]) not some feature we'd apply to every official channel

> TODO: Currently, every channel that is created is a non-official channel (except ShieldBattery which was created as an official channel in a migration), so once we make it possible to update channel settings, we gotta make sure to set the channel owner to NULL if the channel is being made into an official channel. I don't think it's that big of a deal even if we don't, since only staff will most likely be able to make official channels, but it'll make things easier in case of de-modding or something.

I think I would just allow official channels to be created through some admin interface, I don't think it needs to be something you can do on an existing channel

server/lib/chat/chat-models.ts
```diff
@@ -304,12 +305,12 @@ export async function removeUserFromChannel(
...
 
-    const highTrafficChannelResult = await client.query(sql`
+    const officialChannelResult = await client.query(sql`
       SELECT id FROM channels
-      WHERE id = ${channelId} AND high_traffic = true;
+      WHERE id = ${channelId} AND official = true;
     `)
-    if (highTrafficChannelResult.rowCount > 0) {
-      // Don't transfer ownership in "high traffic" channels
+    if (officialChannelResult.rowCount > 0) {
```
tec27, line 312: `rowCount` is actually not the right thing to use here (I learned this from doing stuff on the friends list). `rowCount` is intended to return the number of rows modified by a command (e.g. how many rows were `DELETE`d, `UPDATE`d, etc.). Use `officialChannelResults.rows.length` instead

Sort of relatedly, there's a bunch of queries in here that could return enough info to deal with all of these conditions, instead of having to run back to the db a ton of times, would probably be good to clean that up

tec27 (approved):

tec27 (approved):

PR #869, opened 2022-09-18
Introduce POM for integration tests

v1ktor:
Move common methods related to the login functionality that are used in the integration tests to the [POM](https://playwright.dev/docs/test-pom).

### DIFF
--- a/package.json
+++ b/package.json
@@ -91,6 +91,7 @@
     "start-server": "node ./server/index.js | node ./server/utils/pino-pg | pino-pretty -c -t",
     "test": "jest",
     "test:integration": "playwright test",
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -91,6 +91,8 @@
     "start-server": "node ./server/index.js | node ./server/utils/pino-pg | pino-pretty -c -t",
     "test": "jest",
     "test:integration": "playwright test",
+    "test:integration:debug": "playwright test --debug",
```
tec27, line 94: I'm fine with having these, but I'll also just note you can typically pass arguments to any commands here, e.g. you could (most likely) do this by doing:

```
yarn run test:integration --debug
```

PR #872, opened 2022-09-29
Add a way for admins to search and view any chat channel.

2Pacalypse-:
The main purpose of this work is to let admins check if any channels are used to post illegal content.

This PR also includes an API to search channels for regular users as well, which should be used pretty soon.


server/lib/chat/chat-models.ts
```diff
@@ -578,3 +579,94 @@ export async function findChannelByName(
...
+    done()
+  }
+}
+
+/**
+ * Gets a list of chat channels for a particular user, optionally filtered by a `searchStr`. If the
+ * user is an admin, this method will return a full list of channels. Otherwise, this method will
+ * only return non-private channels and channels that the user is not joined in.
+ */
+export async function getChannelsForUser(
```
tec27, line 612: This is a very weird name for this function, it sounds like it's retrieving channels the user is in or something. I think `searchChannels` or something like that would be better

server/lib/chat/chat-models.ts
```diff
@@ -578,3 +579,94 @@ export async function findChannelByName(
...
+  userId: SbUserId,
+  isAdmin: boolean,
+  limit: number,
+  pageNumber: number,
+  searchStr?: string,
```
tec27, line 617: This many params should really go into an object, not passed positionally (especially true given that one of them is optional and you *also* have an optional client param, and specifying one doesn't mean you'd want to specify the other). [In case it's not clear based on all our other code, the `withClient` param should stay outside of the parameters object, as the second param]

server/lib/chat/chat-models.ts
```diff
@@ -578,3 +579,94 @@ export async function findChannelByName(
...
+  withClient?: DbClient,
+): Promise<{ channels: ChannelInfo[]; total: number }> {
+  const { client, done } = await db(withClient)
+  try {
+    const query = isAdmin
```
tec27, line 622: I don't think this is really a great way to structure this functionality, it becomes rather unwieldy very quickly since the structure of these 2 queries is very different, but you need to add modifications based on things with the same name below. It also just becomes likely that you can break one and only test the other and not know/realize. I would split this into 2 functions: 1 for admin usage, 1 for "normal" usage. I don't think keeping these in the same function and reusing a few lines below is really buying you very much (and if you *really* can't deal with that duplication, you could move that logic into some function they both call into, but I do not think that's very worth it)

server/lib/chat/chat-models.ts
```diff
@@ -578,3 +579,94 @@ export async function findChannelByName(
     done()
   }
 }
+
+/** Returns the number of channels that match a particular condition. */
+async function getChannelsCount(
```
tec27, line 584: Weird name, maybe `countMatchingChannels`? I'm not sure that I really like this API at all though, it feels like My First SQL Injection. Would prefer writing targeted (static) queries rather than making queries we just plug random SQL into

server/lib/chat/chat-models.ts
```diff
@@ -578,3 +579,94 @@ export async function findChannelByName(
...
+      GROUP BY c.id
+      ORDER BY user_count DESC, name
+      LIMIT ${limit}
+      OFFSET ${pageNumber * limit}
+    `)
+
+    const result = await client.query<DbChannel>(query)
+
+    const channels = await Promise.all(result.rows.map(row => convertChannelFromDb(row)))
+    const total = await getChannelsCount(whereCondition, withClient)
```
tec27, line 666: This isn't a safe thing to do, the count can change between the original query you made and this query (and this also needlessly goes to the database twice for this information). Do you really need this information? It makes the query monumentally more expensive (the query now has to run against the entire table each time you grab a page, even though it may be able to deliver the page from a much smaller subset of the data) and it doesn't seem to be particularly useful in any UI

server/lib/chat/chat-models.ts
```diff
@@ -578,3 +579,94 @@ export async function findChannelByName(
...
+    query.append(sql`
+      GROUP BY c.id
+      ORDER BY user_count DESC, name
+      LIMIT ${limit}
+      OFFSET ${pageNumber * limit}
+    `)
+
+    const result = await client.query<DbChannel>(query)
+
+    const channels = await Promise.all(result.rows.map(row => convertChannelFromDb(row)))
```
tec27, line 665: Why is this a Promise.all? This function does not return a promise (and it should absolutely never be async, so I'm at a loss as to why this was ever thought to be needed)

server/lib/chat/chat-models.ts
```diff
@@ -40,7 +40,7 @@ function convertUserChannelEntryFromDb(props: DbUserChannelEntry): UserChannelEn
  * Gets a user channel entry for each channel that a particular user is in, ordered by their channel
  * join date.
  */
-export async function getChannelsForUser(userId: SbUserId): Promise<UserChannelEntry[]> {
+export async function getUserChannelEntriesForUser(userId: SbUserId): Promise<UserChannelEntry[]> {
```
tec27, line 43: I think the previous naming was better, I don't really get why you handed this name to the function you created

tec27 (changes requested):

server/lib/chat/chat-models.ts
2Pacalypse-: Copy-paste fail :d

server/lib/chat/chat-models.ts
2Pacalypse-: Hmm, the only thing I need the total count for is for the infinite-loading-list in the UI (rendered through a carousel), to know when it reached the end. I'll try to think of a better way to implement pagination for that.

tec27: You can do it like we do for chat messages:
1) Load a page
2) Does it have the amount of entries we asked for?
  - Yes: We still have more pages
  - No: We're done

server/lib/chat/chat-models.ts
2Pacalypse-: I went with `listChannels` since the filtering by search string is optional, and by default this function just returns a list of all (available) channels.

client/admin/chat-channels.tsx
```diff
@@ -0,0 +1,269 @@
...
+
+  const onSearchChange = useStableCallback((searchQuery: string) => {
+    debouncedSearchRef.current(searchQuery)
+  })
+
+  const onLoadMoreChannels = useStableCallback(() => {
+    setIsLoadingMoreChannels(true)
+    setCurrentPage(currentPage + 1)
+
+    fetchJson<ChannelInfo[]>(
```
tec27, line 155: You need some way to abort this, as-is you can screw up the results requesting the next page and then triggering a new search probably

client/admin/chat-channels.tsx
```diff
@@ -0,0 +1,269 @@
...
+        setHasMoreChannels(data.length >= SEARCH_CHANNELS_LIMIT)
+      })
+      .catch(err => {
+        setIsLoadingMoreChannels(false)
+        setSearchError(err)
+      })
+  })
+
+  const [selectedChannel, setSelectedChannel] = useState<ChannelInfo>()
+  const channelUsers = useAppSelector(s =>
```
tec27, line 170: imo it'd be a lot better to split this stuff out into a component you use here and in the actual channel, this feels like way too much duplication

server/lib/chat/chat-models.ts
```diff
@@ -578,3 +579,105 @@ export async function findChannelByName(
...
+        FROM channel_users
+        WHERE channel_id NOT IN (SELECT channel_id FROM channel_users WHERE user_id = ${userId})
+      )
+      SELECT c.*, COUNT(cu.user_id) AS user_count
+      FROM channels c
+      INNER JOIN unjoined_channels ON c.id = unjoined_channels.channel_id
+      INNER JOIN channel_users cu ON c.id = cu.channel_id
+    `
+
+    const whereCondition: SQLStatement = sql`WHERE private = false`
```
tec27, line 615: unnecessary type

server/lib/chat/chat-models.ts
```diff
@@ -578,3 +579,105 @@ export async function findChannelByName(
...
+    if (searchStr) {
+      const escapedStr = `%${searchStr.replace(/[_%\\]/g, '\\$&')}%`
```
tec27, line 617: You wrote a function to sanitize these, why are you copy-pasting this?

server/lib/chat/chat-models.ts
```diff
@@ -578,3 +579,105 @@ export async function findChannelByName(
...
+  try {
+    const query = sql`
+      SELECT c.*, COUNT(cu.user_id) AS user_count
+      FROM channels c
+      INNER JOIN channel_users cu ON c.id = cu.channel_id
+    `
+
+    let whereCondition: SQLStatement | undefined
+    if (searchStr) {
+      const escapedStr = `%${searchStr.replace(/[_%\\]/g, '\\$&')}%`
```
tec27, line 664: Again, use the function for this

server/lib/chat/chat-models.ts
```diff
@@ -578,3 +579,105 @@ export async function findChannelByName(
     done()
   }
 }
+
+/**
+ * Returns a list of non-private chat channels that a user is not joined in, optionally filtered by
+ * a `searchStr`.
+ */
+export async function listChannels(
```
tec27, line 587: If this is returning unjoined channels only, it should probably be called `listUnjoinedChannels` or `listJoinableChannels` (probably better considering this filters only visible channels). Alternatively you could make that an option (and probably control it in the API with a query parameter), which would make a lot more sense from a REST perspective

server/lib/chat/chat-models.ts
```diff
@@ -578,3 +579,105 @@ export async function findChannelByName(
...
+): Promise<ChannelInfo[]> {
+  const { client, done } = await db(withClient)
+  try {
+    const query = sql`
+      WITH unjoined_channels AS (
+        SELECT DISTINCT(channel_id)
+        FROM channel_users
+        WHERE channel_id NOT IN (SELECT channel_id FROM channel_users WHERE user_id = ${userId})
+      )
+      SELECT c.*, COUNT(cu.user_id) AS user_count
```
tec27, line 609: This count seems incredibly expensive and should probably be tracked in the channel itself or something. You should definitely look at the query plan for this because I cannot imagine it is good

tec27 (changes requested):

client/admin/chat-channels.tsx
```diff
@@ -0,0 +1,242 @@
...
+  })
+
+  const [selectedChannel, setSelectedChannel] = useState<ChannelInfo>()
+  const channelMessages = useAppSelector(s =>
+    selectedChannel ? s.chat.idToMessages.get(selectedChannel.id) : undefined,
+  )
+
+  const onLoadMoreMessages = useStableCallback(() => {
+    if (selectedChannel) {
+      dispatch(getMessageHistory(selectedChannel.id, CHANNEL_MESSAGES_LIMIT))
```
tec27, line 176: I would very much prefer that you don't call things here that mess with the reducer data we use for normal chat stuff (here and the user list below). I also don't think you've actually tested any of this realistically, because the service throws errors if the user is not in the channel specified, and the whole point of this UI is that the admin doesn't need to be in the channel, so please actually test this code.

tec27 (changes requested):

client/admin/chat-channels.tsx
```diff
@@ -0,0 +1,242 @@
...
+
+    abortControllerRef.current?.abort()
+    abortControllerRef.current = new AbortController()
+
+    fetchJson<ChannelInfo[]>(
+      apiUrl`admin/chat/?q=${searchQuery}&limit=${SEARCH_CHANNELS_LIMIT}&page=${currentPage + 1}`,
+      { signal: abortControllerRef.current.signal },
+    )
+      .then(data => {
+        setIsLoadingMoreChannels(false)
```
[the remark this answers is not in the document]
2Pacalypse-: Hmm, I'm not sure I quite understand the first part. Won't the `fetch` reject with an error if the signal is aborted and these won't get called? Also, can you clarify what you mean by storing a reference to AbortController outside of the ref.

server/lib/chat/chat-models.ts
```diff
@@ -578,3 +580,116 @@ export async function findChannelByName(
...
+          SELECT DISTINCT(channel_id)
+          FROM channel_users
+          WHERE channel_id NOT IN (SELECT channel_id FROM channel_users WHERE user_id = ${userId})
+        )
+        SELECT c.*
+        FROM channels c
+        INNER JOIN unjoined_channels ON c.id = unjoined_channels.channel_id
+      `
+
+    const whereCondition = sql`WHERE private = false`
```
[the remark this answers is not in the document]
2Pacalypse-: I thought about it, and I'm not sure if we ever want to display private channels in the search results for non-admins, even the ones you've joined in. So there are two use cases I was thinking of where we might use this API:
- join-channel dialog: here we'll probably use it without the `joined` flag, so it doesn't matter
- in the autocomplete box when mentioning the channel in chat: here we'll probably use the `joined` flag (and order them on top of results), but I don't think we want to show the private channels you're in when trying to mention a channel

client/admin/chat-channels.tsx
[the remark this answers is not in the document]
tec27: When promises resolve, the callbacks are called asynchronously. Thus you can have:
fetch() -> resolves
abort()
callback()

So you need to check that the AbortSignal has not become aborted since the fetch returned. Is this rare? Yes. Does it happen? Yes, so you need to handle it

As far as storing a reference outside of the ref, you need to store it in a local variable so you can access the same AbortSignal that you called `fetch` with (because in the case that it was aborted by a new fetch, your ref would be set to a new instance of `AbortController`, and thus have a new, non-aborted signal)

server/lib/chat/chat-models.ts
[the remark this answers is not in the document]
tec27: When you're writing external APIs, it's generally more okay to have them targeted to particular usecases, mostly because REST is not particularly expressive as a type system and the effort required to serve a general-purpose API that we will never use is not worth it.

When you're writing APIs for internal usage (which is what all of our model APIs are), they should either be properly limited by the types+parameters they allow, or as general as the types+parameters imply. The parameters you've created here, to me, imply that I will get all the channels that I've joined, regardless of whether or not they are private. This system does not imply that we are limited to 2 usecases only, and does not imply the assumptions you've stated above. If you want a system that does that, you probably want to create 2 model methods instead, one for each usecase, that are named specifically for those usecases. Otherwise, someone is going to come along 2 months from now and use this method thinking it does what they want, when it very much does not, we're not going to notice, it'll get shipped into production, and 6 months down the line we'll realize it's been broken the whole time.

As far as the usecases you stated above, I think your second one is also just mistaken, I don't see why we would ever want to avoid having a joined channel in the mention autocomplete. The channels are just not able to be joined without an invite, they're not hidden, and that has never been the meaning of that flag.

server/lib/chat/chat-models.ts
```diff
@@ -578,3 +580,116 @@ export async function findChannelByName(
...
+  withClient?: DbClient,
+): Promise<ChannelInfo[]> {
+  const { client, done } = await db(withClient)
+  try {
+    const query = sql`
+      SELECT *
+      FROM channels
+    `
+
+    let whereCondition: SQLStatement | undefined
```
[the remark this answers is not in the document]
2Pacalypse-: Hmm, this actually produces a syntax error in SQL for some reason:

image

Adding quotes around the whole search expression gets rid of the syntax error, but produces a different error:

image

This works though for some reason 🤷‍♂️:
```js
const escapedStr = `%${escapeSearchString(searchStr)}%`
query.append(sql`WHERE name ILIKE ${escapedStr}`)
```

2Pacalypse-:
Tested:
- open channel view in admin panel -> expected to see initial channels ordered by user count
- scroll the carousel beyond 10 channels -> expected to load the next page of results
- enter search string in search box -> expected to get channels with that string
- enter search string in search box while on page 2 -> expected for page to reset to 1
- click view in channel card info -> expected for channel messages and users to load
- right click on channel user -> expected to get global menu items
- right click on user mention in channel -> expected to get global menu items
- scroll the channel messages -> expected to load more messages
- click view on a different channel -> expected for new channel to load correctly
- click view on a joined private channel -> expected channel to load correctly
- click view on a non-joined channel -> expected channel to load correctly
- click view on a non-joined private channel -> expected channel to load correctly
- open chat channel normally -> expected nothing to be broken

server/lib/chat/chat-models.ts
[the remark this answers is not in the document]
tec27: That's just because of how the library is choosing to pass in the string, I think. In my code, it passes the string separate from the `%`s, in yours it concatenates them before passing them as a single parameter. So you could fix it in a single line by doing:
```ts
query.append(sql`WHERE name ILIKE ${`%${escapeSearchString(searchStr)}%`}`)
```

server/lib/chat/chat-models.ts
```diff
@@ -578,3 +580,120 @@ export async function findChannelByName(
...
+          FROM channel_users
+          WHERE channel_id NOT IN (SELECT channel_id FROM channel_users WHERE user_id = ${userId})
+        )
+        SELECT c.*
+        FROM channels c
+        INNER JOIN unjoined_channels ON c.id = unjoined_channels.channel_id
+      `
+
+    query.append(sql`WHERE private = false`)
+    if (searchStr) {
```
tec27, line 640: This filter doesn't apply to the joined channels, which just seems broken? Instead of trying to build up this complex series of appends to generate the query you want without duplicating code, I think you would get a lot of value out of just writing the 2 queries you want straightforwardly, with as few conditional statements as possible. I recognize that you will repeat some code to do that, but I promise it is better than trying to figure out all the possible queries that can be generated by this current code.

tec27 (changes requested):

tec27:
Other possible thought here: just delete this non-admin channel search model function. It shouldn't be part of this PR anyway, and at this point has dominated the majority of the comments, delete it and come back to it with a fresh approach a different day.

PR #881, opened 2022-11-23
Add the initial integration of the i18next to become multilingual.

2Pacalypse-:
This commit doesn't actually add any translation strings yet. That should probably come in other (multiple) PRs.

TODO:
- Update the types of the i18next library to cover our use cases.
- Actually use the translation library in all the user-facing strings.
- Extend the setup to cover some non-client files, e.g. mail templates, if possible.
- Create a language switcher and save the user preferences to (server?) settings.
- Use language detection?
- Use local storage caching?


common/i18n.ts
```diff
@@ -0,0 +1,21 @@
+/**
+ * A string representation of each of the languages that we support and expect to have a translation
+ * file for.
+ */
+export enum TranslationLanguages {
```
tec27, line 5: nit: not really the standard naming style for enums, this (and the one below) should be non-plural

common/i18n.ts
```diff
@@ -0,0 +1,21 @@
...
+
+export const ALL_TRANSLATION_LANGUAGES: ReadonlyArray<TranslationLanguages> =
+  Object.values(TranslationLanguages)
+
+/**
+ * A string representation of all the namespaces we're using for our translation files. Currently
+ * we're not using different namespaces and we're just defining the default one.
+ */
+export enum TranslationNamespaces {
+  Translation = 'translation',
```
tec27, line 17: Would prefer something like "Global" or "Root" or something I think

.github/workflows/tests.yml
```diff
@@ -60,6 +60,9 @@ jobs:
       - name: Run unit tests
         run: yarn test
 
+      - name: Check translation files
+        run: yarn gen-translations --fail-on-update
```
tec27, line 64: Would be nice to make this give particular instructions if it fails (e.g. explain what this failure means and how to fix it)

client/index.jsx
```diff
@@ -140,17 +142,19 @@ Promise.all([rootElemPromise])
...
-            <RedirectProvider>
-              <>
-                <App />
-                {ReduxDevToolsContainer ? <ReduxDevToolsContainer /> : null}
-              </>
-            </RedirectProvider>
-          </Router>
-        </ReduxProvider>
-      </RootErrorBoundary>,
+      <React.Suspense fallback={<LoadingIndicator />}>
```
tec27, line 145: This should at the very least be inside the error boundary imo, but you could probably put it deeper than that (e.g. wrap the App directly)

client/users/mini-match-history.tsx
```diff
@@ -168,20 +175,20 @@ export function ConnectedGameListEntry({
     return 'unknown'
   }, [results, forUserId])
 
-  const matchType = getGameTypeLabel(game)
-  const mapName = map?.name ?? 'Unknown map'
+  const matchType = getGameTypeLabel(game, t)
+  const mapName = map?.name ?? t('games.miniMatchHistory.mapNameUnknown', 'Unknown map')
```
tec27, line 179: These sorts of strings are not really particular to any UI context and should be pulled from some common place

common/matchmaking.ts
```diff
@@ -19,12 +20,12 @@ export enum MatchmakingType {
 
 export const ALL_MATCHMAKING_TYPES: ReadonlyArray<MatchmakingType> = Object.values(MatchmakingType)
 
-export function matchmakingTypeToLabel(type: MatchmakingType): string {
+export function matchmakingTypeToLabel(type: MatchmakingType, t?: TFunction): string {
```
tec27, line 23: I know we needed this to be optional for team names, but I don't think this one needs to be? If you're just doing that for migration purposes you should leave a TODO + file an issue for cleaning it up

tec27 (approved):

client/index.jsx
2Pacalypse-: My thinking was that we'd want the error boundary translated as well, so I wasn't sure if this Suspense needs to be above it for some reason. But probably not now that I think about it.

tec27: This wouldn't really make that happen, the suspense doesn't cause anything to be available or not, it just catches the thrown promises and shows a loading indicator. The point of the root error boundary is that it will catch *all* errors, so that nothing ends up totally crashing the app, so it should be outside of everything. (Since, potentially, there could be errors in the translation loading stuff, we don't want to let those pass through uncaught).

I would agree we'd want to translate that stuff if possible, but you need to handle the case that an error happened before they got loaded regardless

tec27: Also just for future reference, if you *do* want to translate strings inside the RootErrorBoundary UI, you need to do so via something that *does not* utilize Suspense/throw promises/etc. It would need to try to use translations if they're available immediately but fall back to the defaults if not, because once the RootErrorBoundary is hit we can no longer trust that any app state is correct, that fetches will work, etc.

2Pacalypse-: Yeah, that's a good point. I'll try to keep it in mind :d

Another thing I was wondering about Suspense that I'm not sure how it works. For example, when the user switches the language, the new translation file is being fetched and the Suspense renders the fallback while that happens. But, does the user loses all transient state when the fallback is rendered (e.g. by unmounting all the component below the Suspense)?

tec27: I'm not sure, I haven't used Suspense extensively. It should be easy to write a small demo that would tell you, though. (I also don't think it's really a big deal were that the case, I'd just change the flow such that language changes notify users they must reload the app, then do so.

Also, after thinking about this more, this Suspense actually needs to be even deeper in the tree, it doesn't make sense to hide the window controls just because translation files are loading

PR #888, opened 2023-02-21
Add auto-save functionality to settings.

2Pacalypse-:
Reworks the settings forms to not have a Save button, but instead do an auto-save on each change.

This PR also reworks the settings so they're not rendered in a dialog, but instead in a dedicated page, a la Discord.

The settings reducer has been moved to Immer and there's been some other changes in here that I now forgot about :(


client/settings/settings.tsx
```diff
@@ -1,228 +1,370 @@
...
-    case SettingsTab.App:
-      contents = (
-        <AppSettings
-          localSettings={tempLocalSettings}
-          formRef={formRef}
-          onChange={onSettingsChange}
-          onSubmit={onSettingsSubmit}
-        />
-      )
+  let contents: React.ReactNode | undefined
```
tec27, line 187: This pattern is imo a mess, write a component that has both of these values on it and does the right thing. Bonus points for making it something that returns from the switch instead of mutating a variable that is used later.

client/settings/settings.tsx
```diff
@@ -1,228 +1,370 @@
...
-      onCancel={onCancel}>
-      <ContentsBody>
-        {lastError ? (
-          <ErrorText>There was an issue saving the settings. Please try again.</ErrorText>
+    <Container style={style}>
+      <NavContainer>
+        {IS_ELECTRON ? (
+          <>
+            <NavSectionTitle>App settings</NavSectionTitle>
+            <NavEntry
```
tec27, line 228: This is also a mess of repeated code with zero changed logic. Write a component that returns the right text label given a subpage enum value (so that they can be translated). Use that component in this piece, as well as in the component you write for rendering the content (from the above comment).

Sections can be rendered as:
```tsx
[AppSettingsSubPage.Sound, AppSettingsSubPage.System].map(s => <NavEntry key={s} subPage={s} isActive={subPage === s} onChangeSubPage={onChangeSubPage} />)
```

tec27 (changes requested):

client/settings/action-creators.ts
```diff
@@ -1,56 +1,68 @@
...
-      )
-    }
+// TODO(2Pac): Clear or migrate the previously saved tab in local storage to this?
+const savedSettingsSubPage = new JsonLocalStorageValue<SettingsSubPage>('settingsSubPage')
+
+export function openSettings(subPage?: SettingsSubPage): OpenSettings {
+  return {
+    type: '@settings/openSettings',
+    payload: {
+      subPage: subPage ?? savedSettingsSubPage.getValue() ?? AppSettingsSubPage.Sound,
```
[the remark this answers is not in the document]
2Pacalypse-: > Why sound?

Because it is rendered in the first position in the new settings navigation :d

I left a TODO to replace it with user account details page once I move that over to the new settings.

client/settings/game/gameplay-settings.tsx
```diff
@@ -172,39 +176,58 @@ const GameplayRemasteredForm = React.forwardRef<
...
+          showTurnRate: model.showTurnRate,
+        },
+        {
+          onSuccess: () => {},
+          onError: () => {},
+        },
+      ),
+    )
+
+    if (model.visualizeNetworkStalls !== localSettings.visualizeNetworkStalls) {
```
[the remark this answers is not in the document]
2Pacalypse-: Yeah, was just a premature optimization. This is the only settings page that has mixed local/scr settings at the moment, so I thought it's unnecessary to update the local settings every time any of the scr settings change.

Especially since this is a dev-only setting. Which reminds me, maybe I should do this only if `DEV_INDICATOR` is `true`? Since that's the only time we render this setting anyway.

client/settings/settings-sub-page.ts
```diff
@@ -0,0 +1,43 @@
...
+}
+
+export type SettingsSubPage = AppSettingsSubPage | GameSettingsSubPage
+
+export const ALL_SETTINGS_SUB_PAGES: ReadonlyArray<SettingsSubPage> = [
+  ...Object.values(AppSettingsSubPage),
+  ...Object.values(GameSettingsSubPage),
+]
+
+export function settingsSubPageToLabel(subPage: SettingsSubPage, t: TFunction): string {
```
tec27, line 24: Don't we think this is an anti-pattern? This should most likely just be a component somewhere

client/settings/settings.tsx
```diff
@@ -1,228 +1,391 @@
...
+
+            <NavSectionTitle>Game settings</NavSectionTitle>
+            <NavEntry
+              subPage={GameSettingsSubPage.StarCraft}
+              isActive={GameSettingsSubPage.StarCraft === subPage}
+              hasError={!isStarcraftHealthy}
+              onChangeSubPage={onChangeSubPage}
+            />
+
+            {isStarcraftHealthy
```
tec27, line 184: I don't think it's great to hide these completely in this case, they should probably just appear disabled

client/settings/settings.tsx
```diff
@@ -1,228 +1,391 @@
...
+      return false
+    },
+  })
+
+  return (
+    <Container style={style}>
+      <NavContainer>
+        {IS_ELECTRON ? (
+          <>
+            <NavSectionTitle>App settings</NavSectionTitle>
```
tec27, line 164: Would prefer that everything here (titles, all sections, separators) were just a single data structure that gets mapped over instead of repeating this mapping code a bunch of times. This mapping is very error-prone and it is extremely unlikely someone is going to notice in a review when you mess it up, so I don't want it to exist 4+ times in the code

tec27: An alternative is to create a function (or a function that creates a function, since we might need to specify that the rendered entries are disabled) that you pass to the `map` call (and fix the one below where you didn't use `map` to render this thing)

client/settings/settings.tsx
```diff
@@ -1,228 +1,391 @@
...
+  const label = settingsSubPageToLabel(subPage, t)
+
+  // TODO(2Pac): Currently we don't have any non-Electron settings so this might look a bit weird.
+  // Soon enough we should probably have some user settings (e.g. editing user account, and changing
+  // a language).
+
+  switch (subPage) {
+    case AppSettingsSubPage.Sound:
+      return IS_ELECTRON ? (
+        <ContentWrapper title={label} onCloseSettings={onCloseSettings}>
```
tec27, line 348: Don't quite get why we put `ContentWrapper` *inside* this function? Why not just pass the `subPage` to `ContentWrapper` and have it call this function? If you're trying to prevent people from ending up on an Electron page when they're not in Electron I mean... how would that do that in the first place without fucking around with what's in their reducer

client/settings/settings.tsx
```diff
@@ -1,228 +1,391 @@
...
+          <AppSoundSettings />
+        </ContentWrapper>
+      ) : null
+    case AppSettingsSubPage.System:
+      return IS_ELECTRON ? (
```
tec27, line 353: it also sucks to repeat this check over and over again imo, it would be better to make two switch cases (an `assertUnreachable` at the end of the function when there are two different switch cases that cover all options should still work)

tec27: That is:

```ts
switch (subPage) {
  // stuff for everything
}

if (!IS_ELECTRON) {
  throw new Error(`Unknown settings page: ${subPage}`)
}

switch (subPage) {
  // stuff for electron
}

return assertUnreachable(subPage)
```

tec27 (changes requested):

client/settings/settings-sub-page.ts
2Pacalypse-: Hmm, I think this is fine to have as a function; we have plenty of other places where we use something like this (e.g. `matchmakingTypeToLabel`).

The problem of trying to make this into a component is that linter complains if you return only a `string` from a component, so you would have to wrap it in some HTML, I think. And then you would have to style that HTML differently based on where it's used. I guess it's not that big of a deal, but I also don't particularly see what's anti-patterny about this approach either.

client/settings/default-settings.ts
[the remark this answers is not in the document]
2Pacalypse-: Not sure what's the correct behaviour here, but the game seems to work fine by doing that. Our `scr-settings.json` gets created with just the `version` field in it, and the `DEFAULT_SCR_SETTINGS` defined in this file are loaded in the reducer correctly. Changing any of the settings saves all of the settings from that sub-page to our `scr-settings.json` file.

client/settings/settings-sub-page.ts
tec27: Okay let me rephrase: this is an anti-pattern and it should be a component somewhere. The only time we need functions like this is if we need a label in a non-React context, and I really struggle to think of anything where that would be the case. Components such as this should retrieve the `t` themselves.

It is not a problem to need to style things differently, just take a `className` prop and use styled-components as normal.

tec27: For context, all of the functions like this exist because I wrote them before we knew what localization strategy we were going to use, and I wanted consistent strings for various things that we could easily know to localize once we got there. We're past that point now, so we can actually write things that are appropriate for what we're using (which is: write a React component and use the relevant hooks).

tec27 (approved):

PR #893, opened 2023-03-02
Add a hook which tracks the Popover open state.

### LINKED ISSUE #771: Opening the "Filter options" popover in maps browser footer while the "Thumbnail size" popover is already open doesn't work as expected
Namely, the "Filter options" popover is opened, but gets immediately closed along with the "Thumbnail size" popover.

Additionally, trying to open the "Sort maps" popover while either of the other popovers are already open doesn't work at all.

2Pacalypse-:
And use it in all Popover usages.

This hook makes sure that there's only one Popover opened for the same source event.

Also, inadvertently fixes #771 and other places where we used the same open/onDismiss state for different Popovers.

### DIFF
--- a/client/users/user-overlays.tsx
+++ b/client/users/user-overlays.tsx
@@ -1,6 +1,6 @@
 import React, { useCallback, useRef, useState } from 'react'
 import { SbUserId } from '../../common/users/sb-user'
-import { OriginX, OriginY, useAnchorPosition } from '../material/popover'
+import { OriginX, OriginY, useAnchorPosition, usePopoverController } from '../material/popover'
 import { ConnectedUserContextMenuProps, MenuItemCategory } from './user-context-menu'
 import { ConnectedUserProfileOverlayProps } from './user-profile-overlay'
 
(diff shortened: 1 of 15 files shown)


client/users/user-overlays.tsx
```diff
@@ -46,34 +46,35 @@ export function useUserOverlays<E extends HTMLElement = HTMLElement>({
...
-    setProfileOverlayOpen(true)
-  }, [])
-  const onCloseProfileOverlay = useCallback(() => {
-    setProfileOverlayOpen(false)
-  }, [])
   const onClick = useCallback(
-    (e?: React.MouseEvent) => {
-      if (!e || !filterClick || !filterClick(userId, e)) {
-        onOpenProfileOverlay()
+    (e: React.MouseEvent) => {
```
tec27, line 56: Wasn't this optional because not all of these were triggered through an actual click? I feel like you're breaking something with this change

tec27 (approved):

client/users/user-overlays.tsx
2Pacalypse-: Hmm, not sure. I checked all the instances where we use this callback and I couldn't find anything like that.

PR #900, opened 2023-03-13
Add unit tests for chat service.

2Pacalypse-:


server/lib/chat/chat-service.test.ts
```diff
@@ -0,0 +1,1505 @@
...
+    userId: user2.id,
+    userName: user2.name,
+    channelId: testChannel.id,
+    sent: new Date('2023-03-11T00:00:00.000Z'),
+    data: {
+      type: ServerChatMessageType.JoinChannel,
+    },
+  }
+
+  // A helper method that mocks everything correctly to "join" a user to a channel.
```
tec27, line 197: nit: If you're documenting a method, use a doc comment

tec27: other nit: this can just be a normal function, there is no need for it to be an arrow function

server/lib/chat/chat-service.test.ts
```diff
@@ -0,0 +1,1505 @@
...
+    test("should throw if user doesn't exist", async () => {
+      asMockedFunction(findUserById).mockResolvedValue(undefined)
+
+      await expect(
+        chatService.joinInitialChannel(user1.id, dbClient, Promise.resolve()),
+      ).rejects.toThrowErrorMatchingInlineSnapshot(`"User doesn't exist"`)
+    })
+
+    test('works when user exists', async () => {
+      await chatService.joinInitialChannel(user1.id, dbClient, Promise.resolve())
```
tec27, line 264: It would be nice to ensure ordering for the things that can only happen after the transaction has been committed (that is, you should pass a `Deferred` for the promise here, and check that those things only happen after you resolve it)

server/lib/chat/chat-service.test.ts
```diff
@@ -0,0 +1,1505 @@
...
+        user: user1,
+        message: {
+          id: joinUser1ShieldBatteryChannelMessage.msgId,
+          type: joinUser1ShieldBatteryChannelMessage.data.type,
+          channelId: joinUser1ShieldBatteryChannelMessage.channelId,
+          userId: joinUser1ShieldBatteryChannelMessage.userId,
+          time: Number(joinUser1ShieldBatteryChannelMessage.sent),
+        },
+      })
+      expect(nydus.subscribeClient).toHaveBeenCalledWith(
```
tec27, line 284: It would be vastly preferred to check that the client received these messages rather than that `subscribeClient` was called. Checking `subscribeClient` this way means that if we e.g. need to make this data get generated in an async way, every test has to change. Checking `publish` on the client instead means that change would work without test changes.

Apply throughout this file

server/lib/chat/chat-service.test.ts
```diff
@@ -0,0 +1,1505 @@
...
+        client1,
+        getChannelPath(shieldBatteryChannel.id),
+        {
+          action: 'init3',
+          channelInfo: shieldBatteryChannel,
+          activeUserIds: [user2.id, user1.id],
+          selfPermissions: channelPermissions,
+        },
+      )
+      expect(nydus.subscribeClient).toHaveBeenCalledWith(
```
tec27, line 294: This particular check is fine to check on subscribeClient, I think (because there isn't a great way to check this otherwise since it has no initial data)

server/lib/chat/chat-service.test.ts
```diff
@@ -0,0 +1,1505 @@
...
+        user1.id,
+        shieldBatteryChannel.id,
+        { type: ServerChatMessageType.JoinChannel },
+        dbClient,
+      )
+      expect(client2.publish).toHaveBeenCalledWith(getChannelPath(shieldBatteryChannel.id), {
+        action: 'join2',
+        user: user1,
+        message: {
+          id: joinUser1ShieldBatteryChannelMessage.msgId,
```
tec27, line 367: You do this like a million times throughout this file, I'd recommend instead writing a simple method that returns this `message` structure given the DB input, it'd reduce a lot of the noise in these tests

server/lib/chat/chat-service.test.ts
```diff
@@ -0,0 +1,1505 @@
...
+        },
+      )
+      expect(nydus.subscribeClient).toHaveBeenCalledWith(
+        client1,
+        getChannelUserPath(shieldBatteryChannel.id, user1.id),
+        undefined,
+      )
+    })
+
+    test("works when channel doesn't exist", async () => {
```
tec27, line 391: nit: avoid `works` when there is some more descriptive phrasing. Here that'd probably be `creates a new channel when it doesn't exist`

server/lib/chat/chat-service.test.ts
```diff
@@ -0,0 +1,1505 @@
...
+
+        await chatService.leaveChannel(shieldBatteryChannel.id, user1.id)
+
+        expect(removeUserFromChannelMock).toHaveBeenCalledWith(user1.id, shieldBatteryChannel.id)
+        expect(client2.publish).toHaveBeenCalledWith(getChannelPath(shieldBatteryChannel.id), {
+          action: 'leave2',
+          userId: user1.id,
+          newOwnerId: undefined,
+        })
+        expect(nydus.unsubscribeClient).toHaveBeenCalledWith(
```
tec27, line 472: Would prefer triggering something that publishes to the channel (like sending a message) and checking that the client did not receive it

server/lib/chat/chat-service.test.ts
```diff
@@ -0,0 +1,1505 @@
...
+    id: makeSbChannelId(2),
+    name: 'test',
+    private: false,
+    official: false,
+  }
+
+  const user1: SbUser = { id: makeSbUserId(1), name: 'USER_NAME_1' }
+  const user2: SbUser = { id: makeSbUserId(2), name: 'USER_NAME_2' }
+
+  const userPermissions: SbPermissions = {
```
tec27, line 111: Move this value to the common permissions file and call it `DEFAULT_PERMISSIONS`. We already have like 10 files that have to be updated whenever a new permission is added, I don't want another

(That value should be readonly in TS, so this would become `const userPermissions = { ...DEFAULT_PERMISSIONS }` if you need this as an actual value somewhere)

tec27 (changes requested):

server/lib/chat/chat-service.test.ts
2Pacalypse-: I also forgot to `expect` that the `createChannel` was actually called in this test, so thanks for that :d

server/lib/chat/chat-service.test.ts
tec27: Still too much coupling to the actual mechanics versus testing that the state is correct. Ideally our tests wouldn't need to know anything about the paths being used at all, but unfortunately they're significant to the behavior (and not really avoidable unless you hooked up the client code and examined the reducer state or something)

common/users/permissions.ts
```diff
@@ -13,3 +13,19 @@ export interface SbPermissions {
...
+  manageMaps: false,
+  manageMapPools: false,
+  manageMatchmakingSeasons: false,
+  manageMatchmakingTimes: false,
+  manageRallyPointServers: false,
+  massDeleteMaps: false,
+  moderateChatChannels: false,
+}
+
+export const userPermissions = { ...DEFAULT_PERMISSIONS }
```
tec27, line 31: huh? why is this here now?

tec27: And when you move this back into the test file where it belongs, do *NOT* put this at the top of the file. When you do that you're contaminating every test if you modify it.

The point of `DEFAULT_PERMISSIONS` being readonly was to make sure that never happens, if you immediately turn it into a top-level non-const thing you have defeated the whole purpose

server/lib/chat/chat-service.test.ts
```diff
@@ -87,6 +89,37 @@ jest.mock('./chat-models', () => ({
   searchChannelsAsAdmin: jest.fn().mockResolvedValue([]),
 }))
 
+// TODO(2Pac): Move these elsewhere?
+
+type DbJoinChannelMessage = ChatMessage & { data: JoinChannelData }
```
tec27, line 94: Call these like `FakeDbFoo`, calling them `DbJoinChannelMessage` is confusing since that's how we name stuff in our model files and these are not what those are

server/lib/chat/chat-service.test.ts
```diff
@@ -270,27 +294,20 @@ describe('chat/chat-service', () => {
...
-        message: {
-          id: joinUser1ShieldBatteryChannelMessage.msgId,
-          type: joinUser1ShieldBatteryChannelMessage.data.type,
-          channelId: joinUser1ShieldBatteryChannelMessage.channelId,
-          userId: joinUser1ShieldBatteryChannelMessage.userId,
-          time: Number(joinUser1ShieldBatteryChannelMessage.sent),
-        },
+        message: toJoinChannelMessageJson(joinUser1ShieldBatteryChannelMessage),
+      })
+      expect(client1.publish).toHaveBeenCalledWith(getChannelPath(shieldBatteryChannel.id), {
```
tec27, line 305: This doesn't verify anything new from the last version you had. You need to clear the mock before resolving the deferred or they'll still have the previous list of calls, and thus a call before or after the deferred being resolved makes zero difference to this succeeding

server/lib/chat/chat-service.test.ts
```diff
@@ -270,27 +294,20 @@ describe('chat/chat-service', () => {
         { type: ServerChatMessageType.JoinChannel },
         dbClient,
       )
+
+      await transactionPromise.resolve()
```
tec27, line 298: `Deferred.resolve` is not async so this await is very strange. I think you probably *do* need to `await` a promise to ensure that async things are completed, but you should split that:

```ts
transactionPromise.resolve()
await Promise.resolve()
```

server/lib/chat/chat-service.test.ts
```diff
@@ -921,14 +914,7 @@ describe('chat/chat-service', () => {
...
-          message: {
-            id: textMessage.msgId,
-            type: textMessage.data.type,
-            channelId: textMessage.channelId,
-            from: textMessage.userId,
-            time: Number(textMessage.sent),
-            text: processedMessageString,
-          },
+          message: toTextMessageJson(textMessage),
           user: {
```
tec27, line 918: Dunno how many times you do this throughout but is it really necessary to split these fields out? can't you just do `user: user1`? It'd be kind of vastly preferable to do that because if/when we add more fields to `SbUser` such tests will all have to be updated, not something that we really want to have to do

tec27 (changes requested):

server/lib/chat/chat-service.test.ts
tec27: Would generally encourage you to make sure tests actually fail when they do the wrong thing, otherwise we end up with a super false sense of protection. Tests that do not test the thing they say they do are actually worse than no tests at all imo

server/lib/chat/chat-service.test.ts
```diff
@@ -232,6 +231,62 @@ describe('chat/chat-service', () => {
...
+    client: InspectableNydusClient,
+  ) {
+    mockTextMessage(user, channel, 'Hello World!', [])
+
+    await chatService.sendChatMessage(channel.id, user.id, textMessage.data.text)
+
+    expect(client.publish).not.toHaveBeenCalledWith(getChannelPath(channel.id), {
+      action: 'message2',
+      message: toTextMessageJson(textMessage),
+      user: {
```
tec27, line 282: You've reproduced the thing I asked you to fix elsewhere

tec27 (approved):

server/lib/chat/chat-service.test.ts
2Pacalypse-: Interesting :)

PR #906, opened 2023-03-17
Add a batch retrieval of chat channels in ConnectedChannelName.

2Pacalypse-:
Also adds a batch retrieval of user infos in ConnectedUsername.

### DIFF
--- a/server/lib/chat/chat-api.ts
+++ b/server/lib/chat/chat-api.ts
@@ -319,14 +319,6 @@ export class ChatApi {
     ctx.status = 204
   }
 
-  @httpGet('/:channelId')
-  @httpBefore(throttleMiddleware(retrievalThrottle, ctx => String(ctx.session!.userId)))
-  async getChannelInfo(ctx: RouterContext): Promise<ChannelInfo> {
-    const channelId = getValidatedChannelId(ctx)
-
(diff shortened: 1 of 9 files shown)


server/lib/chat/chat-api.ts
```diff
@@ -326,6 +318,28 @@ export class ChatApi {
 
     ctx.status = 204
   }
+
+  @httpGet('/:channelId/info')
```
tec27, line 322: It seems like we don't have a `GET` for `/:channelId`, this should probably just be that?

tec27 (changes requested):

server/lib/chat/chat-api.ts
2Pacalypse-: When I make this into just a `/:channelId` route, then all the requests to `/batch-info` get intercepted by this route :/
Maybe that one should be renamed to something else too?

tec27: `batch-info` needs to come before the `/:channelId` route so that the router hits it first. In general you should always order your routes from constants -> tokens

tec27: You could probably also do an explicit match for `channelId`'s format if you want to be extra safe, I think like:

`/:channelId(\\d+)`

PR #907, opened 2023-03-18
Add a new (better) version of the channel info card.

2Pacalypse-:
Mostly plagiarized from the league stuff ><

I haven't actually re-used anything from the league components, as that just seems like a cause for a headache when those designs inevitably diverge.

However, maybe *some* things could be re-used? The badge stuff in particular is like 99% the same, and could perhaps be made into a more general thing. The only problem is I don't know what to call and where to put that more general thing..


2Pacalypse-:
Depends on #906

client/chat/channel-info-card.tsx
```diff
@@ -105,29 +162,22 @@ export function ConnectedChannelInfoCard({
...
-    subtitle = (
+    descriptionText = (
       <ErrorText>
         This channel could not be found. It might not exist, or it may have been re-created by
         someone else.
       </ErrorText>
     )
   } else if (channelInfo?.private && !isUserInChannel) {
-    subtitle = <ErrorText>This channel is private and requires an invite to join.</ErrorText>
+    descriptionText = <ErrorText>This channel is private and requires an invite to join.</ErrorText>
```
tec27, line 174: I'm not sure that either of these two "Errors" really make sense to display in place of the description? It seems like they should just change the Card action (and maybe this text would display in a tooltip for the action button? unsure)

common/chat.ts
```diff
@@ -127,6 +132,12 @@ export interface ChannelInfo {
   id: SbChannelId
   /** The name of the chat channel. */
   name: string
+  /** The description of the chat channel. */
```
tec27, line 135: If these don't exist in the model and you're not adding them here, you should at least file an issue to do that

tec27 (changes requested):

client/chat/channel-info-card.tsx
2Pacalypse-: Yeah I wasn't sure either. For this error specifically, I thought it made sense to display this error in place of the description, since we don't want to display descriptions of private channels, I think? And displaying the default "No channel description" text seemed a bit weird. This is how the card looks like currently for private channels:

image

For the banned error message, I first display the channel info card normally (which means banned users will be able to see channel banner/badge/description/user count), and when they click the "Join" button the card displays the error message:

image

But yeah, now that I think about it, this might not be the best way to deal with that. I think it makes sense to think about which channel information we want to show to the user when they first open the card (for private channels that's none?), and then how to display the error message (for banned users they only realized they're banned _after_ they clicked the "Join" button).

common/chat.ts
2Pacalypse-: This made me think, should this information be available under `joinedChannelData` instead? In other words, should people be able to see banner images, badge images, and description of private channels they're not in?

client/chat/channel-info-card.tsx
tec27: I think if you're not showing descriptions of private channels, then you should have some kind of content there saying it's private. It's not an error that it's private, and thus treating it like an error makes no sense. A better thing you have would have like a lock icon and text saying it's private. In general, don't use error-colored text treatments for things that are not errors.

As far as errors for channels you are banned from, I think replacing the description is weird, and it would be better if that error goes somewhere else (honestly it may make more sense as a dialog, using this popover as a modal for joining channels and surface for displaying the outcome of that feels incredibly weird to me)

common/chat.ts
tec27: Yes they should be able to? I mean how are you planning to make any kind of useful channel directory if you treat all of that data as private to people that have joined the channel?

client/chat/channel-info-card.tsx
tec27: Related to that last point about it being weird to use this as the surface for displaying results of this operation, handling the navigation in `onJoinClick`'s success handler doesn't really work. That should really be handled via the action creator itself, and you should treat any callbacks you have on `abortableThunk`s as valid only as long as the component is mounted. We want to navigate to the newly joined channel whether or not this popover is still up, so it doesn't make sense to do that behavior (or display errors, or whatever else) inside the callback from the `abortableThunk`.

common/chat.ts
2Pacalypse-: I was thinking of excluding private channels from the search results of chat channels. My thinking was that you'd need an invite to join a private channel anyway, so wasn't sure what was the point of showing it in the search results.

Also, `joinedChannelData` is returned for people who have joined a private chat channel, so they should still be able to see this information.

tec27: I don't really get what you're saying? You're saying you want to move `description` to a piece of data that is only returned once you have joined a channel, but we want `description` to be available in search results (and in the cards you are working on in this PR). So it makes no sense to move it there? I don't get what the private channel considerations have to do with this at all, other than that it needs to be optional as a result

2Pacalypse-: Yeah, I guess I haven't expressed myself all too clearly :d.

My thinking was that we should _not_ display `description` (and these other fields) for private channels (we can instead show a piece of content indicating the channel is private that you suggested above). This doesn't matter all too much for search results because I thought we were gonna exclude private channels from search results anyway. It only matters if someone mentions a private channel in chat where anyone can click on it and see its info card.

I guess I was confusingly thinking of `joinedChannelData` as "private" data, i.e. data that is returned for `private` channels as well, but only if the user has joined the channel. This would mean that the user who has joined the channel would see the full info in the channel card, and those who have not joined the channel would only see some limited information.

tec27: No, the way that would be handled is that `description` would not be filtered out for users that were in a private channel. I don't see what this has to do with JoinedChannelData at all

2Pacalypse-: Well, I thought all "private" channel data, i.e. data that we don't want users to see unless they're joined in a channel, would go into the `JoinedChannelData` structure. I was unsure if `description` should be treated like that.

2Pacalypse-: Hmm, I think I finally got what you mean. And my last reply here is wrong, since we already have the `userCount` field which fits that description, but is not inside the `JoinedChannelData` structure. I assume you were saying `description` could be like that?

I think the problem lies in ambiguity of `JoinedChannelData` structure, and I think we could probably remove that structure altogether. I _think_ what I initially meant by `JoinedChannelData` structure is to represent data that is only returned when the user is subscribed to the channel. But this seems weird to separate data in structure based on APIs.

My new thinking is that we could have a flat `ChatInfo` type, where all the data that is not returned for private channels is simply optional, e.g.:
```ts
interface ChannelInfo {
  id: SbChannelId
  name: string
  private: boolean
  official: boolean
  description?: string
  topic?: string
  ownerId?: SbUserId
  bannerPath?: string
  badgePath?: string
  userCount?: number
}
```
So `GetChannelInfo(s)` would return either only required fields for private channels, or all the fields for non-private channels (and private channels that user is joined in).
And subscribing user to a channel would give them the full channel info of course (although weirdly enough without `userCount` atm, since we compute that inline in `GetChannelInfo` method).

I'm not sure if this is better since there's still some weirdness here (e.g. `ownerId` being `undefined` can mean the channel is private _or_ the channel is official and simply doesn't have an owner; although that field will never be used to determine that since we have separate flags for that).

tec27: How is `JoinedChannelData` ambiguous? `JoinedChannelData` is `Data` that you only get for `Channel`s that you have `Joined`. The only way it is ambiguous is that you are trying to put stuff in it now that is *not* that (i.e. `description`, which is very much something you need for channels you have *not* joined).

Please just leave these things how they are and finish this PR instead of trying to endlessly re-architect these things (and either populate the `description` field properly or file an issue for it, like the original comment in this thread was talking about).

client/chat/channel-info-card.tsx
```diff
@@ -1,58 +1,114 @@
...
+  display: -webkit-box;
+  -webkit-box-orient: vertical;
+  line-clamp: 3;
+  -webkit-line-clamp: 3;
+  overflow: hidden;
+  text-overflow: ellipsis;
+`
+
+const PrivateChannelIcon = styled(MaterialIcon).attrs({ icon: 'lock' })`
+  vertical-align: bottom;
```
tec27, line 78: I'm not totally sure what visual treatment you're trying to go for here, but I struggle to imagine a case where "bottom" is the correct value. Seems more likely you want `text-top` or `baseline`?

tec27: And whatever value you choose here, you should look how it behaves when the text it is beside extends across multiple lines, because I think this will surprise you

client/chat/channel-info-card.tsx
```diff
@@ -92,7 +135,7 @@ export function ConnectedChannelInfoCard({
     setIsJoinInProgress(true)
     dispatch(
       joinChannel(channelName, {
-        onSuccess: channel => navigateToChannel(channel.id, channel.name),
+        onSuccess: () => {},
         onError: err => {
           setJoinChannelError(err)
```
tec27, line 140: You need to do something with this, putting it in state and then not looking at that state isn't helping anybody (but again, we have no guarantees that this component is even still mounted when this gets called, so using the card as a surface for displaying any error is a bad idea)

client/chat/channel-info-card.tsx
```diff
@@ -101,34 +144,21 @@ export function ConnectedChannelInfoCard({
...
+  let descriptionText
+  if (channelInfo?.private && !isUserInChannel) {
+    descriptionText = (
+      <NoChannelDescriptionText>
+        <PrivateChannelIcon />
+        This channel is private and requires an invite to join.
+      </NoChannelDescriptionText>
     )
+  } else if (channelInfo?.description) {
+    descriptionText = channelInfo.description
```
tec27, line 159: You should probably put this in a `span` so it mirrors the dom structure of the other cases

client/chat/channel-info-card.tsx
```diff
@@ -101,34 +144,21 @@ export function ConnectedChannelInfoCard({
...
+  } else {
+    descriptionText = <NoChannelDescriptionText>No channel description.</NoChannelDescriptionText>
```
tec27, line 161: This doesn't feel like a very user-focused message, maybe "No description provided." or "This channel has no description." (which is similar in structure to your private message)

client/chat/channel-info-card.tsx
```diff
@@ -1,58 +1,114 @@
...
   display: flex;
-  flex-direction: row;
-  align-items: center;
+  flex-direction: column;
+
+  contain: content;
 `
 
 const StyledLoadingDotsArea = styled(LoadingDotsArea)`
   width: 100%;
```
tec27, line 31: This doesn't do anything, please just use the unstyled version

client/icons/material/chat-40px.svg
tec27, line 1: You moved to the icon font and should delete this now, shouldn't you?

tec27 (changes requested):

client/chat/channel-banner.tsx
```diff
@@ -0,0 +1,62 @@
...
+    <ChannelBannerContainer>
+      <img src={src} alt='' draggable={false} />
+    </ChannelBannerContainer>
+  )
+}
+
+export const ChannelBannerPlaceholderIcon = styled(MaterialIcon).attrs({ icon: 'chat' })`
+  width: 22.727272%;
+  height: auto;
+  font-size: 100%;
```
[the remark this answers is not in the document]
tec27: Think the answer is probably "round the size you pass to MaterialIcon". Using non-whole-number font sizes is probably just a bad idea for pixel alignment generally, but I don't know that MaterialIcon should be forcing that itself.

client/chat/channel-info-card.tsx
2Pacalypse-: Hmm, that's what I thought to. But for some reason it only looked decent to me with `vertical-align: bottom`. Here are some screenshots to show what I mean.

`vertical-align: bottom`:
image

`vertical-align: baseline`:
image

`vertical-align: text-top`:
image

client/chat/channel-info-card.tsx
2Pacalypse-: I am looking at this, but the only thing I'm using it in this card is to disable the "Join" button if you clicked on it and you get an error that you're banned (the snackbar displays the actual error from action-creators).

I initially thought to just close the popover whenever you click the "Join" button, but this component doesn't assume it's rendered in a Popover (and it actually isn't when it's rendered on a channel page).

client/chat/channel-info-card.tsx
2Pacalypse-: I think the issue is that I'm trying to fit a 24px icon in a same line with a text which has a 20px line-height. But even when I increase the combined line-height to 24px, I feel like `vertical-align: bottom` still looks the best for some reason.

tec27: You could just use a smaller icon, it is very easy to pick one that matches your line height now :) But I don't think that looks good regardless, having the text flow around the icon is very much not what we want. I think it would probably look better to use a larger icon, center it, and put this text below.

If that for some reason doesn't work out, you'd probably just want to use a flex layout and put the icon and the text in separate blocks.

client/chat/channel-info-card.tsx
tec27: I mean that is probably an indication that it should be passed a (perhaps optional) event handler and you can do that. At the very least, I will again state that you need to do something with this error. You should not really assume it is the singular error you "expect". If all types of errors are handled elsewhere that's fine, but in that case the proper thing to do would be to note that down here, and set boolean state for whether to disable the button. What you've written here gives the impression that the error will be handled by something when you're just using it as a flag, and that's dangerous if people come along and change the place you *expect* it to have been handled.

client/chat/channel-info-card.tsx
tec27: Yeah, I think using centered text for it might look better, but otherwise fine, The padding between the icon and the channel name also looks a tad large?

client/chat/channel-info-card.tsx
```diff
@@ -92,42 +138,35 @@ export function ConnectedChannelInfoCard({
     setIsJoinInProgress(true)
     dispatch(
       joinChannel(channelName, {
-        onSuccess: channel => navigateToChannel(channel.id, channel.name),
+        onSuccess: () => {},
         onError: err => {
-          setJoinChannelError(err)
           setIsJoinInProgress(false)
+
+          // NOTE(2Pac): We assume the rest of the errors are handled elsewhere since we only care
```
tec27, line 145: `We assume the error has been handled in <ACTUAL PLACE IT IS BEING HANDLED>, we just need to deal with the ban case in this UI`

tec27 (approved):

PR #908, opened 2023-03-18
Add channel mentions.

2Pacalypse-:
Channel can be mentioned by writing the # character preceding the name of the channel. All messaging-related services (chat channels, whispers, lobbies, parties) support this.

Once an existing channel is mentioned, its name will become clickable and will open the channel-info-card when clicked.


tec27 (changes requested):

server/lib/chat/chat-models.ts
```diff
@@ -177,16 +192,16 @@ export interface ChatMessage {
   data: ChatMessageData
 }
 
-type DbChatMessage = Dbify<ChatMessage>
+type DbifiedChatMessage = Dbify<DbChatMessage>
```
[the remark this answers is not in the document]
2Pacalypse-: Yeah, I wasn't sure about changing this either, but I was just a bit annoyed that we have a `ChatMessage` type that we use in the database, and a `ChatMessage` type that we use everywhere else, so importing those type always gave me a bit of a pause on which one I need to import.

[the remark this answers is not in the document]
tec27: Then come up with a different name for one of them that doesn't conflict with the naming scheme of every other model in our app?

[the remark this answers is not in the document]
2Pacalypse-: I'll leave that for another PR :d. This was a lazy attempt at introducing a new terminology anyway, sorry about that.

server/lib/chat/chat-service.test.ts
```diff
@@ -1314,6 +1382,47 @@ describe('chat/chat-service', () => {
...
+        channelId: testChannel.id,
+        sent: new Date('2023-03-11T00:00:00.000Z'),
+        data: {
+          type: ServerChatMessageType.TextMessage,
+          text: `Join <#${shieldBatteryChannel.id}> <#${testChannel.id}> and <#${DELETED_ID}>.`,
+          channelMentions: [shieldBatteryChannel.id, testChannel.id, DELETED_ID],
+        },
+      }
+
+      asMockedFunction(getMessagesForChannel).mockResolvedValueOnce([message])
```
tec27, line 1408: I'm sure you do this elsewhere in this test, but `mockResolvedValueOnce` is extremely brittle, you're writing tests that depend VERY HEAVILY on the ordering and amount of calls to a method, i.e. they are directly dependent on the code they are testing not changing. I think you'd be better off building a "fake DB" of users for the `findUsersById` call (I don't know how many unique users you have in your tests here but I imagine it's like less than 5, this would be a very easy mock to write.

The other stuff here, I'm pretty sure you could just use `mockResolvedValue` for

server/lib/chat/chat-service.ts
```diff
@@ -589,20 +596,31 @@ export default class ChatService {
...
     }
 
-    const mentions = await findUsersById(Array.from(mentionIds))
+    const [users, userMentions, channelMentions] = await Promise.all([
+      findUsersById(Array.from(userIds)),
+      findUsersById(Array.from(userMentionIds)),
+      this.getChannelInfos(Array.from(channelMentionIds), userId),
+    ])
+
+    const deletedChannels = subtract(
```
tec27, line 613: Given the general case is "no channels are deleted", I would probably write this as:
```ts
const deletedChannels = channelMentionIds.length === channelMentions.size ? [] : Array.from(subtract(channelMentionIds, channelMentions.map(c => c.id)))
```

No sense in iterating the whole set/map if we already know it's all there

server/lib/parties/party-service.test.ts
```diff
@@ -140,8 +141,7 @@ describe('parties/party-service', () => {
     webClient = connector.connectClient(webUser, WEB_USER_CLIENT_ID, 'web')
 
     clearTestLogs(nydus)
-    findUsersByNameMock.mockClear()
-    findUsersByIdAsMapMock.mockClear()
+    jest.clearAllMocks()
```
tec27, line 144: This is not behaviorally the same as what was happening previously, if you need to fix something specific then please just add in the mocks you want to clear. Blindly clearing every mock in the system makes it impossible to compose different test utilities safely

server/lib/users/user-model.ts
```diff
@@ -341,22 +341,29 @@ export async function findUserByName(name: string): Promise<SbUser | undefined>
...
 
 /**
- * Returns a `Map` of name -> `User` for a given list of names. Any names that can't be found won't
- * be present in the resulting `Map`.
+ * Returns the data for all users with the specified names. If a user cannot be found it will not
+ * be included in the result. The order of the result is not guaranteed.
  */
-export async function findUsersByName(names: string[]): Promise<Map<string, SbUser>> {
-  // TODO(tec27): Should this just return an array and let callers make the Map if they want?
+export async function findUsersByName(names: string[]): Promise<SbUser[]> {
```
tec27, line 347: In the future I would vastly, vastly, vastly prefer that you make this change in a separate PR and rebase your code on top of it, doing this here makes all of this much harder to review

tec27 (changes requested):

client/chat/chat-reducer.ts
```diff
@@ -164,6 +167,22 @@ function updateMessages(
   channelMessages.hasHistory = channelMessages.hasHistory || sliced
 }
 
+function updateChannels(state: ChatState, channels: ChannelInfo[]) {
+  for (const channel of channels) {
+    if (!state.idToInfo.has(channel.id)) {
```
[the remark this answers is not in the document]
2Pacalypse-: Just making sure :)

server/lib/chat/chat-service.test.ts
2Pacalypse-: This is actually the one place in this test file where I used `mockResolvedValueOnce` by mistake :d

I just checked all my other usages of it, and most of them (all but one) are used to mock the function that gets called twice in a same method, e.g. mocking `getUserChannelEntryForUser` (this is the majority of its usages) in multiple methods that call it like this:
```ts
const [channelInfo, userChannelEntry, targetChannelEntry] = await Promise.all([
  getChannelInfo(channelId),
  getUserChannelEntryForUser(userId, channelId),
  getUserChannelEntryForUser(targetId, channelId),
])
```

Similarly for `findUsersById`:
```ts
const [users, userMentions, channelMentions] = await Promise.all([
  findUsersById(Array.from(userIds)),
  findUsersById(Array.from(userMentionIds)),
  this.getChannelInfos(Array.from(channelMentionIds), userId),
])
```

I was hesitant to use `mockResolvedValueOnce` myself because I didn't like how it relied on stack call order, but I thought it would be fine here since these calls are right next to each other, and Jest documentation says this is the easiest way to mock a (different) return value of a function that was called multiple times. Also, I think it shows the intent quite nicely. For example, when I want to test that the target user is a channel moderator, I just use it like this:
```ts
asMockedFunction(getUserChannelEntryForUser)
  .mockResolvedValueOnce(user1TestChannelEntry)
  .mockResolvedValueOnce({
    ...user2TestChannelEntry,
    channelPermissions: { ...channelPermissions, editPermissions: true },
})
```
I'm not opposed to creating a fake DB(s?) to replace all of the `mockResolvedOnce` usage, but I feel like it would take me some time (not because it's necessarily complicated, but because I don't know how to do it atm :d).

server/lib/parties/party-service.test.ts
2Pacalypse-: Yeah, I did try to at least make sure I didn't break anything with this change :d

I thought that clearing all mocks before each test would be a good testing practice to adopt, but I'm not sure where I read that now ^^. Reverted the change for now.

server/lib/chat/chat-service.test.ts
tec27: I think in those it would generally be better to just write a simple implementation. When I was talking about a fake, to be clear, I was talking about one that would be read-only, which is significantly easier to implement correctly than the other way (but you'd only be using this for data the code depends on that it never actually changes).

For the thing I mentioned in particular, it is absurdly easy to do, and I've already done it in other tests (e.g. see the game loader branch), it's just not easy to share an implementation between tests due to Jest's rules about mocks not accessing external data.

For the stuff you do expect to change (e.g. the last example you have here), I think the better thing to do is just write a simple implementation in the test, where you've written the `mockResolvedValueOnce` now. e.g.

```ts
asMockedFunction(getChannelEntryForUser).mockImplementation(async (channelId, userId) => {
  if (userId === USER_1) {
    return user1TestChannelEntry
  } else if (userId == USER_2) {
    return user2TextChannelEntry
  } else {
    throw new Error('Unexpected user lookup')
  }
})
```

server/lib/parties/party-service.test.ts
tec27: I don't think this is as bad as I thought it was (I thought this was doing the thing that `mockReset` turns out to do, that is, clearing mocked implementation/values as well), but I would still avoid doing this at a blanket level. It's not possible to opt out of this should you need to, other than by going through and finding every mock in the test that you need to reset. It's much better to be prescriptive when you're writing the test, "I know specifically this mock has calls I want to ignore/drop at this time". `clearAllMocks` affects every mock everywhere, so if we add e.g. a test utility that maintains its own mock and provides assertions against that mock, you might be screwing up its functionality and it will absolutely not be obvious when those tests fail where the cause lies.

client/chat/chat-reducer.ts
```diff
@@ -164,9 +172,26 @@ function updateMessages(
...
+    state.idToInfo.set(channel.id, channel)
+    if (state.deletedChannels.has(channel.id)) {
+      state.deletedChannels.delete(channel.id)
+    }
+  }
+}
+
+function updateDeletedChannels(state: ChatState, deletedChannels: SbChannelId[]) {
+  for (const channelId of deletedChannels) {
+    if (!state.deletedChannels.has(channelId)) {
```
tec27, line 186: You know how sets work, right? :) Just call add

client/chat/chat-reducer.ts
```diff
@@ -164,9 +172,26 @@ function updateMessages(
   channelMessages.hasHistory = channelMessages.hasHistory || sliced
 }
 
+function updateChannels(state: ChatState, channels: ChannelInfo[]) {
+  for (const channel of channels) {
+    state.idToInfo.set(channel.id, channel)
+    if (state.deletedChannels.has(channel.id)) {
```
tec27, line 178: Just call delete, this is 2 lookups for something that can be done in 1

client/chat/chat-reducer.ts
```diff
@@ -33,33 +34,39 @@ export interface MessagesState {
...
+  /** A set of joined chat channels that are unread */
   unreadChannels: Set<SbChannelId>
+  /** A set of channel IDs saved in various chat messages that no longer exist. */
+  deletedChannels: Set<SbChannelId>
 }
 
 const DEFAULT_CHAT_STATE: Immutable<ChatState> = {
   joinedChannels: new Set(),
   idToInfo: new Map(),
+  idToJoinedData: new Map(),
```
tec27, line 62: Not an issue for this actual line, but there's an event for leaving channels, I think? You need to use that to delete the data for that channel from this map

server/lib/chat/chat-service.ts
```diff
@@ -490,33 +497,48 @@ export default class ChatService {
...
+    // TODO(2Pac): Get this from DB instead.
+    const isUserInChannel = this.state.users.get(userId)?.has(channelId)
 
     if (!channelInfo) {
       throw new ChatServiceError(ChatServiceErrorCode.ChannelNotFound, 'Channel not found')
     }
 
     let userCount
-    if (!channelInfo.private || this.state.users.get(userId)?.has(channelId)) {
+    if (!channelInfo.private || isUserInChannel) {
```
tec27, line 510: nit: I think all of this would be more clear as a ternary within the return now

If userCount is *also* private (like description) it might be more worthwhile to make the 3rd struct we had talked about, I don't see how you're going to easily keep from overwriting the userCount in other requests otherwise so private channels will have bugs

server/lib/chat/chat-service.ts
```diff
@@ -589,20 +614,36 @@ export default class ChatService {
...
         default:
           return assertUnreachable(msg.data)
       }
     }
 
-    const mentions = await findUsersById(Array.from(mentionIds))
+    const [users, userMentions, channelMentions] = await Promise.all([
+      findUsersById(Array.from(userIds)),
+      findUsersById(Array.from(userMentionIds)),
+      this.getChannelInfos(Array.from(channelMentionIds), userId),
```
tec27, line 628: imo this should return the same data as if the messages were sent 1-by-1, or you're opening the door to weird bugs that will be hard to track down and potentially making this hard to cache in the future (that is, this should NOT include user-specific information)

server/lib/chat/chat-service.ts
```diff
@@ -767,12 +808,22 @@ export default class ChatService {
...
   private subscribeUserToChannel(
     userSockets: UserSocketsGroup,
-    channelInfo: ChannelInfo,
+    channelInfo: FullChannelInfo,
     channelPermissions: ChannelPermissions,
   ) {
     userSockets.subscribe<ChatInitEvent>(getChannelPath(channelInfo.id), () => ({
       action: 'init3',
-      channelInfo,
+      channelInfo: {
```
tec27, line 816: This code is all broken, you need to be VERY careful about stuff like this. You have subscribed a user to the version of the channel that existed when they connected to the server. This object will NEVER be updated. 

You need to write an async method here that retrieves this on demand, passing in the existing object to the subscription does not work

tec27 (changes requested):

server/lib/chat/chat-service.ts
tec27: Yeah, imo you should move to a very simple `ChannelInfo` that is stuff that is always present (name, mostly) and is what is given as dat for channel mentions, another struct for things that are visible only for public + joined channels (ProtectedChannelData? SensitiveChannelData? I dunno, something like that). Might pay to write a db query that returns the small ChannelInfo directly as well, since that's what we'd be querying in bulk and it can be returned using only indexes, but it's a bit of a premature optimization maybe

2Pacalypse-:
Still need to:
- update API route versions where needed (not sure yet which routes need to change)
- finish the TODO about adding something to FakeNydusServer to resolve when all current subscription promises are complete

server/lib/chat/chat-service.ts
```diff
@@ -280,15 +283,35 @@ export default class ChatService {
...
+        ownerId: channel.ownerId,
+        topic: channel.topic,
+      },
+    }
+
     if (isUserInChannel) {
-      return channel!
+      return joinChannelResponse
     } else if (isUserBanned) {
       throw new ChatServiceError(ChatServiceErrorCode.UserBanned, 'User is banned')
```
tec27, line 309: imo this check/throw should be above constructing `joinChannelResponse` now, probably just check it right after the loop

tec27: Could probably restructure the code following this as well to avoid needing to construct this object before the return entirely:

```ts
if (!isUserInChannel) {
  this.updateUserAfterJoining(/* ... */)
}

return {
  // ...
}
```

server/lib/chat/chat-service.ts
```diff
@@ -490,41 +519,48 @@ export default class ChatService {
     })
   }
 
-  async getChannelInfo(channelId: SbChannelId, userId: SbUserId): Promise<ChannelInfo> {
+  async getChannelInfo(channelId: SbChannelId, userId: SbUserId): Promise<GetChannelInfoResponse> {
     const channelInfo = await getChannelInfo(channelId)
+    // TODO(2Pac): Get this from DB instead.
+    const isUserInChannel = this.state.users.get(userId)?.has(channelId)
```
tec27, line 525: Can you please just write this now instead of writing a TODO here, this isn't a hard query

server/lib/chat/chat-service.ts
```diff
@@ -490,41 +519,48 @@ export default class ChatService {
...
-      id: channelInfo.id,
-      name: channelInfo.name,
-      private: channelInfo.private,
-      official: channelInfo.official,
-      userCount,
+      basicChannelInfo: {
+        id: channelInfo.id,
+        name: channelInfo.name,
+      },
+      detailedChannelInfo: {
```
tec27, line 536: As stated in the common file, this should either be all present or not present at all. Should probably also write a TODO here for the issue you filed about actually filling out the other fields (because this code does not make it clear that it has skipped a bunch of optional ones)

server/lib/chat/chat-service.ts
```diff
@@ -490,41 +519,48 @@ export default class ChatService {
...
+  ): Promise<GetChannelInfosResponse> {
     const channelInfos = await getChannelInfos(channelIds)
 
     return channelInfos.map(channel => {
-      let userCount
-      if (!channel.private || this.state.users.get(userId)?.has(channel.id)) {
-        userCount = channel.userCount
-      }
+      // TODO(2Pac): Get this from DB instead (make sure it's not called inside this loop tho :d)
+      const isUserInChannel = this.state.users.get(userId)?.has(channel.id)
```
tec27, line 552: Same as above, please just write the query for this

tec27 (changes requested):

tec27:
> Still need to:
> 
> * update API route versions where needed (not sure yet which routes need to change)

Provided you make the one change I suggested in the `common` file (undo the rename to `basicChannelInfo` in the init struct), I don't think there are any routes that need to be versioned.

> * finish the TODO about adding something to FakeNydusServer to resolve when all current subscription promises are complete

You can just do this in a separate PR and fix all the places that have this TODO, provided that your tests actually work with the current code. If your tests don't work with the `setTimeout` then something else is probably also wrong.

common/chat.ts
```diff
@@ -390,3 +420,28 @@ export interface UpdateChannelUserPermissionsRequest {
...
 }
+
+/**
+ * An object including the basic channel info and detailed channel info.
+ */
+interface ChannelDetails {
+  /** The basic information about the channel. */
+  basicChannelInfo: BasicChannelInfo
+  /** The detailed information about the channel. */
+  detailedChannelInfo: DetailedChannelInfo
```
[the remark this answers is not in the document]
2Pacalypse-: Yeah, the only reason why I made it required here is because it contained `private` and `official` flags which had to be returned. By moving those flags to basic channel info, then I can indeed make this optional.

But also, I would just like to clarify my thinking with this struct and the reason why I didn't even export this struct. I was thinking of just using this struct in the following three responses that are used in channel-info-card, and there the `detailedChannelInfo` seemed required to me, even if all its fields are not set, i.e. `undefined`.

Will probably just remove this struct altogether and define these fields in those responses directly, since you left another comment that I shouldn't structure the API responses in this way anyway; and also it has a vey confusing naming.

server/lib/chat/chat-api.ts
```diff
@@ -361,11 +364,16 @@ export class AdminChatApi {
...
 
     // TODO(2Pac): Move this to the chat-service
-    return await searchChannelsAsAdmin({
+    const channels = await searchChannelsAsAdmin({
       limit,
       pageNumber: page,
       searchStr: searchQuery,
     })
+
+    return await this.chatService.getChannelInfos(
```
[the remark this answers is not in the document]
2Pacalypse-: Yeah, I was gonna refactor this API so it's not admin-specific in the next PR. It's a bit of a mess now, but that seemed preferable than doing the refactoring here, as this PR is already way too long.

common/chat.ts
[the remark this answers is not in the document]
tec27: Yeah, defining it in the response directly is fine (but you may need to adjust DetailedChannelInfo to also have an ID, which means the 3rd struct should probably have an ID as well)

server/lib/chat/chat-api.ts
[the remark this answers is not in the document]
tec27: File an issue for it and that's fine

common/chat.ts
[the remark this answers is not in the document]
2Pacalypse-: Hmm, is that still required if I structure the response where we return multiple channels something like this?
```ts
export interface SearchChannelsResponse {
  channels: {
    channelInfo: BasicChannelInfo
    detailedChannelInfo: DetailedChannelInfo
  }[]
}
```
This way the data would always be grouped and can be iterated over only once. Not sure if this is preferable to having 3 separate structures, each having its own ID.

client/admin/channel-view.tsx
```diff
@@ -69,6 +70,15 @@ function getChannelMessages(
       type: '@users/loadUsers',
       payload: result.mentions.concat(result.users),
     })
+    dispatch({
```
tec27, line 73: I'm 99% sure we've already been over this before, you shouldn't be using the main reducer to show messages to admins. By doing this you are polluting their normal app state. It is okay to do with users because these are data that does not differ depending on the user requesting it, and the data is small (so it's okay to keep around for a long time/forever). With chat, that's not the case, and adding this stuff to the reducer state means it will never be removed, it will just permanently leak memory any time an admin views a channel. I guess you should replace this with a note so that we stop trying to build this in every subsequent PR.

In general, NEVER use redux to store admin state. It is basically never needed to be shared between UIs, and is not necessary to cache, we don't need to use redux for it.

tec27: I actually don't even understand why you did this? The component below wasn't even changed around any of this, so it's still doing the right thing (saving things to its local state)

client/chat/channel-info-selector.ts
```diff
@@ -0,0 +1,54 @@
...
+
+/**
+ * Combined (and flattened) channel info that simplifies the usage in React components.
+ */
+export type ClientChannelInfo = BasicChannelInfo &
+  Partial<DetailedChannelInfo> &
+  Partial<JoinedChannelInfo>
+
+/**
+ * Returns a function for use with `useAppSelector` that finds all the available information about
```
tec27, line 18: I think this is a bad pattern, and doesn't seem particularly easier to use. Please just use the parts that you actually need. By constructing it this way you will force the component to always re-render.

Even if you memoized it, you would still force rerenders of a whole component even if they only depend on one small piece of this state. Just select the pieces separately.

client/chat/channel-info-selector.ts
```diff
@@ -0,0 +1,54 @@
...
+  DetailedChannelInfo,
+  JoinedChannelInfo,
+  SbChannelId,
+} from '../../common/chat'
+import { RootState } from '../root-reducer'
+
+/**
+ * Combined (and flattened) channel info that simplifies the usage in React components.
+ */
+export type ClientChannelInfo = BasicChannelInfo &
```
tec27, line 13: Types like this are a code smell

client/chat/socket-handlers.ts
```diff
@@ -51,7 +51,7 @@ const eventToChatAction: EventToChatActionMap = {
 
   kick: (channelId, event) => (dispatch, getState) => {
     const { auth, chat } = getState()
-    const channelInfo = chat.idToInfo.get(channelId)
+    const channelInfo = chat.idToBasicInfo.get(channelId)
     if (!channelInfo) {
```
tec27, line 55: This hasn't changed as of this PR but this seems extremely weird? Also shouldn't the kick and ban messages contain the information about the channel it applies to?

client/navigation/action-creators.ts
```diff
@@ -8,15 +8,15 @@ export function goToIndex(transitionFn = push): ThunkAction {
...
+      chat: { idToBasicInfo, joinedChannels },
       users: { byId },
       whispers: { sessions },
     } = getState()
     if (lobby.inLobby && IS_ELECTRON) {
       transitionFn(urlPath`/lobbies/${lobby.info.name}`)
     } else if (joinedChannels.size) {
       const [first] = joinedChannels.values()
-      transitionFn(urlPath`/chat/${first}/${idToInfo.get(first)?.name ?? ''}`)
+      transitionFn(urlPath`/chat/${first}/${idToBasicInfo.get(first)?.name ?? ''}`)
```
tec27, line 19: Not a change in this PR, but generally I would replace undefined values with *something*, otherwise you risk problems in adding further parts to the URL in the future. For leagues we use `_` I think. You could pick something else here if you don't want to clash with actual channels/urls (I don't know that we allow 1 character ones anyway?)

common/chat.ts
```diff
@@ -149,18 +136,42 @@ export interface ChannelInfo {
...
+  /** The channel ID. */
+  id: SbChannelId
+  /** The description of the chat channel, if set. */
+  description?: string
+  /** The path to the banner image of the chat channel, if set. */
+  bannerPath?: string
+  /** The path to the badge image of the chat channel, if set. */
+  badgePath?: string
   /**
    * Number of users in the channel. Only available for non-private channels, and for private
```
tec27, line 155: The second part of this comment can be deleted now

common/chat.ts
```diff
@@ -149,18 +136,42 @@ export interface ChannelInfo {
...
+  /** The channel ID. */
+  id: SbChannelId
   /**
-   * Extra properties that are only available when the user has actually joined the channel.
+   * The ID of the user that is considered a channel owner. Usually the person who joined the chat
+   * channel the earliest. Is `undefined` for official channels.
    */
-  joinedChannelData?: JoinedChannelData
+  ownerId?: SbUserId
+  /** A short message used to display the channel's current topic, if there's any. */
```
tec27, line 173: An optional short message to describe the current topic of the channel.

server/lib/chat/chat-service.ts
```diff
@@ -490,43 +533,44 @@ export default class ChatService {
...
-    }
-
+    const channelSummary = toChannelSummary(channelInfo)
     return {
-      id: channelInfo.id,
-      name: channelInfo.name,
-      private: channelInfo.private,
-      official: channelInfo.official,
-      userCount,
+      channelInfo: channelSummary.channelInfo,
```
tec27, line 549: Kinda feels like if we have the `joinedChannelInfo` here and they've joined, we may as well send that here too? Not sure why we just discard it

server/lib/chat/chat-service.ts
```diff
@@ -490,43 +533,44 @@ export default class ChatService {
...
-    })
+    const userJoinedChannelsSet = new global.Set(userChannelEntries.map(e => e.channelId))
+
+    return {
+      channelInfos: channelInfos.map(channel => toChannelSummary(channel).channelInfo),
+      detailedChannelInfos: channelInfos
+        .filter(channel => {
+          return !channel.private || userJoinedChannelsSet.has(channel.id)
+        })
+        .map(channel => toChannelSummary(channel).detailedChannelInfo),
```
tec27, line 572: Doing this conversion every time you need it feels incredibly wasteful. Just write the code to pull the fields you need? Write a for loop to build these arrays iteratively? I dunno, something other than this

Same thing with `joinedChannelInfo` above also applies

server/lib/chat/chat-service.ts
```diff
@@ -589,20 +633,36 @@ export default class ChatService {
...
+              channelMentions.map(c => c.id),
+            ),
+          )
 
     return {
       messages,
       users,
-      mentions,
+      mentions: userMentions,
+      channelMentions: channelMentions.map(c => toChannelSummary(c).channelInfo),
```
tec27, line 664: Given how much this seems to be how you use this over and over again, I really wonder if it wouldn't be better to write 3 separate methods and call them as you need them

server/lib/chat/chat-service.ts
```diff
@@ -765,18 +822,30 @@ export default class ChatService {
...
-    }))
-    userSockets.subscribe(getChannelUserPath(channelInfo.id, userSockets.userId))
+  private subscribeUserToChannel(userSockets: UserSocketsGroup, channelId: SbChannelId) {
+    userSockets.subscribe<ChatInitEvent>(getChannelPath(channelId), async () => {
+      const [channelInfo, userChannelEntry] = await Promise.all([
+        getChannelInfo(channelId),
+        getUserChannelEntryForUser(userSockets.userId, channelId),
+      ])
+      if (!channelInfo) {
+        throw new ChatServiceError(ChatServiceErrorCode.ChannelNotFound, 'Channel not found')
```
tec27, line 832: If you're ever throwing an Error you should ask yourself: What am I expecting to catch this? There's nothing to catch this, you should not be throwing in here.

If you'd like to log something, sure, go ahead. But please do not throw.

tec27: Relatedly, you probably want to wrap this in a try/catch and log the error should one occur (returning undefined in that case is fine, it will not send anything to the client then)

tec27 (changes requested):

client/admin/channel-view.tsx
2Pacalypse-: > I actually don't even understand why you did this?

I honestly have no idea. I have no memory of writing this code and it doesn't even look like something that would be produced by resolving the merge conflicts or something. Weird.

client/chat/socket-handlers.ts
2Pacalypse-: Yeah, I guess it would be weird to receive a message from the server for a channel that the server thinks you're in, but the client doesn't have any information on. Still, this check seemed safer than simply forcing the Typescript to assume the info exists on the client.

I'm not sure I understand your second question; I think they do?

client/navigation/action-creators.ts
2Pacalypse-: > I don't know that we allow 1 character ones anyway?

We do, and `_` is a valid channel name :p

I still think `_` is okay to use here, since the name is only decorative in the URL and will be corrected to its real name once the channel page opens.

server/lib/chat/chat-service.ts
2Pacalypse-: Yeah, I was thinking about that too. But then I felt like that would just be extra information for this API. Maybe if we renamed this API to `getChannelDetails` it would make more sense?

Kinda similarly how we don't send `detailedChannelInfo` and `joinedChannelInfo` in channel mentions, even though we have that information there as well.

client/chat/socket-handlers.ts
tec27: > Still, this check seemed safer than simply forcing the Typescript to assume the info exists on the client.

I don't understand why you would try to be "safe" here? I wasn't suggesting sprinkling `!` on anything, that's obviously unsafe and will cause a crash in this case instead of useful behavior. The client was kicked from a channel, whether they know about it or not. You should tell them they were kicked from the channel. If you don't know what the channels name is, that's fine (I mean, it's not ideal, but it's certainly not a world-ending case), you just display a message that doesn't need the channel name. (Ideally you would display content that could load that info if it didn't have it, but that's going to be hard with a snackbar)

> I'm not sure I understand your second question; I think they do?

I mean if they contain that info then why do you need the reducer state at all? "Oh we don't have the channel name in the reducer so there's nothing I can do" is incredibly bad given the event itself has everything you need?

server/lib/chat/chat-service.ts
tec27: I see no significant difference between `getChannelInfo` and `getChannelDetails` as a name, these are the same thing. I just don't see why we're grabbing data from the DB that can be shown to the client, then throwing it away and making them request this data from somewhere else later (which will then have to re-make the exact same query to get it)

This is not the same as not sending `detailedChannelInfo` and `joinedChannelInfo` in channel mentions, those are being sent to people who we do not know have permission to see those things (and checking that for every message is prohibitively expensive).

client/chat/connected-channel-name.tsx
```diff
@@ -18,22 +27,57 @@ export interface ConnectedChannelNameProps {
...
 
-  const channelName = channelInfo?.name ?? (
-    <LoadingName aria-label={'Channel loading…'}>
-      &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp;
-    </LoadingName>
-  )
+  const [channelInfoCardOpen, openChannelInfoCard, closeChannelInfoCard] = usePopoverController()
+  const [anchor, anchorX, anchorY] = useAnchorPosition('right', 'top')
+
+  const onClick = useStableCallback((e: React.MouseEvent) => {
```
tec27, line 47: I don't think either of these functions are necessary, you can just pass `openChannelInfoCard` and `closeChannelInfoCard` directly?

server/lib/chat/chat-service.ts
```diff
@@ -765,18 +840,37 @@ export default class ChatService {
...
-    userSockets.subscribe(getChannelUserPath(channelInfo.id, userSockets.userId))
+  private subscribeUserToChannel(userSockets: UserSocketsGroup, channelId: SbChannelId) {
+    userSockets.subscribe<ChatInitEvent>(getChannelPath(channelId), async () => {
+      try {
+        const [channelInfo, userChannelEntry] = await Promise.all([
+          getChannelInfo(channelId),
+          getUserChannelEntryForUser(userSockets.userId, channelId),
+        ])
+        if (!channelInfo) {
+          throw new ChatServiceError(ChatServiceErrorCode.ChannelNotFound, 'Channel not found')
```
tec27, line 851: I'm not sure why we're still throwing `ChatServiceErrors` here like they're going to be meaningful in any way, the only thing handling this is the `catch` below, and it does not care about `ChatServiceError` in any way

server/lib/whispers/whisper-service.ts
```diff
@@ -43,6 +48,7 @@ export default class WhisperService {
   private sessionUsers = IMap<SbUserId, ISet<SbUserId>>()
 
   constructor(
+    private chatService: ChatService,
```
tec27, line 51: This dependency doesn't seem to be used? And it would be incredibly weird to have this dependency anyway

server/lib/whispers/whisper-service.ts
```diff
@@ -10,6 +13,8 @@ import {
   WhisperServiceErrorCode,
   WhisperSessionInitEvent,
 } from '../../../common/whispers'
+import { getChannelInfos } from '../chat/chat-models'
+import ChatService, { toBasicChannelInfo } from '../chat/chat-service'
```
tec27, line 17: imo you should move `toBasicChannelInfo` into a different file, I find it weird that this service would ever need to import the chat service

tec27 (approved):

PR #911, opened 2023-03-27
Added translation hooks for Weblate

ExcalZ:


client/matchmaking/find-1v1.tsx
```diff
@@ -88,11 +94,15 @@ const Form1v1 = React.forwardRef<FindMatchFormRef, Form1v1Props>(
...
+          <SectionTitle>
+            {t('findMatch.mapPool', 'Map Pool')}
+          </SectionTitle>
+          {mapPoolOutdated ? <OutdatedIndicator>
+            {t('common.updated', 'Updated')}
+            </OutdatedIndicator> : null}
         </MapSelectionsHeader>
         <DescriptionText>
-          Veto up to 3 maps. Vetoed maps will never be selected for play.
+          {t('findMatch.vetoDescription', 'Veto up to 3 maps. Vetoed maps will never be selected for play.')}
```
tec27: This should have some 1v1-specificity in its name, as the text differs for other matchmaking types (1v1 never allows a vetoed map to be played, 2v2 allows a vetoed map in some cases)

PR #925, opened 2023-04-02
Create a channel list page.

2Pacalypse-:
Currently there's no way to access this page. But the plan is to have the join-channel button navigate to this page as soon as the create-channel button is added to it.

This channel list is meant to be used by both the admins and the public users. Admins will have additional actions that they can use on a channel, e.g. open admin view of a channel.


2Pacalypse-:
Still need to do:
- add create-channel button to the channel list page
- add tests for search channels method in chat-service

client/chat/admin/channel-view.tsx
tec27, line 1: This didn't get marked as a move so make sure you fix this based on the previous PR's changes

client/chat/admin/index.tsx
```diff
@@ -0,0 +1,15 @@
+import React from 'react'
+import { Switch } from 'wouter'
+import { ChannelRoute } from '../route'
+import { AdminChannelView } from './channel-view'
+
+export function ChatAdmin() {
+  return (
+    <Switch>
+      <ChannelRoute
+        path='/chat/admin/channel-view/:channelId/:channelName'
```
tec27, line 10: what is up with this path? This doesn't match any kind of pattern we have in the app

client/chat/route.tsx
```diff
@@ -1,30 +1,56 @@
...
+  const isAdmin = useAppSelector(s => hasAnyPermission(s.auth, 'moderateChatChannels'))
 
   return (
-    <ConnectedChatChannel
-      channelId={makeSbChannelId(channelIdNum)}
-      channelName={params!.channelName}
-    />
+    <Suspense fallback={<LoadingDotsArea />}>
+      <Switch>
+        {isAdmin ? <Route path='/chat/admin/:rest*' component={LoadableChatAdmin} /> : <></>}
```
tec27, line 48: I would recommend always having a route for this and just rendering something that says they don't have permission to access it or something. Not having a route all the time in Leagues has caused problems and needs to be fixed

client/chat/route.tsx
```diff
@@ -1,30 +1,56 @@
...
+        return (
+          <Component channelId={makeSbChannelId(channelIdNum)} channelName={params.channelName} />
+        )
+      }}
+    </Route>
+  )
+}
+
+export function ChannelRouteComponent(props: { params: any }) {
+  // TODO(2Pac): Add a separate permission for managing chat channels
```
tec27, line 42: Why? What purpose would that serve?

client/chat/channel.tsx
```diff
@@ -256,7 +256,7 @@ function ChannelInfoPage({
     } else {
       errorText = (
         <ErrorText>
-          Error getting channel info {channelName}: {error.message}
+          Error getting channel info for {channelName}: {error.message}
```
tec27, line 259: should probably be prefixed with `#` as well

client/chat/chat-reducer.ts
tec27, line 1: This seems like something that should have been in the other PR, not here?

client/chat/create-channel.tsx
```diff
@@ -0,0 +1,105 @@
...
+  const autoFocusRef = useAutoFocusRef<HTMLInputElement>()
+
+  const onFormSubmit = useStableCallback((model: JoinChannelModel) => {
+    const channelName = model.channel
+
+    dispatch(
+      createChannel(channelName, {
+        onSuccess: channel => navigateToChannel(channel.channelInfo.id, channel.channelInfo.name),
+        onError: err => {
+          let message = `An error occurred while joining ${channelName}`
```
tec27, line 58: creating? also should probably be prefixed with `#`

client/chat/create-channel.tsx
```diff
@@ -0,0 +1,105 @@
...
+          if (isFetchError(err) && err.code) {
+            if (err.code === ChatServiceErrorCode.UserBanned) {
+              message = `You are banned from ${channelName}`
+            } else {
+              logger.error(`Unhandled code when joining ${channelName}: ${err.code}`)
+            }
+          } else {
+            logger.error(`Error when joining ${channelName}: ${err.stack ?? err}`)
+          }
+          setErrorMessage(message)
```
tec27, line 69: I would suggest never doing this. The nice part about handling the error here is that we can have access to the actual error when rendering. If you turn everything into a single string and use that as state you have destroyed most of this advantage. Just save the error in state and do the logic to convert it to something visual in render (where you can like, do something nicer than just display a single line of text)

client/material/left-nav/clickable-subheader.tsx
```diff
@@ -0,0 +1,71 @@
...
+`
+
+interface SubheaderProps {
+  icon?: React.ReactNode
+  children: React.ReactNode
+  className?: string
+  onClick?: (event: React.MouseEvent) => void
+}
+
+export const ClickableSubheader = React.forwardRef(
```
tec27, line 55: Use a link for this type of stuff please. If you're just gonna navigate to a url, use a link

server/lib/chat/chat-api.ts
```diff
@@ -337,21 +342,18 @@ export class ChatApi {
...
 
   @httpGet('/')
-  @httpBefore(checkAllPermissions('moderateChatChannels'))
+  @httpBefore(
+    featureEnabled(MULTI_CHANNEL),
+    throttleMiddleware(channelRetrievalThrottle, ctx => String(ctx.session!.userId)),
+  )
   async searchChannels(ctx: RouterContext): Promise<SearchChannelsResponse> {
     const {
       query: { q: searchQuery, limit, page },
```
tec27, line 359: `limit` should almost certainly have a maximum

tec27: `page` just seems weird as hell too, why are we passing in a value that we have to multiply by another value to get to anything useful (and the client likely has to divide to send to us)? Just pass the offset

server/lib/chat/chat-api.ts
```diff
@@ -363,18 +365,19 @@ export class AdminChatApi {
...
   }
+}
+
+@httpApi('/admin/chat')
+@httpBeforeAll(ensureLoggedIn, convertChatServiceErrors)
+export class AdminChatApi {
+  constructor(private chatService: ChatService) {}
 
   @httpGet('/:channelId/messages')
   @httpBefore(checkAllPermissions('moderateChatChannels'))
```
tec27, line 383: Think you should just put this on the class so we don't accidentally forget it one time, I can't see a use for any other permission here

server/lib/chat/chat-service.ts
```diff
@@ -573,6 +575,39 @@ export default class ChatService {
...
+    limit,
+    pageNumber,
+    searchStr,
+  }: {
+    userId: SbUserId
+    limit: number
+    pageNumber: number
+    searchStr?: string
+  }): Promise<SearchChannelsResponse> {
+    const channels = await searchChannels({
```
tec27, line 589: It doesn't really seem like there's another use for this model method, so why not just do this with a join and return the channels + the UserChannelEntry's in one go?

tec27 (changes requested):

client/chat/admin/index.tsx
2Pacalypse-: Hmm, is `/chat/admin/:channelId/:channelName/view` better?
Assuming we'll also have `/chat/admin/:channelId/:channelName/manage`.

client/chat/route.tsx
2Pacalypse-: Hmm, my thinking was that we might have channel moderators that shouldn't have access to change channel's banners and descriptions. But maybe that's all a part of channel moderation?

client/chat/chat-reducer.ts
2Pacalypse-: Perhaps, but at this point I just want to get that other PR merged :d

And it's not like it's broken in the other PR, just a stylistic choice I opted for after adding the `@chat/searchChannels` action, which has the same payload as `@chat/getBatchChannelInfo`.

client/chat/create-channel.tsx
2Pacalypse-: Yeah, that's what I usually do too. The reason I moved it inside the error handler here is because I could read the `channelName` here that the form was submitted with, without saving it somewhere in state. I tried using the `getInputValue` first to read it directly from the form model, but then the error message gets updated as well if the search input changes.

Maybe there's something in the form hook that can be used to read the values that the form was last submitted with?

server/lib/chat/chat-api.ts
2Pacalypse-: > `page` just seems weird as hell too, why are we passing in a value that we have to multiply by another value to get to anything useful (and the client likely has to divide to send to us)? Just pass the offset

Hmm, the client just increases the `page` by 1 when it reaches the end of a list and wants to load more channels. But I guess increasing the current offset by the `limit` works too.

client/chat/route.tsx
tec27: I think you are confusing "channel moderation" as in "people that have mod in a channel" and "channel moderation" as in "people who can access and moderate every channel on the server". The latter seems like such a wide permission that I don't see why we would trust people to do it but couldn't also trust that same person to manage a banner setting or something

client/chat/create-channel.tsx
tec27: Just set it in state? I don't see why you're trying to take some worse route just to avoid 2 lines of code

tec27: Alternatively just don't include the channel name in the message? I don't really see how that serves much purpose in a real-world scenario

server/lib/chat/chat-api.ts
tec27: You don't have to do any math, the list already contains the number of elements you have seen and is thus the starting point of the next request. The only time you need to deal with the limit is to know if you received less than you asked for (so you're done), but that should probably be a `boolean` returned in the response anyway

client/chat/route.tsx
```diff
@@ -1,30 +1,59 @@
...
 import { replace } from '../navigation/routing'
+import { LoadingDotsArea } from '../progress/dots'
+import { useAppSelector } from '../redux-hooks'
 import { ConnectedChatChannel } from './channel'
+import { ChannelList } from './channel-list'
+import { CreateChannel } from './create-channel'
 
-export function ChannelRouteComponent(props: { params: any }) {
-  const [matches, params] = useRoute('/chat/:channelId/:channelName')
+const getLoadableChatAdminComponent = (isAdmin: boolean) =>
```
tec27, line 12: Can you please just write this inline? Routes can just have children instead of passing a `component` and this makes things needlessly complex

client/chat/route.tsx
```diff
@@ -1,30 +1,59 @@
...
+  isAdmin
+    ? React.lazy(async () => ({
+        default: (await import('./admin')).ChatAdmin,
+      }))
+    : // TODO(2Pac): Create a better-designed, general component for this
```
tec27, line 17: Can we stop sprinkling TODOs everywhere? It is just about as many characters to write this TODO as it is to write a component that translates this value to some common string, so please just do that if that's what you want. I can't imagine what else this is about, as this is just a span with text in it.

client/navigation/url-hotkey.ts
```diff
@@ -0,0 +1,50 @@
...
+  transitionFn?: (url: string) => void
+}
+
+/**
+ * A hook which allows any component to register a hotkey (or multiple hotkeys) which upon pressing
+ * will navigate the application to the given URL.
+ *
+ * If the currently active URL is the same as the given one, pressing the hotkey will do nothing.
+ */
+export function useUrlHotkey({ url, hotkey, disabled, transitionFn = push }: UrlHotkeyProps) {
```
tec27, line 26: This hook seems unnecessary, you could just click the link in the same way that we click any other button, and then it would also still activate the ripple and such?

If you really have to have it (and I would very much that we don't), at least rename it `useNavigationHotkey`

client/material/left-nav/clickable-subheader.tsx
```diff
@@ -0,0 +1,74 @@
...
+  icon?: React.ReactNode
+  children: React.ReactNode
+  className?: string
+}
+
+export const ClickableSubheader = React.forwardRef(
+  ({ icon, children, className }: SubheaderProps, ref: React.ForwardedRef<HTMLDivElement>) => {
+    const [isActive] = useRoute('/chat/list')
+
+    return (
```
tec27, line 65: This still needs a ripple and such somewhere. Ideally the icon would also behave like an icon button (or just make it an icon button), just the click can be handled by this component

client/material/left-nav/clickable-subheader.tsx
```diff
@@ -0,0 +1,74 @@
...
+import { overline, singleLine } from '../../styles/typography'
+
+const Container = styled.div<{ $isActive: boolean }>`
+  width: 100%;
+  height: 36px;
+  background-color: ${props => (props.$isActive ? 'rgba(255, 255, 255, 0.08)' : 'transparent')};
+  color: ${props => (props.$isActive ? colorTextPrimary : colorTextSecondary)};
+  cursor: pointer;
+
+  &:hover {
```
tec27, line 14: As stated below, please just handle this type of stuff through a ripple instead of adding a bespoke hover state

client/material/left-nav/clickable-subheader.tsx
```diff
@@ -0,0 +1,74 @@
...
+import styled from 'styled-components'
+import { Link, useRoute } from 'wouter'
+import { colorTextPrimary, colorTextSecondary } from '../../styles/colors'
+import { overline, singleLine } from '../../styles/typography'
+
+const Container = styled.div<{ $isActive: boolean }>`
+  width: 100%;
+  height: 36px;
+  background-color: ${props => (props.$isActive ? 'rgba(255, 255, 255, 0.08)' : 'transparent')};
+  color: ${props => (props.$isActive ? colorTextPrimary : colorTextSecondary)};
```
tec27, line 11: I don't think any of this active coloring is really desirable, that should be reserved for channels/whispers/etc., not the headers above them

client/chat/channel-list.tsx
```diff
@@ -0,0 +1,168 @@
...
+    abortControllerRef.current?.abort()
+    abortControllerRef.current = new AbortController()
+
+    dispatch(
+      searchChannels(searchQuery, SEARCH_CHANNELS_LIMIT, channels?.length ?? 0, {
+        signal: abortControllerRef.current.signal,
+        onSuccess: data => {
+          setIsLoadingMoreChannels(false)
+          setChannels((channels ?? []).concat(data.channelInfos))
+          setHasMoreChannels(data.channelInfos.length >= SEARCH_CHANNELS_LIMIT)
```
tec27, line 106: Like I said before, it would be better if any paginated APIs just returned this information in their response instead of doing stuff like this. The service can implement this in a similar way, it just makes it so we don't have to keep and re-use this value on the client to do this check (and the server may have a better idea of how many results there actually are, so it can avoid the client requesting empty lists in some cases)

I understand that some of these APIs existed before, but the search one is new, so I don't know why we're repeating bad design choices in it (and in some cases coming up with brand new bad design choices)

server/lib/chat/chat-service.ts
```diff
@@ -565,6 +567,49 @@ export default class ChatService {
...
+    const { channels, joinedChannels } = await searchChannelsAsUser({
+      userId,
+      limit,
+      offset,
+      searchStr,
+    })
+
+    const userJoinedChannelsSet = new global.Set(joinedChannels)
+
+    const [detailedChannelInfos, joinedChannelInfos] = channels.reduce<
```
tec27, line 590: Can you just write this as a for loop please? This is an extremely complex way of doing "iterate through a list and occasionally push values into 2 lists"

server/lib/chat/chat-models.ts
```diff
@@ -721,12 +732,15 @@ export async function searchChannelsAsAdmin(
...
+      OFFSET ${offset}
     `)
 
-    const result = await client.query<DbChannel>(query)
+    const result = await client.query<DbChannel & { joined: boolean }>(query)
 
-    return result.rows.map(row => convertChannelFromDb(row))
+    return {
+      channels: result.rows.map(row => convertChannelFromDb(row)),
+      joinedChannels: result.rows.filter(row => row.joined).map(row => row.id),
```
tec27, line 742: We go to a bunch of extra effort to basically join this up, such that the list of joined channels is only what is in the returned list, but this doesn't actually feel useful in any way? I think it'd be better to remove this functionality from this method entirely and just send 2 queries to the DB: one to do the search and one to get the list of joined channels. If you want a function that returns just the joined channel IDs without any info, that's fine to create (and very simple)

tec27 (changes requested):

client/chat/route.tsx
tec27: I'll also just note that `React.lazy` creates a *component*, so what you have done here is made a thing that creates a brand new component type every time these routes are rendered, so please never write code like this

https://react.dev/reference/react/lazy#my-lazy-components-state-gets-reset-unexpectedly

client/material/left-nav/clickable-subheader.tsx
2Pacalypse-: Hmm, I'm not sure how to do this.

Just to recap, we want channel header to be a button (because that's the only way I can apply ripple effect to it afaik), with an icon button inside it (which has its own built-in ripple effect). Also, the whole thing needs to be inside a link as well?

client/chat/channel-info-card.tsx
```diff
@@ -46,12 +54,26 @@ const ChannelCardBadge = styled.div`
...
-const ChannelName = styled.div`
+const ChannelNameContainer = styled.div`
+  position: relative;
+`
+
+const ChannelName = styled.div<{ $hasOverflowMenu: boolean }>`
   ${headline6};
+  ${singleLine};
   margin-top: 4px;
+  margin-right: ${props => (props.$hasOverflowMenu ? '48px' : '0px')};
```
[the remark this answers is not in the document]
2Pacalypse-: Yeah, the only reason why I added this is because the overflow menu button is absolutely positioned. Maybe I can just position it relatively to align it better without taking it out completely out of the flex layout. Or group the channel name + user count together in a flex-column which is in a flex-row together with the overflow button.

client/material/left-nav/clickable-subheader.tsx
tec27: No, it is a link that is styled like a button. We want it to be a link so that search engine's can follow it (and because all the button did was navigate). We have plenty of other cases of "things that are not `<button>`s that use ripple, etc." for you to look at

client/chat/channel-info-card.tsx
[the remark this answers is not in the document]
tec27: That's what I'm saying, you shouldn't absolutely position the overflow menu button. You are reimplementing a layout algorithm in a bad way by doing what you're doing here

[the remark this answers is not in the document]
tec27: If the channel name is where League names are, I don't quite see why this padding is necessary though (the overflow menu should probably be at the top right of the card, and thus these elements are not adjacent ever). If that is the case, then absolute positioning is fine (but you don't need what you're doing here at all). I assume you have instead put it by the title, where absolute positioning makes zero sense. But I think I would prefer you just move it to the top right. Either way, this kind of prop on this style is unnecessary.

server/lib/chat/chat-models.ts
```diff
@@ -692,17 +693,14 @@ export async function findChannelsByName(
...
- * Returns a full list of chat channels, optionally filtered by a `searchStr`. Only admins should be
- * able to call this function.
+ * Returns a list of chat channels, optionally filtered by a `searchStr`.
  */
-export async function searchChannelsAsAdmin(
+export async function searchChannels(
   {
-    limit,
-    pageNumber,
+    offset,
```
tec27, line 700: imo you should still have a limit parameter, just specifying the offset and not the "page number"

Really this whole method of pagination is bad (it gets messed up if a channel is ever added/removed), but I'm not gonna make you rewrite *all* of this right now

server/lib/chat/chat-service.ts
```diff
@@ -565,6 +563,41 @@ export default class ChatService {
     }
   }
 
+  async searchChannels({
+    userId,
+    offset,
+    searchStr,
+  }: {
+    userId: SbUserId
+    offset: number
```
tec27, line 572: as in the model, I think it still makes sense to have a limit parameter

tec27 (changes requested):

PR #932, opened 2023-04-22
Send the ordered list of chat channels and whisper sessions to client.

2Pacalypse-:
Also moves the loading code to ts/immer.

And adds a test to the chat-service for handling new users.


client/loading/loading-reducer.ts
```diff
@@ -0,0 +1,49 @@
...
+const DEFAULT_LOADING_STATE: ReadonlyDeep<LoadingState> = {
+  audio: true,
+  chat: true,
+  clientSubscriptions: true,
+  userSubscriptions: true,
+  whispers: true,
+}
+
+export default immerKeyedReducer(DEFAULT_LOADING_STATE, {
+  [AUDIO_MANAGER_INITIALIZED as any](state: any) {
```
tec27, line 22: why did you make `state` `any` here? The type is already inferred from the default you passed in here, and you're destroying type safety by doing that

client/loading/loading-reducer.ts
```diff
@@ -0,0 +1,49 @@
...
+
+  ['@loading/subscribedUser'](state) {
+    state.userSubscriptions = false
+  },
+
+  ['@loading/whispersReady'](state) {
+    state.whispers = false
+  },
+
+  [NETWORK_SITE_DISCONNECTED as any](state: any) {
```
tec27, line 42: Same as above

server/lib/chat/chat-service.test.ts
```diff
@@ -396,6 +396,79 @@ describe('chat/chat-service', () => {
     clearTestLogs(nydus)
   })
 
+  describe('handleNewUser', () => {
+    test('subscribes user to their joined channels', async () => {
+      const user3: SbUser = { id: 3 as SbUserId, name: 'USER_NAME_3' }
+      const user3ShieldBatteryChannelEntry = { ...user1ShieldBatteryChannelEntry, userId: user3.id }
+      const user3TestChannelEntry = { ...user1TestChannelEntry, userId: user3.id }
```
tec27, line 403: You should adjust the times on these channel entries (in the constant you're pulling from here) imo, this isn't really realistic atm (they have the same join time, and that's what we'd be sorting on). It's possible in the future we'd need to sort this data in the server code or something, in which case this test could break even if the server code was actually correct.

server/lib/chat/chat-service.ts
```diff
@@ -909,21 +910,24 @@ export default class ChatService {
...
       this.publisher.publish(getChannelPath(userChannel.channelId), {
         action: 'userActive2',
         userId: userSockets.userId,
       })
       this.subscribeUserToChannel(userSockets, userChannel.channelId)
     }
-    userSockets.subscribe(`${userSockets.getPath()}/chat`, () => ({ type: 'chatReady' }))
+    userSockets.subscribe<ChatReadyEvent>(`${userSockets.getPath()}/chat`, () => ({
+      type: 'chatReady',
+      channelIds: channelIdsSet.toArray(),
```
tec27, line 929: This will always use the channel IDs that were set when the user connected, so it is broken if they join or leave a channel and then connect a new socket. You need to re-retrieve it from the state or DB.

Whenever you write a `subscribe` data getter, you should almost never be using values from the closure. That is very rarely a correct thing to do (if it were correct, you could simply pass the object and not use a getter function)

server/lib/whispers/whisper-service.ts
```diff
@@ -290,7 +289,10 @@ export default class WhisperService {
       })
     }
 
-    userSockets.subscribe(`${userSockets.getPath()}/whispers`, () => ({ type: 'whispersReady' }))
+    userSockets.subscribe<WhispersReadyEvent>(`${userSockets.getPath()}/whispers`, () => ({
+      type: 'whispersReady',
+      targetIds: targetIdsSet.toArray(),
```
tec27, line 294: Same problem as in the chat service, you need to re-retrieve this data

tec27 (changes requested):

server/lib/chat/chat-service.ts
```diff
@@ -924,10 +924,20 @@ export default class ChatService {
...
+        // TODO(2Pac): Can read this from state once that's moved to non-immutable.js `Set` (since
+        // the current one is unordered).
+        const joinedChannels = await getChannelsForUser(userSockets.userId)
+        return {
+          type: 'chatReady',
+          channelIds: joinedChannels.map(c => c.channelId),
+        }
+      } catch (err) {
+        logger.error({ err }, 'Error retrieving the list of ordered joined channels for the user')
+        return undefined
```
tec27, line 938: This seems very dangerous to do, as we hold the client from rendering anything until `chatReady` happens. You should probably send an empty `channelIds` list in that case

tec27 (approved):

PR #933, opened 2023-04-22
Limit the number of joined and owned channels that the user can have.

2Pacalypse-:


client/chat/action-creators.ts
```diff
@@ -55,7 +55,9 @@ export function joinChannelWithErrorHandling(
         let message = `An error occurred while joining ${channelName}`
 
         if (isFetchError(err) && err.code) {
-          if (err.code === ChatServiceErrorCode.UserBanned) {
+          if (err.code === ChatServiceErrorCode.MaximumJoinedChannels) {
+            message = "You can't join anymore channels."
```
tec27, line 59: "any more" is what you'd want here (English sucks I know :) )

I think it might be nice to explain this a bit further though, not totally clear to me *why* I can't join more channels with this. Perhaps:

"You have reached the limit of joined channels. You must leave one before you can join another."

client/chat/create-channel.tsx
```diff
@@ -70,7 +70,9 @@ export function CreateChannel() {
     errorMessage = 'An error occurred while creating the channel.'
 
     if (isFetchError(error) && error.code) {
-      if (error.code === ChatServiceErrorCode.UserBanned) {
+      if (error.code === ChatServiceErrorCode.MaximumOwnedChannels) {
+        errorMessage = "You've reached the limit to how many channels you can create."
```
tec27, line 74: Similar to the above, might be good to provide some remediation action here (e.g. suggest they leave or transfer ownership of a channel)

server/lib/chat/chat-models.ts
```diff
@@ -116,6 +117,41 @@ export async function getUserChannelEntriesForUser(
   }
 }
 
+export async function countUserJoinedChannels(
```
tec27, line 120: It's bad to have these be separate queries, there's nothing stopping this from being changed in the middle of queries running. With stuff like this you either:

1) Need to lock a row that all other queries modifying this data will also lock (this requires transactions)
2) Ensure the limit in the same query (*query*, not transaction) that you're doing the update (e.g. have a where clause that won't apply if they're over the limit)

I think 2 likely makes more sense here, as we don't really have a good single row to lock and requiring transactions kinda sucks.

server/lib/chat/chat-models.ts
```diff
@@ -116,6 +117,41 @@ export async function getUserChannelEntriesForUser(
...
+  withClient?: DbClient,
+): Promise<number> {
+  const { client, done } = await db(withClient)
+  try {
+    const result = await client.query<DbUserChannelEntry>(sql`
+      SELECT *
+      FROM channel_users
+      WHERE user_id = ${userId};
+    `)
+    return result.rowCount
```
tec27, line 131: I've already recommended to delete all this code, but for future reference: uh what? Why are you using `rowCount` to perform a `COUNT(*)`? The database can potentially optimize the `COUNT`, especially since it operates only on indexes, but instead you wrote a query to return a bunch of data and then immediately throw it all away? Please don't ever do this

server/lib/chat/chat-models.ts
```diff
@@ -116,6 +117,41 @@ export async function getUserChannelEntriesForUser(
...
+export async function countUserJoinedChannels(
+  userId: SbUserId,
+  withClient?: DbClient,
+): Promise<number> {
+  const { client, done } = await db(withClient)
+  try {
+    const result = await client.query<DbUserChannelEntry>(sql`
+      SELECT *
+      FROM channel_users
+      WHERE user_id = ${userId};
```
tec27, line 129: With whatever new additions to queries you write, I think that `official` channels should not count towards this limit

tec27 (changes requested):

server/lib/chat/chat-models.ts
```diff
@@ -206,16 +173,21 @@ export async function createChannel(
   userId: SbUserId,
   channelName: string,
   withClient?: DbClient,
-): Promise<FullChannelInfo> {
+): Promise<FullChannelInfo | undefined> {
```
tec27, line 176: You should document why this would be undefined

server/lib/chat/chat-models.ts
```diff
@@ -225,20 +197,22 @@ export async function addUserToChannel(
   userId: SbUserId,
   channelId: SbChannelId,
   withClient?: DbClient,
-): Promise<UserChannelEntry> {
+): Promise<UserChannelEntry | undefined> {
```
tec27, line 200: You should document why this would be undefined

server/lib/chat/chat-service.ts
```diff
@@ -174,6 +170,9 @@ export default class ChatService {
       throw new ChatServiceError(ChatServiceErrorCode.ChannelNotFound, 'Channel not found')
     }
 
+    // NOTE(2Pac): This method can technically return an `undefined`, but that would mean we're
+    // trying to add user to a bunch of non-official channels when they create an account which,
+    // you know... we probably shouldn't do.
     const userChannelEntry = await addUserToChannel(userId, channelId, client)
```
tec27, line 176: If this is the place you are assuming it is defined, this is the place you should add the `!`, not everywhere you try to use the value

server/lib/chat/chat-service.ts
```diff
@@ -257,6 +251,11 @@ export default class ChatService {
 
             try {
               userChannelEntry = await addUserToChannel(userId, channel.id, client)
+              if (!userChannelEntry) {
```
tec27, line 254: This feels fairly unwieldy, adding more booleans for mutually exclusive conditions doesn't feel great. Imo you should have an enum/union type of some sort and combine all 3 exit conditions into one variable.

As evidence for this being a problem, you added these condition variables but did not add them as a loop exit condition, so this loop will always run the maximum number of times in the case that they're at a limit.

tec27 (changes requested):

server/lib/chat/chat-service.ts
```diff
@@ -309,21 +313,21 @@ export default class ChatService {
...
     }
-    if (maximumOwnedChannels) {
+    if (exitCode === JoinChannelExitCode.MaximumOwnedChannels) {
       throw new ChatServiceError(
         ChatServiceErrorCode.MaximumOwnedChannels,
         'Maximum owned channels reached',
       )
     }
-    if (isUserBanned) {
+    if (exitCode === JoinChannelExitCode.UserBanned) {
```
tec27, line 330: You should add an `assertUnreachable` clause after these to make sure these all get checked:
```ts
if (exitCode !== undefined) {
  assertUnreachable(exitCode)
}
```

tec27 (approved):

PR #937, opened 2023-05-03
Throws error in development if game hasn't been built

williamhammond:
I noticed this  while looking at replays while running locally. I ran into it because I hadn't run game/build.bat

I'm not sure if throwing is too aggressive here, I could use a similar dialog as a normal installation issue.  

image

### DIFF
--- a/common/flags.ts
+++ b/common/flags.ts
@@ -9,6 +9,8 @@ const DEV = () => process.env.NODE_ENV !== 'production'
 const ON = () => true
 /* eslint-enable @typescript-eslint/no-unused-vars */
 
+/** Special error handling for local development */
+export const DEV_ERROR = DEV()
 /** Show the "dev mode" indicator on the site (don't move this past dev mode). */
 export const DEV_INDICATOR = DEV()
 /** Allow users to join multiple channels (and leave channels). */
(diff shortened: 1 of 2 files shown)


tec27:
I feel like a better thing to do might just be to use the `DEV` feature flag and render different content in this dialog, e.g. in the dialog body:
```tsx
{DEV ?
  <div>Couldn't find necessary ShieldBattery files, you probably need to run game/build.bat</div> : 
  <>{/* existing content */}</>
}
```

Would generally prefer that approach over throwing since it makes it easier to test that this dialog shows up properly, while still letting us give a nicer hint to developers.

common/flags.ts
```diff
@@ -9,6 +9,8 @@ const DEV = () => process.env.NODE_ENV !== 'production'
 const ON = () => true
 /* eslint-enable @typescript-eslint/no-unused-vars */
 
+/** Special error handling for local development */
+export const DEV_ERROR = DEV()
```
williamhammond, line 13: Wasn't sure if it made sense to export `DEV()`

tec27: This seems fine, yeah

tec27:
Thank you!

PR #938, opened 2023-05-03
Add translation strings to the auth folder.

2Pacalypse-:
i18next-parser currently throws an error while trying to generate the translation keys. Will need to figure out what trips it (the stack trace it shows seems fairly useless).

### DIFF
--- a/client/auth/forgot.jsx
+++ b/client/auth/forgot.jsx
@@ -40,17 +40,52 @@ import {
   SuccessContainer,
 } from './auth-content'
 
+// TODO(2Pac): Use the `useTranslation` hook once this is moved over to a functional component. Note
+// that I'm using the global version of the `t` function here. react-i18next also exposes a HOC that
+// can be used with class components to make the `t` function reactive, but making that work with
+// form validators here would be quite cumbersome, so this seemed easier until it gets replaced with
+// hooks.
+import { Trans } from 'react-i18next'
+import { t } from '../i18n/i18next'
+
 const emailValidator = composeValidators(
-  required('Enter an email address'),
-  minLength(EMAIL_MINLENGTH, `Use at least ${EMAIL_MINLENGTH} characters`),
-  maxLength(EMAIL_MAXLENGTH, `Use at most ${EMAIL_MAXLENGTH} characters`),
-  regex(EMAIL_PATTERN, 'Enter a valid email address'),
+  required(t('auth.emailValidator.required', 'Enter an email address')),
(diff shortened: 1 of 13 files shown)


client/auth/forgot.jsx
```diff
@@ -120,7 +161,11 @@ class ForgotFormHolder extends React.Component {
         </AuthContentContainer>
         {loadingContents}
         <AuthBottomAction>
-          <BottomActionButton label='Back to login' onClick={this.onBackClick} tabIndex={1} />
+          <BottomActionButton
+            label={t('auth.forgot.backToLogin', 'Back to login.')}
```
tec27, line 165: Why does this string suddenly have a period?

client/auth/user-error-display.tsx
```diff
@@ -16,60 +17,96 @@ export interface UserErrorDisplayProps {
...
+          </Trans>
+        </span>
+      )
 
     default:
       return (
         <span>
-          An error occurred: {error.status} {error.statusText}
+          <Trans t={t} i18nKey='auth.userErrorDisplay.defaultError'>
+            An error occurred: {error.status} {error.statusText}
```
tec27, line 86: I would guess this is probably the issue with the extractor and that you need to assign status/statusText to variables first

tec27: The `Trans` examples also use double braces (`{{name}}`) and you've used single throughout this stuff

tec27: You're also supposed to pass these values in on the `Trans` component, so yeah, this just isn't being used correctly here and is almost certainly the issue, just look at the examples: https://react.i18next.com/latest/trans-component

tec27: https://react.i18next.com/latest/trans-component#using-with-react-components

client/i18n/i18next.ts
```diff
@@ -10,6 +10,15 @@ import { makeServerUrl } from '../network/server-url'
...
 const isDev = __WEBPACK_ENV.NODE_ENV !== 'production'
 
+/**
+ * Type to use for interpolations in `Trans` components since React doesn't allow objects as
+ * children.
+ *
+ * Taken as a best solution from this comment:
+ * https://github.com/i18next/react-i18next/issues/1483#issuecomment-1268455602
+ */
+export type TI = any
```
tec27, line 20: Please expand this to an actual name, `TI` is meaningless

Issue #936, opened 2023-05-02
Missing replay folder isn't handled


williamhammond:
I noticed on a fresh install, that the underlying `no such file or directory` error is displayed when trying to look at replays.

I think the expected behavior should be the same as an empty replay directory

Actual:
![image](https://user-images.githubusercontent.com/3045409/235757633-dc63aea0-2576-4c9a-8eea-26bf887c745e.png)

Expected:
![image](https://user-images.githubusercontent.com/3045409/235758138-c94d1544-ecc7-4e09-904a-96e798214e98.png)

tec27:
The screen this was filed against (file-browser-based replay page) no longer exists. The replay library that replaced it (#1358) treats a missing/unreadable replay folder exactly as this issue asked: logged warning, shown as empty, no user-facing ENOENT error.

PR #939, opened 2023-05-09
Adds custom error type for handling FS error codes

williamhammond:
This partially fixes an electron issues where fs error code information isn't getting shared from the main process to the render process https://github.com/electron/electron/issues/24427. Currently the error code information is getting lost when communicating over IPC

### DIFF
--- a/app/app.ts
+++ b/app/app.ts
@@ -12,7 +12,16 @@ import { pipeline } from 'stream/promises'
 import { container } from 'tsyringe'
 import { URL } from 'url'
 import swallowNonBuiltins from '../common/async/swallow-non-builtins'
-import { FsDirent, TypedIpcMain, TypedIpcSender } from '../common/ipc'
+import {
+  FsDirent,
+  FsError,
+  FsErrorCode,
+  FsReadDirResult,
+  FsReadFileResult,
+  FsStatResult,
+  TypedIpcMain,
+  TypedIpcSender,
+} from '../common/ipc'
 import { ReplayShieldBatteryData } from '../common/replays'
 import { checkShieldBatteryFiles } from './check-shieldbattery-files'
 import currentSession from './current-session'
@@ -396,32 +405,76 @@ function setupIpc(localSettings: LocalSettings, scrSettings: ScrSettings) {
   )
 
   ipcMain.handle('fsReadFile', async (_, filePath) => {
-    return fsPromises.readFile(filePath)
+    try {
+      const arrayBuffer = await fsPromises.readFile(filePath)
+      const result: FsReadFileResult = {
+        file: arrayBuffer,
+      }
+      return result
+    } catch (e) {
+      if ((e as any).code === 'ENOENT') {
+        const error: FsError = {
+          error: true,
+          code: FsErrorCode.FileOrFolderMissing,
+        }
+        return error
+      }
+      throw e
+    }
   })
   ipcMain.handle('fsReadDir', async (_, dirPath, options) => {
-    const result = await fsPromises.readdir(dirPath, options)
-    return result.map<FsDirent>(d => ({
-      isFile: d.isFile(),
-      isDirectory: d.isDirectory(),
-      name: d.name,
-    }))
+    try {
+      const result = await fsPromises.readdir(dirPath, options)
+      const files = result.map<FsDirent>(d => ({
+        isFile: d.isFile(),
+        isDirectory: d.isDirectory(),
+        name: d.name,
+      }))
+      const dirents: FsReadDirResult = {
+        entries: files,
+      }
+      return dirents
+    } catch (e) {
+      if ((e as any).code === 'ENOENT') {
+        const error: FsError = {
+          code: FsErrorCode.FileOrFolderMissing,
+          error: true,
+        }
+        return error
+      }
+      throw e
+    }
   })
   ipcMain.handle('fsStat', async (_, filePath) => {
-    const result = await fsPromises.stat(filePath)
-    return {
-      isFile: result.isFile(),
-      isDirectory: result.isDirectory(),
-      size: result.size,
-      blksize: result.blksize,
-      blocks: result.blocks,
-      atimeMs: result.atimeMs,
-      mtimeMs: result.mtimeMs,
-      ctimeMs: result.ctimeMs,
-      birthtimeMs: result.birthtimeMs,
-      atime: result.atime,
-      mtime: result.mtime,
-      ctime: result.ctime,
-      birthtime: result.birthtime,
+    try {
+      const result = await fsPromises.stat(filePath)
+      const stat: FsStatResult = {
+        stats: {
+          isFile: result.isFile(),
+          isDirectory: result.isDirectory(),
+          size: result.size,
+          blksize: result.blksize,
+          blocks: result.blocks,
+          atimeMs: result.atimeMs,
+          mtimeMs: result.mtimeMs,
+          ctimeMs: result.ctimeMs,
+          birthtimeMs: result.birthtimeMs,
+          atime: result.atime,
+          mtime: result.mtime,
+          ctime: result.ctime,
+          birthtime: result.birthtime,
+        },
+      }
+      return stat
+    } catch (e) {
+      if ((e as any).code === 'ENOENT') {
+        const error: FsError = {
+          error: true,
+          code: FsErrorCode.FileOrFolderMissing,
+        }
+        return error
+      }
+      throw e
     }
   })
 
--- a/client/file-browser/get-files.ts
+++ b/client/file-browser/get-files.ts
@@ -1,4 +1,4 @@
-import { TypedIpcRenderer } from '../../common/ipc'
+import { FsErrorCode, TypedIpcRenderer } from '../../common/ipc'
 import { FileBrowserEntry, FileBrowserEntryType } from './file-browser-types'
 
 const ipcRenderer = new TypedIpcRenderer()
@@ -13,34 +13,43 @@ function splitExtension(filename: string): [base: string, ext: string] {
 }
 
 export default async function readFolder(folderPath: string): Promise<FileBrowserEntry[]> {
-  try {
-    const entries = (await ipcRenderer.invoke('fsReadDir', folderPath, { withFileTypes: true }))!
-    return await Promise.all(
-      entries.map(async entry => {
-        const isFolder = entry.isDirectory
-        const [name, extension] = isFolder ? [entry.name, ''] : splitExtension(entry.name)
-        const stats = await ipcRenderer.invoke('fsStat', folderPath + '\\' + entry.name)
-        const path = folderPath + '\\' + entry.name
-
-        return isFolder
-          ? {
-              type: FileBrowserEntryType.Folder,
-              name,
-              path,
-            }
-          : {
-              type: FileBrowserEntryType.File,
-              name,
-              path,
-              extension: extension.toLowerCase(),
-              date: stats?.mtime ?? new Date(0),
-            }
-      }),
-    )
-  } catch (err) {
-    if ((err as any).code === 'ENOENT') {
+  const readDirResult = (await ipcRenderer.invoke('fsReadDir', folderPath, {
+    withFileTypes: true,
+  }))!
+  if (readDirResult.error) {
+    if (readDirResult.code === FsErrorCode.FileOrFolderMissing) {
       throw new Error('Filepath ' + folderPath + " doesn't exist")
+    } else {
+      throw new Error('Unexpected FsError: ' + readDirResult.code)
     }
-    throw err
   }
+  return await Promise.all(
+    readDirResult.entries.map(async entry => {
+      const isFolder = entry.isDirectory
+      const [name, extension] = isFolder ? [entry.name, ''] : splitExtension(entry.name)
+      const statsResult = await ipcRenderer.invoke('fsStat', folderPath + '\\' + entry.name)!
+      if (statsResult.error) {
+        if (statsResult.code === FsErrorCode.FileOrFolderMissing) {
+          throw new Error('Filepath ' + folderPath + " doesn't exist")
+        } else {
+          throw new Error('Unexpected FsError: ' + statsResult.code)
+        }
+      }
+      const path = folderPath + '\\' + entry.name
+
+      return isFolder
+        ? {
+            type: FileBrowserEntryType.Folder,
+            name,
+            path,
+          }
+        : {
+            type: FileBrowserEntryType.File,
+            name,
+            path,
+            extension: extension.toLowerCase(),
+            date: statsResult.stats.mtime ?? new Date(0),
+          }
+    }),
+  )
 }
--- a/common/ipc.ts
+++ b/common/ipc.ts
@@ -53,6 +53,30 @@ export interface FsStats {
   birthtime: Date
 }
 
+export enum FsErrorCode {
+  FileOrFolderMissing,
+}
+
+export interface FsError {
+  error: true
+  code: FsErrorCode
+}
+
+export interface FsReadDirResult {
+  entries: FsDirent[]
+  error?: false
+}
+
+export interface FsReadFileResult {
+  file: ArrayBuffer
+  error?: false
+}
+
+export interface FsStatResult {
+  stats: FsStats
+  error?: false
+}
+
 /** RPCs that can be invoked by the renderer process to run code in the main process. */
 interface IpcInvokeables {
   activeGameStartWhenReady: (gameId: string) => void
@@ -61,10 +85,13 @@ interface IpcInvokeables {
 
   // TODO(tec27): Support the non-filetypes version if we need it, overloads don't seem to work
   // well with the current approach for typing these invokes =/
-  fsReadDir: (dirPath: string, options: { withFileTypes: true }) => Promise<FsDirent[]>
+  fsReadDir: (
+    dirPath: string,
+    options: { withFileTypes: true },
+  ) => Promise<FsReadDirResult | FsError>
   // TODO(tec27): Add types for options + returning a string if encoding is specified?
-  fsReadFile: (filePath: string) => Promise<ArrayBuffer>
-  fsStat: (filePath: string) => Promise<FsStats>
+  fsReadFile: (filePath: string) => Promise<FsReadFileResult | FsError>
+  fsStat: (filePath: string) => Promise<FsStatResult | FsError>
 
   logMessage: (level: string, message: string) => void
 
--- a/client/maps/upload.ts
+++ b/client/maps/upload.ts
@@ -1,4 +1,4 @@
-import { TypedIpcRenderer } from '../../common/ipc'
+import { FsErrorCode, TypedIpcRenderer } from '../../common/ipc'
 import { fetchJson } from '../network/fetch'
 
 const ipcRenderer = new TypedIpcRenderer()
@@ -14,11 +14,18 @@ function getExtension(filePath: string) {
 
 export async function upload<T>(filePath: string, apiPath: string): Promise<T> {
   const extension = getExtension(filePath)
-  const file = await ipcRenderer.invoke('fsReadFile', filePath)!
+  const readFileResult = await ipcRenderer.invoke('fsReadFile', filePath)!
+  if (readFileResult.error) {
+    if (readFileResult.code === FsErrorCode.FileOrFolderMissing) {
+      throw new Error('Filepath ' + filePath + " doesn't exist")
+    } else {
+      throw new Error('Unexpected FsError: ' + readFileResult.code)
+    }
+  }
 
   const formData = new FormData()
   formData.append('extension', extension)
-  formData.append('file', new Blob([file]))
+  formData.append('file', new Blob([readFileResult.file]))
 
   const fetchParams = {
     method: 'post',


client/file-browser/get-files.ts
```diff
@@ -13,34 +13,36 @@ function splitExtension(filename: string): [base: string, ext: string] {
...
-      throw new Error('Filepath ' + folderPath + " doesn't exist")
-    }
-    throw err
+  const readDirResult = (await ipcRenderer.invoke('fsReadDir', folderPath, {
+    withFileTypes: true,
+  }))!
+  if (readDirResult.isError && readDirResult.errorCode !== FsErrorCode.ENOENT) {
+    throw new Error('Unexpected FsError: ' + readDirResult.errorCode)
   }
+  const entries: FsDirent[] = readDirResult.isError ? [] : readDirResult.payload!
```
williamhammond, line 22: This is the biggest divergence in behavior, instead of returning an error on a non-existent directory. For it's existing usage in `file-browser`, I think that's correct but I could see it not being a safe assumption in the future.

common/ipc.ts
```diff
@@ -23,6 +23,11 @@ const ipcRenderer =
     : null
 const ipcMain = IS_ELECTRON && !IS_RENDERER ? require('electron').ipcMain : null
 
+export interface FsResult<T> {
```
2Pacalypse-, line 26: I'm not a big fan of defining types in this way, where you have multiple fields which are exclusionary with one another. These functions that we're typing here return an error with an error code _OR_ they return a payload for that function. Mixing those fields together in one type usually means you have to make some or all of them optional, which makes the types less correct. If the `fsStat` method returns successfully it needs to have a certain payload that is not optional. Another benefit would be not having this ambiguous property `payload`, and instead have a proper name for field(s) each function returns.

I would prefer to have something similar to what I showed on the Discord, or to at least try to make it work with what tec suggested, by using the user defined type guards that you linked from TS documentation. For what it's worth, I think we're using the former pattern more frequently throughout the code.

2Pacalypse- (changes requested):

common/ipc.ts
williamhammond: The issue I ran into with the solutions we were talking about in the discord was that in the case of `fsReadDir`, and `fsReadFile`, I couldn't add an `isError` field without creating a new result type to wrap `ArrayBuffer` and `FsDirent[]`. I figured since I needed to wrap those types, I could consolidate all the calls into a consistent API.

What do you think about this https://github.com/ShieldBattery/ShieldBattery/pull/939/commits/16b68b58710aeb7d0bd78df01c9dac882d943991? By using the type union, we know at compile time if the result has an error or has a value. It'll still have the issue of an ambiguous field name, but with context and type information it should be pretty clear most the time

I still need to cleanup some stuff, but I wanted to give you a chance to give the general API 👍/👎

2Pacalypse-: >The issue I ran into with the solutions we were talking about in the discord was that in the case of fsReadDir, and fsReadFile, I couldn't add an isError field without creating a new result type to wrap ArrayBuffer and FsDirent[].

I don't think this is really an issue at all. In fact, I would consider this a benefit since the type is very simple, you immediately know what it contains from its property names, and it's easy to extend in the future (in case we need to add something to it). Also, if you make the `error` (or `isError`) property optional like tec suggested in the Discord, it becomes even simpler to use.

Trying to consolidate types here seem like a more complex solution to me with very little benefit.

I've updated my [TypeScript playground](https://www.typescriptlang.org/play?#code/KYOwrgtgBAYgzgUQE5IPZIMKoCbCgbwFgAoKKAWQEs45KQBzGSgGzwF4oByKmuxl4JwA0JMj1oMYqZriRQO3ahMbTZwkgF8SJOgBdgSAGYBDAMZ4AQsbjB4AJWBwwzXQVFQDaJAH4AXFAAjVGlgYxBNbWI9AxNzWEQUdA8AD30QbDgoKxt7R2dXIlIPRKR-XSQwYHdTHGB-eGQvLFwI4h0QfSMzPFzjbAARSjlgVNAMrOtbOAcnFzciz3Q-KBNmG3dDATh-OHK+AG0AXVb2ztie6dDsJlYUtPHsqZn8+bJFn39V9aLN1n8AQRQxgAnhYwIZDAYTsRDGAQKZdJRUCAVpc+oMkAAKACU9TRAyGUAAPvFGklCmRKIYoJiAIzY15kKBIYC6MBIFEUplkX6OfxHERFMhaIoaDxrPBcsgstkcxnc95lCrAQXcsg1XB4smYWoAOnEfCkMgMqqZIuF0Nh8MRyNRDj6N2AOLx9uuAmJpJKjKpNPp8uZrPZnPc3N5-hAwAA7lBAUgQWCIQY6djTRbReKbP6ZUH-W8SkrKqmmRq6p6mnqDZIBEXzVARSKSDUQLsA7NdLT5HarhicZEfZiWW3abr3gyuQB6cet-Id6hlpIACwMVSKTbgIV1zFQ9AHeRcw5L2M0Gcl7kn0-3UDnvQJciXLOqyPXrE3293Q91vLgR+IDeIa9cQd8gAJk7Qx8UdXs2iial3xAkcSjHM8pyAlxQOvBIvCge8V3VJ8Ny3HdUN0YDdUPY9gAlRlz2I9DMhvR1sOXR9mwIt9aM-AQfxFIA) to show how it would look like for both the `fsReadDir` and `fsReadFile`.

williamhammond: Fair enough. I updated let me know if it's good and I"ll squash the commits.

A small annoyance is that it looks like Promise is confusing the structural typing with the type union so I have to manually set them to variables before returning. He's a [playground ](https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAMwM4AUoRMJBTJj0hwDkyA9GAVrkQLABQ92AdgK6EBiSAolOlAMIQAJtjgBvenDgBZLEmBMA5u2AAbUQF5ispPKUr1RADSSZchcoiqRUOFqI69l69ijH6AX3r0FMVwgBDAGNRACEAnE4AJVwWVXgJOilXPgB+AC44ACMIK2wApk9vOl9-YNFOHj44bAAPPyYhPHDIpBikOITTFOhMmCgWbFMg4WxMyt5oQREihhKmPyhAkLho-KEAEWBbOoamuBbsNY748W7JqAzEANUcUwQ1XEykfosAbQBdWZ8FspW1gJCAyiXbMfaHY6dM5JGoXK6BW5DGEPdSZACCvACAE9QiwEAhXN86AgWEwgjBgBAmIg2ustlAABQASkyaAwVAAPADNts4AAfVbcC4APmhUmACDgDIAjEyxVI4FBsDAWFBqYkFQqUU84J8TDCpF4YR4aoj5VIlSq1eaFT0oH0Bth9ZqpCMROMhXxptgAHSOCzsKw2Z0Ko2GokkskUqk0mKA4HM1mUHBc2nxx6igVszApiZ8UUauASqWym2W1Xq0ya7WZJjYADucAxUGxuPxrhlTJD4eNppwZeVFZtyQuDsG3YVbrGgqqU1GfvM+keE7DcCNRvoIyYL0VsXi0rssbp22ZxWLDKVJxg0p9drlhdIpF3V4PWBnFzgAAtXEjXVSkHkPqqBAigXnu14+lOTKeH2ogPk+l6dK+eDcvSX4-sM-6AcBoGIfuPrakg0F0BudBbjueEwAATIeyBxkCjynnM56UVRt4XPepiPs+nQ0W+ebQOhSqYdu2EgWBV5sVBMHYGa8E8fEfEoWmDHqEJv5wORYm4eBbHasRRpAA) illustrating it so you can see the error

williamhammond: Without manually specifying error on the result types, I'd need to add a https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards for each I think, or have to cast

2Pacalypse-: >A small annoyance is that it looks like Promise is confusing the structural typing with the type union so I have to manually set them to variables before returning. He's a [playground ](https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAMwM4AUoRMJBTJj0hwDkyA9GAVrkQLABQ92AdgK6EBiSAolOlAMIQAJtjgBvenDgBZLEmBMA5u2AAbUQF5ispPKUr1RADSSZchcoiqRUOFqI69l69ijH6AX3r0FMVwgBDAGNRACEAnE4AJVwWVXgJOilXPgB+AC44ACMIK2wApk9vOl9-YNFOHj44bAAPPyYhPHDIpBikOITTFOhMmCgWbFMg4WxMyt5oQREihhKmPyhAkLho-KEAEWBbOoamuBbsNY748W7JqAzEANUcUwQ1XEykfosAbQBdWZ8FspW1gJCAyiXbMfaHY6dM5JGoXK6BW5DGEPdSZACCvACAE9QiwEAhXN86AgWEwgjBgBAmIg2ustlAABQASkyaAwVAAPADNts4AAfVbcC4APmhUmACDgDIAjEyxVI4FBsDAWFBqYkFQqUU84J8TDCpF4YR4aoj5VIlSq1eaFT0oH0Bth9ZqpCMROMhXxptgAHSOCzsKw2Z0Ko2GokkskUqk0mKA4HM1mUHBc2nxx6igVszApiZ8UUauASqWym2W1Xq0ya7WZJjYADucAxUGxuPxrhlTJD4eNppwZeVFZtyQuDsG3YVbrGgqqU1GfvM+keE7DcCNRvoIyYL0VsXi0rssbp22ZxWLDKVJxg0p9drlhdIpF3V4PWBnFzgAAtXEjXVSkHkPqqBAigXnu14+lOTKeH2ogPk+l6dK+eDcvSX4-sM-6AcBoGIfuPrakg0F0BudBbjueEwAATIeyBxkCjynnM56UVRt4XPepiPs+nQ0W+ebQOhSqYdu2EgWBV5sVBMHYGa8E8fEfEoWmDHqEJv5wORYm4eBbHasRRpAA)illustrating it so you can see the error

I think the problem here is returning a Promise from a non-async function. If you make that function async, the error should go away, e.g. [TS playground](https://www.typescriptlang.org/play?#code/KYOwrgtgBAYgzgUQE5IPZIMKoCbCgbwFgAoKKAWQEs45KQBzGSgGzwF4oByKmuxl4JwA0JMj1oMYqZriRQO3ahMbTZwkgF8SJOgBdgSAGYBDAMZ4AQsbjB4AJWBwwzXQVFQDaJAH4AXFAAjVGlgYxBNbWI9AxNzWEQUdA8AD30QbDgoKxt7R2dXIlIPRKR-XSQwYHdTHGB-eGQvLFwI4h0QfSMzPFzjbAARSjlgVNAMrOtbOAcnFzciz3Q-KBNmG3dDATh-OHK+AG0AXVb2ztie6dDsJlYUtPHsqZn8+bJFn39V9aLN1n8AQRQxgAnhYwIZDAYTsRrMCQKYVmB4bpKKgQCtLn1BkgABQASn8AAU0BBqMAADy9AZDKAAH3ijXQAD5XlBKIYoDiAIx41lkJDAXRgJDowpkcUrLb+I4iIpkLRFDQeNZ4MXigVCkV88XvMoVYCyiVkGq4eoJJq1AB04j4UhkBkN4oV8uhsPhiORqPRhkx1wE+KJJLJlN9Nzw9IaJRZavZnJ52o1wtF7glvzqUBAwAA7lBAUgQWCIQZuXjHS7FcqbAnBUntW8SnrKmXxSb05GLbhrUpbQJm86oAqFSQaiBdlABbNdFz5FBjFnjJRXD6HFihvjIrGcRP8lzLe9eWqAPSH8d5FzT6gMkpQAAWBiqRRHcBCluYqHoW7PU8trbxmkrqruMep6ThemRUtit73tUaLPqwr7vp+oGWmmcB-sQQ7EE+rjbi4ABMM5zguS6hv66E6BySH5Hhe4lAeQEnrhugEZe7ZJHeAowaOL5vh+TE0b+-7ACqrLAfxbLgaRtwcQ+xqwTxiH8ShAjoZhIwAA7oAUCpAA)

williamhammond: whoops, i should have looked at the playground error more closely. This is the error in the actual code. It looks like it's trying to combine the two return objects when building the structural type information? 

```
TS2345: Argument of type '(_: IpcMainInvokeEvent, dirPath: string, options: { withFileTypes: true; }) => Promise<{ files: FsDirent[]; error: boolean; code?: undefined; } | { code: FsErrorCode; error: boolean; files?: undefined; }>' is not assignable to parameter of type '(event: IpcMainInvokeEvent, dirPath: string, options: { withFileTypes: true; }) => Promisable<Promise<FsDirents> | Promise<FsError>>'.   Type 'Promise<{ files: FsDirent[]; error: boolean; code?: undefined; } | { code: FsErrorCode; error: boolean; files?: undefined; }>' is not assignable to type 'Promisable<Promise<FsDirents> | Promise<FsError>>'.     Type 'Promise<{ files: FsDirent[]; error: boolean; code?: undefined; } | { code: FsErrorCode; error: boolean; files?: undefined; }>' is not assignable to type 'Promise<FsDirents>'.       Type '{ files: FsDirent[]; error: boolean; code?: undefined; } | { code: FsErrorCode; error: boolean; files?: undefined; }' is not assignable to type 'FsDirents'.         Type '{ files: FsDirent[]; error: boolean; code?: undefined; }' is not assignable to type 'FsDirents'.           Types of property 'error' are incompatible.             Type 'boolean' is not assignable to type 'false'.
```

client/file-browser/get-files.ts
```diff
@@ -13,34 +13,36 @@ function splitExtension(filename: string): [base: string, ext: string] {
...
-    )
-  } catch (err) {
-    if ((err as any).code === 'ENOENT') {
-      throw new Error('Filepath ' + folderPath + " doesn't exist")
-    }
-    throw err
+  const readDirResult = (await ipcRenderer.invoke('fsReadDir', folderPath, {
+    withFileTypes: true,
+  }))!
+  if (readDirResult.error && readDirResult.code !== FsErrorCode.ENOENT) {
```
2Pacalypse-, line 19: You need to handle all errors that `fsReadDir` might return in its results here, not just specific ones. This will also allow you not needing to check for the errors below when using the entries.

Also, this still needs to be inside a try-catch block, to handle all the regular errors that we're not special-handling.

client/file-browser/get-files.ts
```diff
@@ -13,34 +13,36 @@ function splitExtension(filename: string): [base: string, ext: string] {
...
+    throw new Error('Unexpected FsError: ' + readDirResult.error)
   }
+  const entries: FsDirent[] = readDirResult.error ? [] : readDirResult.files
```
2Pacalypse-, line 22: If you do the above thing then you won't need to check for the error here anymore and this can simply be:
```ts
const entries = readDirResult.files
```
Or you can delete the line altogether then and just use the full name when doing the mapping.

Also, now that I think about it, it might make more sense to name the property inside the `FsReadDirResult` `entries`, instead of `files`, since that API returns both folders and files.

app/app.ts
```diff
@@ -396,32 +405,77 @@ function setupIpc(localSettings: LocalSettings, scrSettings: ScrSettings) {
...
+        atime: result.atime,
+        mtime: result.mtime,
+        ctime: result.ctime,
+        birthtime: result.birthtime,
+        error: false,
+      }
+      return stat
+    } catch (e) {
+      if ((e as any).code === 'ENOENT') {
+        const error: FsError = {
```
2Pacalypse-, line 472: You can just return the object directly instead of creating a dedicated variable for it, e.g.
```ts
return {
  error: true,
  code: FsErrorCode.FilepathNotFound,
}
```

app/app.ts
```diff
@@ -396,32 +405,77 @@ function setupIpc(localSettings: LocalSettings, scrSettings: ScrSettings) {
...
-      mtimeMs: result.mtimeMs,
-      ctimeMs: result.ctimeMs,
-      birthtimeMs: result.birthtimeMs,
-      atime: result.atime,
-      mtime: result.mtime,
-      ctime: result.ctime,
-      birthtime: result.birthtime,
+    try {
+      const result = await fsPromises.stat(filePath)
+      const stat: FsStats = {
```
2Pacalypse-, line 453: Same as above, you can just return this directly.

app/app.ts
```diff
@@ -396,32 +405,77 @@ function setupIpc(localSettings: LocalSettings, scrSettings: ScrSettings) {
...
+        name: d.name,
+      }))
+      const dirents: FsDirents = {
+        files,
+        error: false,
+      }
+      return dirents
+    } catch (e) {
+      if ((e as any).code === 'ENOENT') {
+        const error: FsError = {
```
2Pacalypse-, line 441: Same as above, you can just return this directly.

app/app.ts
```diff
@@ -396,32 +405,77 @@ function setupIpc(localSettings: LocalSettings, scrSettings: ScrSettings) {
...
-      name: d.name,
-    }))
+    try {
+      const result = await fsPromises.readdir(dirPath, options)
+      const files = result.map<FsDirent>(d => ({
+        isFile: d.isFile(),
+        isDirectory: d.isDirectory(),
+        name: d.name,
+      }))
+      const dirents: FsDirents = {
```
2Pacalypse-, line 434: Same as above, you can just return this directly.

app/app.ts
```diff
@@ -396,32 +405,77 @@ function setupIpc(localSettings: LocalSettings, scrSettings: ScrSettings) {
...
+        files,
+        error: false,
```
2Pacalypse-, line 436: If you make the `error` property optional on the result type, then you can omit this.

app/app.ts
```diff
@@ -396,32 +405,77 @@ function setupIpc(localSettings: LocalSettings, scrSettings: ScrSettings) {
...
+    try {
+      const arrayBuffer = await fsPromises.readFile(filePath)
+      const result: FsFile = {
+        file: arrayBuffer,
+        error: false,
+      }
+      return result
+    } catch (e) {
+      if ((e as any).code === 'ENOENT') {
+        const error: FsError = {
```
2Pacalypse-, line 417: Same as above, you can just return this directly.

app/app.ts
```diff
@@ -396,32 +405,77 @@ function setupIpc(localSettings: LocalSettings, scrSettings: ScrSettings) {
   )
 
   ipcMain.handle('fsReadFile', async (_, filePath) => {
-    return fsPromises.readFile(filePath)
+    try {
+      const arrayBuffer = await fsPromises.readFile(filePath)
+      const result: FsFile = {
```
2Pacalypse-, line 410: Same as above, you can just return this directly.

2Pacalypse- (changes requested):

common/ipc.ts
2Pacalypse-: I'm not sure where exactly do you get this error, but maybe it will go away once you address the other comments I left now 🤞

common/ipc.ts
```diff
@@ -33,6 +40,16 @@ export interface FsDirent {
...
 }
 
+export interface FsDirents {
+  files: FsDirent[]
+  error: false
+}
+
+export interface FsFile {
+  file: ArrayBuffer
+  error: false
```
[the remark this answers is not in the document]
williamhammond: If it's optional, the type inference fails.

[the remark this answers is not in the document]
2Pacalypse-: It probably fails because of something else then, since it works in the TS playground just fine. I'll have to take a closer look.

client/file-browser/get-files.ts
williamhammond: This line is checking for non-ENOENT errors, and the next line handles the ENOENT error case, so all errors are covered. It is a little odd though since error handling is mixed between throwing and error codes. I wonder If I should go ahead and add all the error codes to the enum? That'll lead to a lot of boiler plate for what will end up being a small number of special cases.

Currently, every error other than ENOENT is still throwing so since previously the try/catch rethrew errors the behavior is the same.

common/ipc.ts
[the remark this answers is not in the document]
williamhammond: Oh whoops, I think I mixed it up with the return type issue. I readded the optional and it worked fine

client/file-browser/get-files.ts
2Pacalypse-: Yeah, it's a weird way to handle errors. But you have to remember that we're only doing this because Electron doesn't allow us to do this through normal error handling :d

I definitely don't think you should go ahead and add all FS errors. The idea is to only handle the errors that we care about through this system, and have some general error handling for all other cases. Actually now that I think about it, you don't need to use the try-catch here anymore, since the catch block would just rethrow the error.

So the only thing I would change here is to handle all potential errors we care about, e.g.
```ts
if (readDirResult.error) {
  if (readDirResult.code === FsErrorCode.FolderNotFound) {
    throw new Error('Folder not found')
  }
  throw new Error('Unexpected FsError: ' + readDirResult.error)
}
```

williamhammond: ah sorry, my line comment here must have got swallowed at some point when I was making changes. 

This is throwing on all none missing file errors
https://github.com/ShieldBattery/ShieldBattery/pull/939/files#diff-be7de2630127e1ca9494b92af4ef574add9d864f56577fb428bef0b0d57fafbbR19

and this is setting the list of entries to be `[]` if there is no folder

https://github.com/ShieldBattery/ShieldBattery/pull/939/files#diff-be7de2630127e1ca9494b92af4ef574add9d864f56577fb428bef0b0d57fafbbR22

This explains why I was doing that https://github.com/ShieldBattery/ShieldBattery/issues/936. This is fine for the current usages of readFile, but i'm not sure if it's a safe thing to do in general

2Pacalypse-: Yeah, that doesn't sound like a great idea to do here. I'm not sure we want to return an empty array every time we try to read a folder that doesn't exist.

You could potentially extend this error you're throwing here with our custom error code so you can check it in place it's used, in case you want to handle that error specially at that place. But I'm not sure that #936 is even the best way to handle that case in the first place. If you're missing the root replay folder, shouldn't we let the users know that and maybe how they can fix that? I guess it's potentially fine doing what you suggested too, if the user doesn't really need to do anything to fix it since the folder will be created when they play their first game?

But in any case, I would probably add the error code to the error you're throwing here, so the file-browser can check that code and do whatever it wants when the folder doesn't exist.

williamhammond: Alright I updated to have the same behavior as before.

> If you're missing the root replay folder, shouldn't we let the users know that and maybe how they can fix that? 

It doesn't exist for fresh installs. I happened to run into the error after setting up a new PC.

common/ipc.ts
williamhammond: Full Error:

```
TS2345: Argument of type '(_: IpcMainInvokeEvent, dirPath: string, options: { withFileTypes: true; }) => Promise<{ entries: FsDirent[]; code?: undefined; error?: undefined; } | { code: FsErrorCode; error: boolean; entries?: undefined; }>' is not assignable to parameter of type '(event: IpcMainInvokeEvent, dirPath: string, options: { withFileTypes: true; }) => Promisable<Promise<FsReadDirResult | FsError>>'.   Type 'Promise<{ entries: FsDirent[]; code?: undefined; error?: undefined; } | { code: FsErrorCode; error: boolean; entries?: undefined; }>' is not assignable to type 'Promisable<Promise<FsReadDirResult | FsError>>'.     Type 'Promise<{ entries: FsDirent[]; code?: undefined; error?: undefined; } | { code: FsErrorCode; error: boolean; entries?: undefined; }>' is not assignable to type 'Promise<FsReadDirResult | FsError>'.       Type '{ entries: FsDirent[]; code?: undefined; error?: undefined; } | { code: FsErrorCode; error: boolean; entries?: undefined; }' is not assignable to type 'FsReadDirResult | FsError'.         Type '{ code: FsErrorCode; error: boolean; entries?: undefined; }' is not assignable to type 'FsReadDirResult | FsError'.           Type '{ code: FsErrorCode; error: boolean; entries?: undefined; }' is not assignable to type 'FsError'.             Types of property 'error' are incompatible.               Type 'boolean' is not assignable to type 'true'.
``` 

If i change `error` to boolean for both the result types and error i'm able to return the object without the variable, but that breaks type inference when it's called and we'd need to cast. I can also return one without a variable but not the other without the type checker complaining 😅 .

I can also hoist the (lambda?) into a helper function and do something like 

```
  async function fsStat(filePath: string): Promise<FsStatResult | FsError> {
    try {
      const result = await fsPromises.stat(filePath)
      return {
        stats: {
          isFile: result.isFile(),
          isDirectory: result.isDirectory(),
          size: result.size,
          blksize: result.blksize,
          blocks: result.blocks,
          atimeMs: result.atimeMs,
          mtimeMs: result.mtimeMs,
          ctimeMs: result.ctimeMs,
          birthtimeMs: result.birthtimeMs,
          atime: result.atime,
          mtime: result.mtime,
          ctime: result.ctime,
          birthtime: result.birthtime,
        },
      }
    } catch (e) {
      if ((e as any).code === 'ENOENT') {
        return {
          error: true,
          code: FsErrorCode.FileOrFolderMissing,
        }
      }
      throw e
    }
  }
  ipcMain.handle('fsStat', async (_, filePath) => fsStat(filePath))
  ```

app/app.ts
2Pacalypse-: Ah ok, sorry I assumed that resolved comments means they're handled and pushed.

common/ipc.ts
2Pacalypse-: Yeah, I think this will require tec or me (probably tec :d) playing around with this to see what's wrong. It's probably caused by something how the IPC types are setup in the first place.

tec27: Can you check in code with the ideal path for this stuff (even though the typechecker will complain, instead of trying a bunch of non-ideal things to work around this issue) so I can take a look?

PR #940, opened 2023-05-10
Add automatic language detection and a way to change language for signed

2Pacalypse-:
out users.

The detected language is first detected from the "navigator" settings and saved into the local storage and is not currently being sent to the server during the signup process. That will come in the next commit.

User can change the language on all of the pages that render top-links; basically splash/faq/login/signup.

### DIFF
--- a/client/i18n/i18next.ts
+++ b/client/i18n/i18next.ts
@@ -25,8 +25,7 @@ export const i18nextPromise = i18n.use(HttpBackend).use(initReactI18next)
 
 if (LANGUAGE_SUPPORT) {
   const languageDetector = new LanguageDetector(null, {
-    order: ['localStorage', 'navigator'],
-    caches: ['localStorage'],
+    order: ['navigator'],
   })
 
   i18n.use(languageDetector)
--- a/client/i18n/i18next.ts
+++ b/client/i18n/i18next.ts
@@ -1,6 +1,8 @@
 import i18n from 'i18next'
+import LanguageDetector from 'i18next-browser-languagedetector'
 import HttpBackend, { HttpBackendOptions } from 'i18next-http-backend'
 import { initReactI18next } from 'react-i18next'
+import { LANGUAGE_SUPPORT } from '../../common/flags'
 import {
   ALL_TRANSLATION_LANGUAGES,
   TranslationLanguage,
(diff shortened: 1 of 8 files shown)


client/i18n/i18next.ts
```diff
@@ -10,34 +12,42 @@ import { makeServerUrl } from '../network/server-url'
...
+i18n.init<HttpBackendOptions>({
+  backend: {
+    loadPath: makeServerUrl('/locales/{{lng}}/{{ns}}.json'),
+    addPath: makeServerUrl('/locales/add/{{lng}}/{{ns}}'),
+  },
+
+  saveMissing: isDev,
+  saveMissingTo: 'all', // Save the missing keys to all languages
+
+  supportedLngs: ALL_TRANSLATION_LANGUAGES,
```
tec27, line 35: If you want the feature flag to work you need to make this list contain only English don't you?

client/landing/top-links.tsx
```diff
@@ -78,42 +97,111 @@ const HideWhenSmall = styled.span`
   }
 `
 
-const TopLinks = () => {
+const HideWhenElectron = styled.li`
+  visibility: ${IS_ELECTRON ? 'hidden' : 'visible'};
```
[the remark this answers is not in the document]
tec27: Also honestly feels like it might be easier to just not render these things in that case?
```tsx
{IS_ELECTRON ? <>
  <Link href='/splash'>Home</Link>
  <Link href='/faq'>FAQ</Link>
  <Link href='/ladder'>Ladder</Link>
</> : // ...
}
```

Think you overcomplicated things with the CSS approach, the value can't change at runtime anyway (which is why the sizing one was done this way)

tec27 (changes requested):

client/i18n/i18next.ts
tec27: And also probably remove the caches and such? Because otherwise if we were to ship this it would be setting people's languages to English even if there's a better language for them

client/landing/top-links.tsx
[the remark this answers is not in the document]
2Pacalypse-: Yeah, the only reason I used the hidden visibility is because I thought the top-links bar looked kind of weird with just 2 centered links, but maybe it's fine :d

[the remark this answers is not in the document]
tec27: Well the `Spacer` element should deal with that already shouldn't it? Like that should take up whatever space is left to push things to the right still

client/i18n/i18next.ts
2Pacalypse-: >If you want the feature flag to work you need to make this list contain only English don't you?

Hmm I thought it was fine because the feature flag prevents users to change their language and it disables language detector (so all users will basically use our fallback language, i.e. English).

client/landing/top-links.tsx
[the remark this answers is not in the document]
2Pacalypse-: Oh you're right, I wasn't rendering the spacer either so the links just got centered.

client/i18n/i18next.ts
2Pacalypse-: >And also probably remove the caches and such? Because otherwise if we were to ship this it would be setting people's languages to English even if there's a better language for them

Just tested this and it seems not to do anything? What the hell i18next... I'll have to investigate this. But anyway, removing the cache and the localStorage from options still adds an entry for the selected language to local storage for me.

tec27: Yeah, if the language detector is disabled and nothing gets saved then that's fine. I don't know that we particularly need a feature flag for this anyway as it seems to all be implemented in this commit *anyway* but :)

(This feature flag is not useful for turning individual languages off anyway, so it being a "wait for more translations" thing isn't really useful)

2Pacalypse-: My main idea behind the feature flag was to not have an UI that allows users to select a language and then still seeing the content in English :p

Also, I was thinking we could merge this PR as is, and I would create a new PR for sending the language to login/signup APIs and saving it to the DB (and another PR after that for displaying that language in the settings page and being able to change it).

tec27: Yeah but for what purpose exactly? The point of feature flags is to be able to implement things in pieces and get them checked in, even though parts of them are broken/non-working. All of these pieces work on their own once checked in, so I'm not really sure why we're adding the complexity of maintaining 2 implementations for all these things

tec27: The flag doesn't help with the problem you stated (selecting a language and still seeing content in English). That will continue happening ~forever, every time we add new strings, those strings will be visible as English if they have not yet been translated. Every time we add a new language its strings will all be in English until they get translated. The feature flag system cannot fix this issue, it is not designed for that sort of problem

2Pacalypse-: Yes, all fair points :d
I'll take the feature flag out.

client/landing/top-links.tsx
```diff
@@ -78,42 +99,84 @@ const HideWhenSmall = styled.span`
...
+  const dispatch = useAppDispatch()
+  const [languageMenuOpen, openLanguageMenu, closeLanguageMenu] = usePopoverController()
+  const [anchor, anchorX, anchorY] = useAnchorPosition('left', 'bottom')
+
+  const onChangeLanguage = useStableCallback(async (language: TranslationLanguage) => {
+    closeLanguageMenu()
+    try {
+      await i18n.changeLanguage(language)
+    } catch (error) {
+      dispatch(openSnackbar({ message: 'Something went wrong changing the language' }))
```
tec27, line 113: This should be a translated string.

Also probably: "Something went wrong when changing the language"

tec27: (If you have another PR in the works for translating the rest of the content here, it's fine to leave the translation until then, but probably good to get in the habit of translating everything from here on out)

tec27 (approved):

PR #942, opened 2023-05-14
Save the user language to the DB on server.

2Pacalypse-:
Currently, the language is only saved when users create a new account or
they login with their existing account. Next PR will allow users to
change the language through the settings UI in the app.

Also moves the signup and login pages to TS/hooks to make things easier.


2Pacalypse-:
Depends on #940

common/users/sb-user.ts
```diff
@@ -38,6 +39,8 @@ export interface SelfUser extends SbUser {
   acceptedTermsVersion: number
   /** The last version of the acceptable use policy this user has seen/accepted. */
   acceptedUsePolicyVersion: number
+  /** The language that this user has chosen to use. */
+  language: TranslationLanguage
```
tec27, line 43: I'd be a little careful assuming this matches a language we actually have active. I don't know what i18next is doing, but imo it makes sense to detect a language and save that info even if we don't currently support it, so that if we support it in the future we can auto-switch to a better experience for that person. That implies that the value we save in the DB is just a locale code, not that it is a `TranslationLanguage` (so this should probably be typed as a string and you could have some method for checking/converting it to a `TranslationLanguage` if needed)

server/lib/session/session-api.ts
```diff
@@ -96,10 +98,11 @@ export class SessionApi {
         // TODO(tec27): Make this required in future versions (cur v8.0.2). This is just to allow
         // old clients to log in so it triggers auto-update
         clientIds: joiClientIdentifiers(),
+        language: Joi.valid(...ALL_TRANSLATION_LANGUAGES),
```
tec27, line 101: As stated in the user type, I don't think it makes sense to validate this way. Would be better to accept any BCP-47 tag probably (but sort of depends what i18next's thing is doing). It's very possible you should just be sending whatever value i18next looks at to do "detection" instead of using their resolved value for this, as I assume the purpose of their stuff is "given a known browser language, pick from the languages that are enabled" rather than "what is the browser language"

tec27: Also I can't leave a comment on it, but imo you should be adding a query parameter to `getCurrentSession` that includes a language code, and migrate existing accounts if they don't have a language set in that

server/lib/session/session-api.ts
```diff
@@ -148,6 +151,15 @@ export class SessionApi {
     const perms = (await getPermissions(user.id))!
     await maybeMigrateSignupIp(user.id, ctx.ip)
 
+    if (language && language !== user.language) {
```
tec27, line 154: No, don't do this. If a user has set a specific language, that setting always takes precedent over whatever is auto-detected, period. Please think through user flows when deciding stuff like this:

- User logs in, sees the language in Spanish but they want English
- User goes to settings, picks English
- User logs out for whatever reason
- User opens app later, Electron picks Spanish again
- User logs in

 With your code, their setting is now Spanish again, not good.

The only time you should be updating this column is:

1) When a user logs in or retrieves a current session, and they had no detected language stored
2) When the user explicitly changes their language (via settings)

The language parameter here is not a "please change my account's language" directive, it is "here is a language to use to migrate my account if necessary"

server/migrations/20230514095648-add-language-to-users.js
```diff
@@ -0,0 +1,29 @@
+exports.up = async function (db) {
+  await db.runSql(`
+    ALTER TABLE users
+    ADD COLUMN language text;
+  `)
+
+  // Set existing users to english language
+  await db.runSql(`
```
tec27, line 8: No, existing users should have nothing set and they should be migrated when they log in, that was the whole idea behind this

server/migrations/20230514095648-add-language-to-users.js
```diff
@@ -0,0 +1,29 @@
...
+    UPDATE users
+    SET language = 'en';
+  `)
+
+  // Make the column non-null
+  await db.runSql(`
+    ALTER TABLE users
```
tec27, line 15: This column should always be nullable

tec27 (changes requested):

client/auth/action-creators.ts
```diff
@@ -4,14 +4,35 @@ import { apiUrl } from '../../common/urls'
...
+      openSimpleDialog(
+        i18n.t('auth.language.changeErrorHeader', 'Error changing the language'),
+        i18n.t(
+          'auth.language.changeErrorMessage',
+          'Something went wrong when changing the language',
+        ),
+        true,
+      ),
+    )
+    logger.error(`There was an error changing the language: ${error}`)
```
tec27, line 32: Always use `error.stack` if it's available. Generally: `${error?.stack ?? error}`

client/auth/login.tsx
```diff
@@ -0,0 +1,241 @@
...
+    abortControllerRef.current?.abort()
+    abortControllerRef.current = new AbortController()
+
+    dispatch(
+      logIn(
+        {
+          username: model.username,
+          password: model.password,
+          remember: model.rememberMe,
+          locale: i18n.language,
```
tec27, line 71: No, like I said before, the thing we should be sending to the server is *the locale as reported by the browser*. This is *the language that was chosen based on the available languages*.

You need to keep track of when the user has set an explicit language preference in this client (via the dropdown), so you can send in the following priorities:

1) The explicit language preference they picked (if nothing explicit was chosen in the dropdown, this should be undefined. I don't consider opening the dropdown but not choosing anything to be an explicit selection)
2) The current locale as reported by the browser

client/auth/login.tsx
```diff
@@ -0,0 +1,241 @@
...
+      username: queryModel.username ?? '',
+      password: '',
+      rememberMe: false,
+    },
+    {
+      username: composeValidators(
+        required(t('auth.usernameValidator.required', 'Enter a username')),
+        minLength(
+          USERNAME_MINLENGTH,
+          t('auth.usernameValidator.minLength', {
```
tec27, line 97: I really don't understand why we have different strings for this and password. They are the same string, they have the same context (they don't reference anything about the field they are on). You can likely use the same string for *every* min length validator in the app (and max length as well)

client/auth/login.tsx
```diff
@@ -0,0 +1,241 @@
...
+            defaultValue: `Enter at least {{minLength}} characters`,
+            minLength: PASSWORD_MINLENGTH,
+          }),
+        ),
+      ),
+    },
+    { onSubmit: onFormSubmit },
+  )
+
+  const onSplashClick = useStableCallback(() => {
```
tec27, line 128: Just use a link for this

client/auth/login.tsx
```diff
@@ -0,0 +1,241 @@
...
+
+  const onCreateAccountClick = useStableCallback(() => {
+    const search = queryString.stringify({
+      ...queryString.parse(location.search),
+      username: getInputValue('username'),
+    })
+    push({ pathname: '/signup', search })
+  })
+
+  const onForgotUsernameClick = useStableCallback(() => {
```
tec27, line 140: Just use a link for this

client/auth/login.tsx
```diff
@@ -0,0 +1,241 @@
...
+    push({ pathname: '/forgot-user' })
+  })
+
+  const onForgotPasswordClick = useStableCallback(() => {
```
tec27, line 144: Just use a link for this

client/auth/signup.tsx
```diff
@@ -0,0 +1,395 @@
...
+    abortControllerRef.current?.abort()
+    abortControllerRef.current = new AbortController()
+
+    dispatch(
+      signUp(
+        {
+          username: model.username,
+          email: model.email,
+          password: model.password,
+          locale: i18n.language,
```
tec27, line 183: Exactly the same issues as in login

client/forms/validators.js
```diff
@@ -66,3 +66,7 @@ export function numberRange(min, max, msg) {
     return parsed >= min && parsed <= max ? null : msg
   }
 }
+
+export function checked(msg) {
```
tec27, line 70: Call this `requireChecked`

client/index.jsx
```diff
@@ -109,10 +109,13 @@ Promise.all([rootElemPromise, i18nextPromise])
     if (IS_ELECTRON || !window._sbInitData) {
       configPromise = fetchJson('/config', { method: 'get' })
       sessionPromise = new Promise((resolve, reject) => {
-        action = getCurrentSession({
-          onSuccess: () => resolve(),
-          onError: err => reject(err),
-        })
+        action = getCurrentSession(
+          { locale: i18n.language },
```
tec27, line 113: Same issues as in login/signup

common/users/sb-user.ts
```diff
@@ -38,6 +38,11 @@ export interface SelfUser extends SbUser {
   acceptedTermsVersion: number
   /** The last version of the acceptable use policy this user has seen/accepted. */
   acceptedUsePolicyVersion: number
+  /**
+   * The locale that we have saved for this user. We will try to match user's locale to one of our
```
tec27, line 42: This is not an accurate description

Here's a more accurate one:

The locale that this user has set, either as an explicit choice or as reported by their browser. This string is a BCP 47 code that will be used locally to select from the available languages for app content, falling back to English if no better alternative can be found. This field was added after many accounts were created, and only accounts that have logged in since then will have a locale present.

server/lib/session/migrate-sessions.ts
```diff
@@ -25,6 +25,7 @@ export function migrateSessions() {
         ctx.session.acceptedPrivacyVersion = retrievedUser!.acceptedPrivacyVersion
         ctx.session.acceptedTermsVersion = retrievedUser!.acceptedTermsVersion
         ctx.session.acceptedUsePolicyVersion = retrievedUser!.acceptedUsePolicyVersion
+        ctx.session.locale = retrievedUser!.locale
```
tec27, line 28: This is not the appropriate block to put this in, you've written code that will only add locale info to the session if the session did not have `acceptedPrivacyVersion`

You need to put this in its own if statement (outside of the one you put it in). And then you should actually *test* the code using a session that doesn't have locale info in it yet, instead of just assuming that everything you have written works as you expected

server/lib/session/session-api.ts
```diff
@@ -46,6 +47,15 @@ export class SessionApi {
 
   @httpGet('/')
   async getCurrentSession(ctx: RouterContext): Promise<ClientSessionInfo> {
+    const {
+      query: { locale },
+    } = validateRequest(ctx, {
+      query: Joi.object<{ date: number; locale?: string }>({
+        date: Joi.number().required(),
+        locale: Joi.string(),
```
tec27, line 55: I think validating these completely is kind of overly complicated and probably not necessary, but you should at least check the length imo. Max length should be 35 characters:
https://stackoverflow.com/a/17863380

server/lib/session/session-api.ts
```diff
@@ -96,10 +115,11 @@ export class SessionApi {
         // TODO(tec27): Make this required in future versions (cur v8.0.2). This is just to allow
         // old clients to log in so it triggers auto-update
         clientIds: joiClientIdentifiers(),
+        locale: Joi.string(),
```
tec27, line 118: Same as above, you could make some common validator for it

server/lib/users/user-api.ts
```diff
@@ -196,10 +197,11 @@ export class UserApi {
           .trim()
           .required(),
         clientIds: joiClientIdentifiers().required(),
+        locale: Joi.string(),
```
tec27, line 200: Same as other file

tec27 (changes requested):

client/auth/login.tsx
2Pacalypse-: There's actually two versions of the minLength and maxLength strings. One (on the login page) is:
```
Enter at least {{minLength}} characters
```
and the other one (on signup page) is:
```
Use at most {{maxLength}} characters
```
I guess the slight difference in verbiage makes sense given the context. I named the keys for thsi `minLength` and `minLength2` fwiw.

But yeah, we can at least use the same keys wherever the string is same.

server/lib/session/migrate-sessions.ts
2Pacalypse-: Oh yeah, my bad (it was late :d).

Btw, what's the point of the `user` variable in this function? It's initialized to `undefined` and not reassigned anywhere, so I feel like I'm missing something here.

client/auth/login.tsx
tec27: That's a minlength versus a maxlength, so yes, I would expect those strings to be different :) My point was that *all* minlength strings in the app can be the same, and *all* maxlength strings in the app can be the same, not that we need to combine minlength and maxlength.

What you have as "minLength" and "minLength2" are the same thing, they are just for the password and confirmation password, respectively.

server/lib/session/migrate-sessions.ts
tec27: The point of `user` is that if we have retrieved the `user` previously in this function (because multiple things needed to be migrated), we don't need to re-run that database query N times.

This function exists to fixup old sessions that did not have all the data on them. Yes you need to have handling for `locale` here.

tec27: And yes, there is a bug here, `user` should be set to `retrievedUser` in these blocks such that the intention is met, so you should fix that

client/auth/login.tsx
2Pacalypse-: Oh sorry, I copy-pasted the wrong thing :p

What I meant to indicate is that in some places we use the `Enter ...` (e.g. when logging in), and in some places we use `Use ...` (e.g. when creating a new account). Should we preserve that difference or use the same word in all cases (and which one)?

tec27: No, that difference is just weirdness in how this stuff got created/written initially, just make all the strings the same.

When you're doing this stuff, consider how we can minimize the amount of work it takes to translate the app, this is the prime concern. If there are very minor content differences that serve no real purpose and provide no actual value to the app, we should be seeking to standardize stuff so it only needs to be translated once.

client/i18n/i18next.ts
```diff
@@ -20,6 +21,16 @@ const isDev = __WEBPACK_ENV.NODE_ENV !== 'production'
  */
 export type TransInterpolation = any
 
+/**
+ * The locale that was reported to us by the user's browser. This locale can be overwritten by
+ * user's explicit choice in the top-links dropdown. We send this locale to the server during
+ * login/signup/getCurrentSession actions.
+ *
+ * NOTE(2Pac): Session storage gets cleared after logging in, so be careful about using this only in
```
tec27, line 29: We should probably stop doing that tbh, especially if we're storing anything else in there. I don't think the `storageKey` stuff really needs to be on `SessionStorage` (sharing it between tabs is irrelevant), it could just be a global `Map`

client/auth/edit-account.jsx
```diff
@@ -50,14 +50,14 @@ const emailValidator = composeValidators(
   minLength(
     EMAIL_MINLENGTH,
     t('common.validators.minLength', {
-      defaultValue: `Use at least {{minLength}} characters`,
+      defaultValue: `Enter at least {{minLength}} characters`,
```
tec27, line 53: Think it'd probably be better to have something that deals with the defaultValue and just takes in the minLength/maxLength/etc. (could even be a hook, `useMinLengthValidator`?). So that we don't need to repeat + fix this default value everywhere (even though I know it will be checked for same-ness).

This can be fixed in a different PR though

client/auth/action-creators.ts
```diff
@@ -128,7 +127,7 @@ export function logOut() {
     const result = await fetchJson<void>(apiUrl`sessions`, {
       method: 'delete',
     })
-    sessionStorage.clear()
+
```
tec27, line 130: The point of doing this before was to clear the message storage specifically, so you can't just delete this now. You need to preserve the existing semantics of the stuff that was previously using sessionStorage, namely, it all needs to get cleared when the current user changes.

(This is done in 2 places, here and the other line you deleted below, so make sure to deal with it in both)

tec27: Another alternative is to use the current user ID as part of the map key. In any case, you need to preserve the existing behavior here

tec27 (changes requested):

tec27 (approved):

PR #946, opened 2023-05-20
Use wouter's Redirect instead of our custom redirect solution.

### LINKED ISSUE #941: Logging out of the application generates the wrong `nextPath` query in URL
image

It basically runs the `doRedirect` function in LoggedInFilter twice, so the `nextPath` gets nested. I'm not really sure why it does that, or what's the correct solution.

2Pacalypse-:
Fixes #941


client/main-layout.tsx
```diff
@@ -372,11 +372,9 @@ export function MainLayout() {
       <ConnectedLeftNav />
       <Content>
         <Switch>
-          <ConditionalRoute
-            path='/admin/:rest*'
-            filters={[IsAdminFilter]}
-            component={LoadableAdminPanel}
-          />
+          <Route path='/admin/:rest*'>
```
tec27, line 375: You can just not render this Route in that case, rather than doing a redirect (not rendering the routes and being on an admin route will make it hit the fallback route here, which will redirect anyway)

tec27 (approved):

PR #948, opened 2023-05-27
Improve the settings visuals.

### LINKED ISSUE #945: Fix Settings visuals
Since I feel like this is getting lost behind other stuff being worked on.

- Close button needs to be more prominent and probably in its own column
- StarCraft path page feels incredibly cramped and should probably take up the full space (not half)
- Most other pages could probably be a single column as well (but at the very least this should be chosen on a per-page basis)

2Pacalypse-:
- Move the Close button into its own column and increase its size
- Lower the max-width from 960px -> 840px
- Make all of the setting pages a single column
- Add/remove spacers where it made sense (probably botched this one)

Closes #945

### DIFF
--- a/client/settings/settings-content.tsx
+++ b/client/settings/settings-content.tsx
@@ -1,8 +1,8 @@
 import styled from 'styled-components'
 
-export const FormContainer = styled.div<{ $doubleColumn?: boolean }>`
+export const FormContainer = styled.div`
   display: grid;
-  grid-template-columns: ${props => (props.$doubleColumn ? '1fr 1fr' : '1fr')};
+  grid-template-columns: 1fr;
   column-gap: 40px;
   width: 100%;
 `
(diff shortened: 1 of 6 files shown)


2Pacalypse-:
Few screenshots:

image

image

image

client/settings/settings-content.tsx
```diff
@@ -1,8 +1,13 @@
 import styled from 'styled-components'
 
-export const FormContainer = styled.div`
+export const FormContainer = styled.div<{ $doubleColumn?: boolean }>`
```
tec27, line 3: Do we ever even use this now? Seems like it might be better to just remove this

tec27 (approved):
The slider's feel a bit too spaced out to me now, and I still think the close button needs to be a bit more prominent, but I'll fix those up in a follow-up commit so we don't have to go back and forth a bunch of times

PR #950, opened 2023-06-04
Allow users to change their language in settings.

2Pacalypse-:
Since the language is changed automatically each time user selects a different option, there is a debounced time of 200ms before we actually make the request. I'm not sure if we should display some kind of an overlay that "disables" the settings until the language is fully changed.

### DIFF
--- a/client/auth/action-creators.ts
+++ b/client/auth/action-creators.ts
@@ -5,7 +5,7 @@ import { SbUserId, SelfUser } from '../../common/users/sb-user'
 import { ClientSessionInfo } from '../../common/users/session'
 import type { PromisifiedAction, ReduxAction } from '../action-types'
 import type { ThunkAction } from '../dispatch-registry'
-import { changeLanguageLocally } from '../i18n/action-creators'
+import { maybeChangeLanguageLocally } from '../i18n/action-creators'
 import { abortableThunk, RequestHandlingSpec } from '../network/abortable-thunk'
 import { encodeBodyAsParams, fetchJson } from '../network/fetch'
 import { AccountUpdateSuccess, AuthChangeBegin } from './actions'
@@ -96,9 +96,7 @@ export function logIn(
       payload: result,
     })
 
-    if (result.user.locale && result.user.locale !== locale) {
-      dispatch(changeLanguageLocally(result.user.locale))
-    }
+    dispatch(maybeChangeLanguageLocally(result.user.locale))
   })
 }
 
@@ -139,9 +137,7 @@ export function signUp(
       payload: result,
     })
 
-    if (result.user.locale && result.user.locale !== locale) {
-      dispatch(changeLanguageLocally(result.user.locale))
-    }
+    dispatch(maybeChangeLanguageLocally(result.user.locale))
   })
 }
 
@@ -162,9 +158,7 @@ export function getCurrentSession(
       payload: result,
     })
 
-    if (result.user.locale && result.user.locale !== locale) {
-      dispatch(changeLanguageLocally(result.user.locale))
-    }
+    dispatch(maybeChangeLanguageLocally(result.user.locale))
   })
 }
 
--- a/client/auth/action-creators.ts
+++ b/client/auth/action-creators.ts
@@ -4,34 +4,14 @@ import { apiUrl } from '../../common/urls'
 import { SbUserId, SelfUser } from '../../common/users/sb-user'
 import { ClientSessionInfo } from '../../common/users/session'
 import type { PromisifiedAction, ReduxAction } from '../action-types'
-import { openSimpleDialog } from '../dialogs/action-creators'
-import type { DispatchFunction, ThunkAction } from '../dispatch-registry'
-import i18n from '../i18n/i18next'
-import logger from '../logging/logger'
+import type { ThunkAction } from '../dispatch-registry'
+import { changeLanguageLocally } from '../i18n/action-creators'
 import { abortableThunk, RequestHandlingSpec } from '../network/abortable-thunk'
 import { encodeBodyAsParams, fetchJson } from '../network/fetch'
 import { AccountUpdateSuccess, AuthChangeBegin } from './actions'
 import { getBrowserprint } from './browserprint'
 
 const typedIpc = new TypedIpcRenderer()
-
-async function changeLanguage(locale: string, dispatch: DispatchFunction<ReduxAction>) {
-  try {
-    await i18n.changeLanguage(locale)
-  } catch (error) {
-    dispatch(
-      openSimpleDialog(
-        i18n.t('auth.language.changeErrorHeader', 'Error changing the language'),
-        i18n.t(
-          'auth.language.changeErrorMessage',
-          'Something went wrong when changing the language',
-        ),
-        true,
-      ),
-    )
-    logger.error(`There was an error changing the language: ${(error as any)?.stack ?? error}`)
-  }
-}
 
 type IdRequestable = Extract<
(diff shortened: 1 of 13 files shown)


tec27:
Disabling the UI each time they change a setting would defeat the purpose of debouncing (it would just add 200ms of extra latency on top of each request) so I do not think that is the right approach :)

client/i18n/action-creators.ts
```diff
@@ -0,0 +1,68 @@
...
+        openSimpleDialog(
+          i18n.t('auth.language.changeErrorHeader', 'Error changing the language'),
+          i18n.t(
+            'auth.language.changeErrorMessage',
+            'Something went wrong when changing the language',
+          ),
+          true,
+        ),
+      )
+      logger.error(`There was an error changing the language: ${(error as any)?.stack ?? error}`)
```
tec27, line 27: For future reference, if you're going to use `error` as `any` you can just put that in the catch:
```ts
} catch (error: any) {
```

client/i18n/action-creators.ts
```diff
@@ -0,0 +1,68 @@
...
+  return abortableThunk(spec, async (dispatch, getState) => {
+    try {
+      const {
+        auth: {
+          user: { id },
+        },
+      } = getState()
+      const result = await fetchJson<ChangeLanguagesResponse>(apiUrl`users/${id}/language`, {
+        method: 'POST',
+        body: JSON.stringify({ language }),
```
tec27, line 45: You should always use `encodeBodyAsParams` for things that can be encoded that way (and the TS typings will prevent using it on things that cannot). Using JSON means the request has to be pre-flighted which can add a lot of latency for people that are far from the server

client/material/radio.tsx
```diff
@@ -150,23 +160,25 @@ export const RadioButton = React.memo(
...
+    name,
     selected,
     onChange,
   }: RadioButtonProps<T>) => {
     const id = useId()
 
     const [buttonProps, rippleRef] = useButtonState({ disabled })
 
-    const handleChange = useStableCallback(() => {
+    const handleChange = useStableCallback((event: React.ChangeEvent<HTMLInputElement>) => {
```
tec27, line 171: If you're going to pass the event through there's no particular reason for this to exist I don't think. In the props you pass to the input just do:
```ts
onChange: !disabled ? onChange : undefined
```

client/settings/settings.tsx
```diff
@@ -173,12 +181,17 @@ function Settings({
       <NavContainer>
         {IS_ELECTRON ? (
           <>
-            <NavSectionTitle>App settings</NavSectionTitle>
+            <NavSectionTitle>{t('settings.user.label', 'User')}</NavSectionTitle>
```
tec27, line 184: Thank you for matching what I have in the gql branch :)

client/settings/settings.tsx
```diff
@@ -378,6 +391,8 @@ function SettingsSubPageDisplay({
   }
```
tec27, line 392: I can't comment on it but you need to remove/change the Electron check above this, this settings page should be available on the web too. If you look at my version in the `gql` branch I have a version of this function that should work + deal with assertUnreachable properly

client/settings/user/language-settings.tsx
```diff
@@ -0,0 +1,75 @@
...
+import SubmitOnEnter from '../../forms/submit-on-enter'
+import { changeUserLanguage } from '../../i18n/action-creators'
+import { RadioButton, RadioGroup } from '../../material/radio'
+import { useAppDispatch } from '../../redux-hooks'
+import { useStableCallback } from '../../state-hooks'
+import { colorTextSecondary } from '../../styles/colors'
+import { overline } from '../../styles/typography'
+import { FormContainer } from '../settings-content'
+
+const RadioOverline = styled.div`
```
tec27, line 20: Wonder if this should just be part of the `RadioGroup` component? Seems like we'd generally always want to label these things

client/material/radio.tsx
tec27: Actually not even sure you need to be attaching the change handlers there at all though? May as well just attach it on the RadioGroup element and handle all the change events when they bubble there (the browser should properly prevent change events from happening on disabled inputs anyway)

server/lib/users/user-api.ts
```diff
@@ -139,6 +142,12 @@ const relationshipsThrottle = createThrottle('accountrelationships', {
   window: 60000,
 })
 
+const changeLanguageThrottle = createThrottle('changelanguage', {
```
tec27, line 145: I think it is more expensive to check this throttle than it is to change a single DB column so I don't know that this is really worth it

tec27 (changes requested):

client/auth/action-creators.ts
```diff
@@ -117,7 +97,7 @@ export function logIn(
     })
 
     if (result.user.locale && result.user.locale !== locale) {
```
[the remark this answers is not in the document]
2Pacalypse-: Hmm, not sure I understand. Do you mean a part of the `@auth/loadCurrentSession` action? Because I'm just calling the `i18n`'s `changeLanguage` here (which will internally try to best match the user's locale to one of our `TranslationLanguage`, using the fallback language if none is found), there's no reducer changes.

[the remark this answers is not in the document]
tec27: No I mean in `changeLanguageLocally`. You're dispatching it, it's an action creator, whether or not it makes any reducer changes is irrelevant to that.

I want you to not perform this repeated condition around dispatching this in every place you call it, and instead just perform that check in the action creator. If we needed to add a 4th place that calls this, whoever adds that would *also* need to add this same condition (or potentially introduce weird bugs/behaviors in that place that don't happen in the others), which isn't good. It's better to just always call the function and have it no-op itself if nothing needs to be done (this is why I tend to create `maybeDoX` functions)

client/material/radio.tsx
2Pacalypse-: I did the first part of this, but when I do the second part, I get this error:
image

Could probably be avoided using `defaultChecked` (even though I don't see that attribute documented anywhere?), but maybe just passing the `onChange` directly to the `input` is good as well?

tec27: Possibly fine, alternatively you can just do one of:
1) declare a no-op function at the top-level in this file and pass that instead of `undefined`
2) do what it suggests and also set `readOnly` in the disabled case

2Pacalypse-: I couldn't really make the 2. work (I still can't find the documentation for `defaultChecked` anywhere), so I went with the 1. option.

tec27: `defaultChecked` isn't relevant, that's just for uncontrolled components. That's a separate thing they're suggesting from setting `onChange` or `readOnly`.

If you set `defaultChecked`, the browser handles the checked/unchecked value from then on, just like if you set `defaultValue` on a text input. The error is saying that either your component needs to be uncontrolled, *or* you need to have one of: `onChange` or `readOnly` set.

I think solving it the way you did is fine, but just for future reference

client/material/radio.tsx
```diff
@@ -1,25 +1,45 @@
...
+
+  padding: 8px 0;
 `
 
 export interface RadioGroupProps<T> {
   children: Array<ReturnType<typeof RadioButton> | null | undefined>
   value: T
-  onChange?: (value: T) => void
+  name?: string
+  overline?: React.ReactNode
```
tec27, line 30: think this should be called `label` probably

tec27 (approved):

PR #951, opened 2023-06-04
Add the translations strings to a few files.

2Pacalypse-:


client/activities/activity-back-button.tsx
```diff
@@ -22,6 +24,10 @@ export function ActivityBackButton() {
   }, [dispatch])
 
   return shouldShow ? (
-    <BackButton icon={<MaterialIcon icon='arrow_back' />} title='Back' onClick={onBackClick} />
+    <BackButton
+      icon={<MaterialIcon icon='arrow_back' />}
+      title={t('common.literals.back', 'Back')}
```
tec27, line 29: nit: No idea what 'literals' is really meant to contain in this context. Maybe it should be `common.actions`?

client/download/download.tsx
```diff
@@ -81,46 +83,54 @@ const InstallerButtonLabel = styled.span`
...
-          </a>
-        </SupportLinks>
-      </React.Fragment>
-    )
-  }
+export function Download() {
+  const { t } = useTranslation()
 
-  onDownloadClick = () => {
+  const onDownloadClick = useStableCallback(() => {
```
tec27, line 89: This is a memory/runtime cost for 0 gain, just use a normal function

The rule is: If you are creating a component that may be used in a bunch of places and you can't be sure when re-renders will matter, memoize. If you are creating a top-level-ish component and you need to be sure components that you're passing functions to don't re-render if their other props don't change, memoize.

This case is neither, this is a top-level component that is pretty shallow and doesn't care about re-rendering (and, in fact, has no props that can even change so...)

client/download/update-overlay.tsx
```diff
@@ -127,38 +128,48 @@ export function UpdateDialog({
...
+          {t(
+            'clientUpdate.overlay.newUpdateReady',
+            'A new update has been downloaded and is ready to install. ' +
+              'Please restart the application to continue.',
+          )}
         </Subtitle1>
         <RaisedButton
           onClick={() => ipcRenderer.send('updaterQuitAndInstall')}
-          label='Restart now'
+          label={t('clientUpdate.overlay.restartNowLabel', 'Restart now')}
```
tec27, line 160: I don't really understand the reason for putting a `Label` suffix on things in this way, we don't put it on everything so the decision here seems very random.

client/ladder/ladder.tsx
```diff
@@ -620,18 +625,39 @@ export function LadderTable(props: LadderTableProps) {
...
-          <SelectOption value={DivisionFilter.Diamond} text={'Diamond'} />
-          <SelectOption value={DivisionFilter.Platinum} text={'Platinum'} />
-          <SelectOption value={DivisionFilter.Gold} text={'Gold'} />
-          <SelectOption value={DivisionFilter.Silver} text={'Silver'} />
-          <SelectOption value={DivisionFilter.Bronze} text={'Bronze'} />
-          <SelectOption value={DivisionFilter.Unrated} text={'Unrated'} />
+          <SelectOption value={DivisionFilter.All} text={t('ladder.divisionOption.all', 'All')} />
+          <SelectOption
+            value={DivisionFilter.Champion}
+            text={t('ladder.divisionOption.champion', 'Champion')}
```
tec27, line 635: These should use text that is shared among all things that need to refer to these divisions, this is not specific to this dropdown

tec27 (changes requested):

client/activities/activity-back-button.tsx
2Pacalypse-: Yeah, not sure. I already added the `common.literals.X` key during the auth stuff, mostly for some very common strings, but not sure if this organization makes much sense: https://github.com/ShieldBattery/ShieldBattery/blob/e411f812dc80ad535d72ef893ebf8ffd645c758b/server/public/locales/en/global.json#L114-L124

tec27: I do not think it really makes sense, no. There are some things in here that are pretty common (Cancel, Continue, Required, Save), but there are bunch of things that are auth-specific and should be in the auth stuff, not some random common stuff. `literals` has no meaning here, "string literal" in programming just means "a string that is directly in a code, not a variable" so I don't really see what that can mean in this context.

Other nit: "login" is a bad naming because the string is "Log in", it should be "logIn"

2Pacalypse-: Yeah, makes sense. I removed the `common.literals.X` key, and instead used `common.actions.X` for a few of those common actions. For all other stuff I just used the key that was already used for that file.

tec27: Sounds good 👍

client/auth/signup.tsx
```diff
@@ -260,8 +260,8 @@ export function Signup() {
         required(t('auth.passwordValidator.confirm', 'Confirm your password')),
         matchesOther('password', t('auth.passwordValidator.matching', 'Enter a matching password')),
       ),
-      ageConfirmation: requireChecked(t('common.literals.required', 'Required')),
-      policyAgreement: requireChecked(t('common.literals.required', 'Required')),
+      ageConfirmation: requireChecked(t('auth.ageConfirmation.required', 'Required')),
```
tec27, line 263: Would probably be good to use the same string for all `required/requireChecked` validators if possible (similar to min/maxLength)

client/ladder/ladder.tsx
```diff
@@ -620,18 +625,36 @@ export function LadderTable(props: LadderTableProps) {
...
-          <SelectOption value={DivisionFilter.Diamond} text={'Diamond'} />
-          <SelectOption value={DivisionFilter.Platinum} text={'Platinum'} />
-          <SelectOption value={DivisionFilter.Gold} text={'Gold'} />
-          <SelectOption value={DivisionFilter.Silver} text={'Silver'} />
-          <SelectOption value={DivisionFilter.Bronze} text={'Bronze'} />
-          <SelectOption value={DivisionFilter.Unrated} text={'Unrated'} />
+          <SelectOption value={DivisionFilter.All} text={t('ladder.division.all', 'All')} />
+          <SelectOption
+            value={DivisionFilter.Champion}
+            text={t('ladder.division.champion', 'Champion')}
```
tec27, line 635: `ladder.division` and `ladder.divisionLabel` seem *very hard* to differentiate.

I'd probably call this one `ladder.divisionGroup` (and the `Label` suffix on the other one still feels weird/unnecessary to me)

tec27 (approved):

PR #956, opened 2023-06-08
Size egui screen to always be 1080 points tall

### LINKED ISSUE #954: Dynamically size replay/obs UI based on current resolution
The current UI seems quite small at larger screen sizes, would be nice to keep it proportional (but we also need to deal with resolution changes mid-game)

neivv:
(Using higher pixels:points ratio if the render target is larger)

Closes #954

Supposing the changes work properly on higher resolution screens =)


tec27:
Merged as 3c4cd85955c8dc9d1942d14b8b5664c24151457b

PR #958, opened 2023-06-15
Add the translations guide. Written by ExcalZ.

2Pacalypse-:


tec27:
I don't know if you're waiting for a review on this or something but I don't think it needs one

PR #965, opened 2023-06-20
Add translation keys to the matchmaking folder.

2Pacalypse-:


client/matchmaking/find-match-forms.tsx
```diff
@@ -214,7 +216,7 @@ export function MapVetoesControl({
         ))}
       </MapSelections>
       <VetoStatus>
-        <VetoStatusLabel>Vetoes left: </VetoStatusLabel>
+        <VetoStatusLabel>{t('matchmaking.findMatch.vetoesLeft', 'Vetoes left:')} </VetoStatusLabel>
```
tec27, line 219: This should probably be a `Trans` component with both of these parts in one string

tec27 (approved):

PR #968, opened 2023-06-22
Add translation keys to a notifications, starcraft, and whispers

2Pacalypse-:
folders.

### DIFF
--- a/client/starcraft/shieldbattery-health.jsx
+++ b/client/starcraft/shieldbattery-health.jsx
@@ -65,27 +65,31 @@ export class ShieldBatteryHealthDialog extends React.Component {
           </Text>
         ) : (
           <div>
-            <Trans t={this.props.t} i18nKey='starcraft.shieldbatteryHealth.contents'>
-              <Text>
+            <Text>
+              <Trans t={this.props.t} i18nKey='starcraft.shieldbatteryHealth.topContents'>
                 We've detected that the following ShieldBattery files are missing or have been
                 modified:
-              </Text>
-              <FileList>
-                {{ initDescription }}
-                {{ mainDescription }}
-              </FileList>
+              </Trans>
+            </Text>
+            <FileList>
+              {{ initDescription }}
+              {{ mainDescription }}
+            </FileList>
 
-              <Text>
+            <Text>
+              <Trans t={this.props.t} i18nKey='starcraft.shieldbatteryHealth.middleContents'>
                 This is often the result of installed anti-virus software taking action on false
                 positives. You may need to add exceptions for these files, or tell the software to
                 remove them from quarantine. You can also try re-installing ShieldBattery.
-              </Text>
+              </Trans>
+            </Text>
 
-              <Text>
+            <Text>
+              <Trans t={this.props.t} i18nKey='starcraft.shieldbatteryHealth.bottomContents'>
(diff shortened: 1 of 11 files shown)


client/starcraft/shieldbattery-health.jsx
```diff
@@ -57,29 +65,34 @@ export class ShieldBatteryHealthDialog extends React.Component {
...
           <div>
-            <Text>
-              We've detected that the following ShieldBattery files are missing or have been
-              modified:
-            </Text>
-            <FileList>
-              {initDescription}
-              {mainDescription}
-            </FileList>
+            <Trans t={this.props.t} i18nKey='starcraft.shieldbatteryHealth.contents'>
```
tec27, line 68: This is gonna be hard for translators to deal with. You should split this kind of thing up into:
- The text at the top
- The next paragraph
- The last paragraph
(The file list doesn't need to be translated so that can stay as is)

None of these really need to be a `Trans` as a result

tec27 (approved):

client/starcraft/shieldbattery-health.jsx
2Pacalypse-: I've generally preferred using the `<Trans>` component for text that spans multiple lines in JSX. It's just so much more convenient and easier to change than doing the text concatenation. As long as only the text is wrapped with the `<Trans>` component, and not other elements, I don't see the downside of using it like this. Do you agree with this?

tec27: No, I don't agree. There's no "text concatenation" here, you're putting stuff in separate elements (`<Text>` in this case). Save Trans for stuff that needs special formatting but is part of the same content (i.e. a sentence containing a link).

In the previous version of your commit, this generates:
```
<0>We've detected that the following ShieldBattery files are missing or have been modified:</0>
<1>{{initDescription}}{{mainDescription}}</1><2>This is often the result of installed anti-virus
software taking action on false positives. You may need to add exceptions for these files, or tell
the software to remove them from quarantine. You can also try re-installing ShieldBattery.</2>
<3>If you are able to, reporting these as false positives to your anti-virus vendor will help this stop
happening for other users as well!</3>
```

This is not going to be easy for translators to deal with, it's way too easy to put a sentence in the wrong "slot", accidentally move tags around, etc. and then we end up with content that doesn't match the formatting/structure we have in the English version.

I understand this feels convenient for you while you're writing the code, but we *really* need to prioritize making this easy for the people that do not have a good understanding of the code and have way worse tools to edit this, it's going to be 100 times harder to review the work they're doing and make sure it all lines up. Avoiding Trans usage when possible greatly reduces the possible failure states and maximizes the amount of text that can be machine translated.

tec27: If you're just asking about whether it's okay to use Trans for the *new* content (split into one string for each element), that's fine provided that it doesn't generate unnecessary tags in the generate string. My main point is that we need to keep the generated strings as simple as possible

2Pacalypse-: Yes sorry, I should've been more clear. I was asking is it ok to use `<Trans>` component the way I did it in the "Address PR comment." commit. I agree with keeping the generated strings simple and not having unnecessary tags in the generated string.

PR #971, opened 2023-06-25
Add translation strings to the lobbies folder.

2Pacalypse-:


client/lobbies/lobby.tsx
```diff
@@ -154,7 +155,7 @@ function renderChatMessage(msg: SbMessage) {
   }
 }
 
-interface LobbyProps {
+interface LobbyProps extends WithTranslation {
```
tec27, line 158: I would not suggest doing this, as these props are more intended to be the externally visible API of a component (e.g. you could use them to typecheck props you intend to forward to a `Lobby`). You should instead add it in the generic, e.g. `class Lobby extends React.Component<LobbyProps & WithTranslation>`

tec27 (approved):

PR #972, opened 2023-06-25
Add translations to the root folder.

2Pacalypse-:
And also rename "league." key prefix to "leagues." to better match feature name.

### DIFF
--- a/client/root-error-boundary.tsx
+++ b/client/root-error-boundary.tsx
@@ -101,6 +101,12 @@ interface ContentsErrorBoundaryState {
   hasTranslationError?: boolean
 }
 
+/**
+ * A React error boundary that is specifically meant to catch the errors in calling
+ * translation-related functions when displaying the error contents. If there are any errors in
+ * displaying the translated version of the error contents, we fallback to displaying the static,
+ * English version of the error contents.
+ */
 class ContentsErrorBoundary extends React.Component<
   ContentsErrorBoundaryProps,
   ContentsErrorBoundaryState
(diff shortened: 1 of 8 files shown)


client/root-error-boundary.tsx
```diff
@@ -9,7 +10,7 @@ import ResetStyle from './styles/reset'
 import { Headline5, subtitle1 } from './styles/typography'
 import { WindowControls, WindowControlsStyle } from './system-bar/window-controls'
 
-export interface RootErrorBoundaryProps {
+export interface RootErrorBoundaryProps extends WithTranslation {
```
tec27, line 13: same thing as the other PR, you should put `WithTranslation` in the generic instead of adding it to this interface

client/root-error-boundary.tsx
```diff
@@ -82,14 +84,16 @@ export class RootErrorBoundary extends React.Component<
           <WindowControlsStyle />
           <WindowControls />
           <Container>
-            <Headline5>Something went wrong :(</Headline5>
+            <Headline5>{t('rootErrorBoundary.title', 'Something went wrong :(')}</Headline5>
```
tec27, line 87: Given that we know an exception has already occurred and we don't know what state things are in, I think you probably want to try and be *very safe* with making calls into i18next here. Calls should probably be wrapped in a try/catch and fallback to the English values on failure

client/root-error-boundary.tsx
```diff
@@ -82,14 +84,16 @@ export class RootErrorBoundary extends React.Component<
...
-              .
+              <Trans t={t} i18nKey='rootErrorBoundary.contents'>
+                Please report this issue to us in our{' '}
+                <a href={DISCORD_URL} title='Discord' target='_blank' rel='noopener'>
+                  Discord
+                </a>
+                .
+              </Trans>
             </Instructions>
             <RaisedButton label='Reload app' color='primary' onClick={this.reloadApp} />
```
tec27, line 98: You missed a string

tec27 (approved):

PR #977, opened 2023-06-29
Add the rest of the translation strings that were previously missed.

2Pacalypse-:


client/auth/signup.tsx
```diff
@@ -316,22 +316,24 @@ export function Signup() {
...
-                  />
-                  , and{' '}
-                  <DialogLink
-                    dialogType={DialogType.PrivacyPolicy}
-                    text={t('auth.signup.privacyLink', 'Privacy')}
-                  />{' '}
-                  policies
+                  <Trans t={t} i18nKey='auth.signup.readAndAgree'>
+                    I have read and agree to the{' '}
+                    <DialogLink
```
tec27, line 321: This is really all the same string, so you probably need to change `DialogLink` to take this string as a child instead

client/matchmaking/matchmaking-disabled-card.tsx
```diff
@@ -112,7 +112,7 @@ export function ConnectedMatchmakingDisabledCard({
           <Headline6>{dateFormat.format(nextStartDate)}</Headline6>
           {nextEndDate && nextEndDate > nextStartDate ? (
             <>
-              <ToText>to</ToText>
+              <ToText>{t('matchmaking.disabledCard.startToEnd', 'to')}</ToText>
```
tec27, line 115: generally I think this should be a `Trans` containing both dates and the bridge word, translating the bridge word on its own both doesn't give context and may not be a general enough solution to cover how date ranges are localized. Would also accept just replacing this with an en dash so it doesn't need to be translated

tec27 (approved):

PR #987, opened 2023-07-04
Fix the issue where the channels can potentially end up in wrong order.

2Pacalypse-:
The previous code could produce a bug in a following case:
- connect with one client to the server and create a channel
- connect with the second client to the server
- second client will receive both the `chatReady` and `init3` events in the non-consistent order (depends on how things execute on the server)
- if the second client receives the `init3` event before the `chatReady` event, their channel list will be out of order

This fix makes it so we don't send the `init3` event when they connect to the server, but only immediatelly after they join a channel.

Additionally, we could also change the reducer on the frontend to always rewrite the channel list when it receives the `chatReady` event, but it seems unnecessary to me.

### DIFF
--- a/server/lib/chat/chat-service.ts
+++ b/server/lib/chat/chat-service.ts
@@ -145,24 +145,20 @@ export default class ChatService {
       userSockets.subscribe(getChannelPath(channelId))
       userSockets.subscribe(getChannelUserPath(channelId, userSockets.userId))
 
-      try {
-        const [channelInfo, userChannelEntry] = await Promise.all([
-          getChannelInfo(channelId),
-          getUserChannelEntryForUser(userSockets.userId, channelId),
-        ])
+      const [channelInfo, userChannelEntry] = await Promise.all([
+        getChannelInfo(channelId),
+        getUserChannelEntryForUser(userSockets.userId, channelId),
+      ])
 
-        if (channelInfo && userChannelEntry) {
-          this.publisher.publish(getChannelPath(channelId), {
-            action: 'init3',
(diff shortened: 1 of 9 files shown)


server/lib/chat/chat-service.ts
```diff
@@ -143,7 +142,28 @@ export default class ChatService {
...
+      userSockets.subscribe(getChannelUserPath(channelId, userSockets.userId))
+
+      try {
+        const [channelInfo, userChannelEntry] = await Promise.all([
+          getChannelInfo(channelId),
+          getUserChannelEntryForUser(userSockets.userId, channelId),
+        ])
+
+        if (channelInfo && userChannelEntry) {
+          this.publisher.publish(getChannelPath(channelId), {
```
tec27, line 155: You are now publishing this event to *every connected user* in the channel

server/lib/chat/chat-service.ts
```diff
@@ -115,7 +114,7 @@ export default class ChatService {
       .on('userQuit', userId => this.handleUserQuit(userId))
   }
 
-  private updateUserAfterJoining(
+  private async updateUserAfterJoining(
```
tec27, line 117: If you change this method to async, you need to be changing every single place that calls this. If your intention is that this method could never fail, you'd be better off not making this async but instead wrapping the contents with `Promise.resolve().then(async () => { /* ... */ }).catch(err => { /* do whatever error handling */ })`

In this case though, I think it's a pretty tough argument to make that the error handling should always be generic, it seems like the things calling this should decide how they want to handle failures

tec27 (changes requested):

PR #999, opened 2023-08-02
Add admin panel to upload, edit, and view channel content.

2Pacalypse-:
For now, the only channel content that we support is channel banners (bigger images that channel owners can use to style their channels). Once we have everything working for the channel banners, the next thing would be to support channel badges (smaller images that serve as channel logos) as well.


common/images.ts
```diff
@@ -0,0 +1,47 @@
...
+export async function resizeImage(
+  filePath: string,
+  width: number,
+  height: number,
+): Promise<[image: sharp.Sharp, imageExtension: string]> {
+  const image = sharp(filePath)
+  const metadata = await image.metadata()
+
+  let imageExtension: string
+  if (metadata.format !== 'jpg' && metadata.format !== 'jpeg' && metadata.format !== 'png') {
```
tec27, line 34: I do not think this logic is applicable for every case (and for channel banners, I would say jpg is almost certainly the preferred option). You should make this configurable in some way (e.g. provide a list of allowed types + a fallback type to use if it isn't one of those)

common/images.ts
```diff
@@ -0,0 +1,47 @@
...
+    image.toFormat('png')
+    imageExtension = 'png'
+  } else {
+    imageExtension = metadata.format
+  }
+
+  image.resize(width, height, {
+    fit: sharp.fit.cover,
```
tec27, line 42: This should probably also be configurable

server/lib/chat/channel-banner-models.ts
```diff
@@ -0,0 +1,134 @@
+import sql from 'sql-template-strings'
+import { assertUnreachable } from '../../../common/assert-unreachable'
+import { ChannelBanner, ChannelBannerId } from '../../../common/chat-channels/channel-banners'
+import db, { DbClient } from '../db'
+import { Dbify } from '../db/types'
+import { getUrl } from '../file-upload'
+import { Patch } from '../leagues/league-models'
```
tec27, line 7: If you're going to start using this then please complete the TODO and move it, it's not okay for this file to depend on Leagues

server/lib/chat/chat-api.ts
```diff
@@ -439,4 +463,182 @@ export class AdminChatApi {
...
+    if (!channelBanner) {
+      throw new ChatServiceError(
+        ChatServiceErrorCode.ChannelBannerNotFound,
+        'channel banner not found',
+      )
+    }
+
+    const channelIds = new Set<SbChannelId>()
+    for (let j = 0; j < channelBanner.availableIn.length; j++) {
+      channelIds.add(channelBanner.availableIn[j])
```
tec27, line 489: just do `new Set(channelBanner.availableIn)`?

server/lib/chat/chat-api.ts
```diff
@@ -439,4 +463,182 @@ export class AdminChatApi {
...
+    }
+    const channelInfos = await getChannelInfos(Array.from(channelIds))
```
tec27, line 491: Honestly pretty confused why you need this `Set` stuff at all though, we're pulling this list out of the DB (where it should have already been deduped), so we're deduping a list that we can be pretty sure contains no dupes. Just pass the array you got to this function instead of doing this unnecessary conversion through a Set

server/lib/chat/chat-api.ts
```diff
@@ -439,4 +463,182 @@ export class AdminChatApi {
...
+    }
+    const [image, imageExtension] = await resizeImage(
+      imageFile.filepath,
+      CHANNEL_BANNER_WIDTH,
+      CHANNEL_BANNER_HEIGHT,
+    )
+
+    return await transact(async client => {
+      const channelInfos =
+        availableIn && availableIn.length > 0 ? await findChannelsByName(availableIn) : []
```
tec27, line 545: should probably dedupe these (unless `findChannelsByName` does that, but I don't think it does)

server/lib/chat/chat-api.ts
```diff
@@ -439,4 +463,182 @@ export class AdminChatApi {
...
+        updatedChannelBanner.imagePath = imagePath
+      }
+
+      const channelBanner = await adminUpdateChannelBanner(bannerId, updatedChannelBanner, client)
+
+      if (image && imagePath) {
+        const buffer = await image.toBuffer()
+        await writeFile(imagePath, buffer, {
+          acl: 'public-read',
+          type: imageExtension === 'png' ? 'image/png' : 'image/jpeg',
```
tec27, line 635: I don't think this logic is particularly great here given that you've pulled the logic for choosing conversion types away from all of this, a safer thing to do would be to use some library to choose a mimetype based on extension (or figure out if some part of this process already does that, or pull this behavior into the `writeFile` stuff with an optional override)

server/migrations/20230731144843-create-channel-banners.js
```diff
@@ -0,0 +1,25 @@
+exports.up = async function (db) {
+  await db.runSql(`
```
tec27, line 2: You're sorting by `uploaded_at` so you should probably have an index to support that

tec27: You should also consider how you're going to sort these for users (name? I don't know if uploaded_at really makes sense there) and index to support that as well

server/lib/chat/chat-api.ts
```diff
@@ -439,4 +464,195 @@ export class AdminChatApi {
...
+      channelBanner: toChannelBannerJson(channelBanner),
+      channelInfos: channelInfos.map(c => toBasicChannelInfo(c)),
+    }
+  }
+
+  @httpGet('/banners')
+  @httpBefore(checkAllPermissions('manageChannelContent'))
+  async getChannelBanners(ctx: RouterContext): Promise<AdminGetChannelBannersResponse> {
+    const channelBanners = await adminGetChannelBanners()
+    const channelIds = new Set<SbChannelId>()
```
tec27, line 500: You can write this more simply as:
```ts
const channelIds = new Set(channelBanners.flatMap(c => c.availableIn))
```

server/lib/http/patch-type.ts
tec27, line 1: Don't really see what at all this has to do with HTTP, and I don't agree that it is "mostly useful in PATCH API requests". Put it in `common/patch.ts` and fix the comment to be:
```
An object that specifies desired changes to another existing object. All keys become optional, and any fields that were originally optional become nullable. `undefined` implies that no changes will be made to the field, `null` implies that the field should be deleted.
```

tec27 (approved):

PR #1001, opened 2023-08-07
Make the maximum veto count dynamic for each map pool.

2Pacalypse-:


client/matchmaking/find-1v1.tsx
```diff
@@ -98,24 +98,31 @@ const Form1v1 = React.forwardRef<FindMatchFormRef, Form1v1Props>(
...
+              <SectionTitle>{t('matchmaking.findMatch.mapPool', 'Map pool')}</SectionTitle>
+              {mapPoolOutdated ? (
+                <OutdatedIndicator>
+                  {t('matchmaking.findMatch.updated', 'Updated')}
+                </OutdatedIndicator>
+              ) : null}
+            </MapSelectionsHeader>
+
+            <DescriptionText>
+              {t('matchmaking.findMatch.veto1v1', {
```
tec27, line 113: Like I said yesterday, this message should be chosen based on whether or not the veto count allows map overlaps. If the veto count is high enough, this would be a lie, so we can't just pick this message always. Make a component that takes the veto count, map pool size, and number of players and can pick between the 2 messages.

client/matchmaking/find-2v2.tsx
```diff
@@ -59,25 +59,30 @@ const Form2v2 = React.forwardRef<FindMatchFormRef, Form2v2Props>(
...
+                significantly less often than other maps.
+              </Trans>
+            </DescriptionText>
+            <MapVetoesControl
+              {...bindCustom('mapSelections')}
+              mapPool={mapPool}
+              disabled={disabled}
+            />
+          </>
+        ) : null}
```
tec27, line 85: You probably want a loading indicator, not just... nothing

server/lib/matchmaking/matchmaking-api.ts
```diff
@@ -106,6 +108,20 @@ export class MatchmakingApi {
...
     const { clientId, preferences, identifiers } = body
 
+    const currentMapPool = await getCurrentMapPool(preferences.matchmakingType)
+    if (!currentMapPool) {
+      throw new MatchmakingServiceError(
+        MatchmakingServiceErrorCode.InvalidMapPool,
+        "Map pool doesn't exist",
+      )
+    }
+    if (preferences.mapSelections.length > currentMapPool.maxVetoCount) {
```
tec27, line 118: This error isn't a good addition. If the player previously had vetoes set and the next map pool decreases the count, they can end up being over the count, and now you're just stopping them from matchmaking. Just truncate their vetoes to the proper size.

Alongside that, you should first remove any vetoes that are not maps in the current pool, *before* doing that truncation.

server/lib/matchmaking/matchmaking-preferences-api.ts
```diff
@@ -39,6 +39,10 @@ export class MatchmakingPreferencesApi {
       throw new httpErrors.BadRequest('invalid matchmaking type')
     }
 
+    if (body.mapSelections && body.mapSelections.length > currentMapPool.maxVetoCount) {
```
tec27, line 42: This error is also bad here. These things made sense when the veto count was static but since it can change between map pools and matchmaking preferences are saved across pools, this is a problem

server/lib/models/matchmaking-map-pools.ts
```diff
@@ -8,6 +8,7 @@ interface DbMapPool {
   matchmaking_type: MatchmakingType
   start_date: Date
   maps: string[]
+  max_veto_count: number
```
tec27, line 11: instead of just adding another field here can you please move these types to the standard pattern used everywhere else in the app?

tec27 (changes requested):

server/lib/matchmaking/matchmaking-api.ts
```diff
@@ -115,20 +114,21 @@ export class MatchmakingApi {
...
-    }
 
     await this.userIdManager.upsert(ctx.session!.userId, identifiers)
 
     if (await this.userIdManager.banUserIfNeeded(ctx.session!.userId)) {
       throw new httpErrors.Unauthorized('This account is banned')
     }
 
-    await this.matchmakingService.find(ctx.session!.userId, clientId, identifiers, preferences)
+    const truncatedMapSelections = preferences.mapSelections
```
tec27, line 124: Feel like this should probably be put somewhere common

tec27 (changes requested):

PR #1004, opened 2023-08-11
Add a ChannelHeader component that is displayed on top of chat channels.

2Pacalypse-:


client/chat/channel-header.tsx
```diff
@@ -0,0 +1,252 @@
...
+import { shadow2dp } from '../material/shadows'
+import { Tooltip, TooltipContent } from '../material/tooltip'
+import { ExternalLink } from '../navigation/external-link'
+import { LoadingDotsArea } from '../progress/dots'
+import { useStableCallback } from '../state-hooks'
+import { background700, colorTextFaint, colorTextSecondary } from '../styles/colors'
+import { caption, headline6, singleLine } from '../styles/typography'
+import { ChannelBadge } from './channel-badge'
+
+const ChannelHeaderRoot = styled.div<{ $hasRightPadding: boolean }>`
```
tec27, line 29: This isn't a good name for a prop, it should correspond to *why* this value differs, not *what* the value changes

client/chat/channel-header.tsx
```diff
@@ -0,0 +1,252 @@
...
+  gap: 8px;
+`
+
+const UserCountContainer = styled.div`
+  display: flex;
+  align-items: center;
+  gap: 4px;
+`
+
+const UserCountIcon = styled(MaterialIcon).attrs({ icon: 'groups' })`
```
tec27, line 104: In my new mocks I have this at `20px`

client/chat/channel-header.tsx
```diff
@@ -0,0 +1,252 @@
...
+        text={t('chat.channelHeader.actionItems.leaveChannel', 'Leave channel')}
+        onClick={onLeaveChannelClick}
+      />,
+    )
+  }
+
+  return (
+    <ChannelHeaderRoot $hasRightPadding={actions.length === 0}>
+      {!basicChannelInfo ? (
+        <LoadingDotsArea />
```
tec27, line 200: Think it would be a lot better to create a skeleton layout for this, dots are probably gonna be pretty jarring here

client/chat/channel-header.tsx
```diff
@@ -0,0 +1,252 @@
...
+                  ContentComponent={StyledTooltipContent}>
+                  <ChannelTopic ref={channelTopicRef}>{parsedChannelTopic}</ChannelTopic>
+                </StyledTooltip>
+              ) : null}
+            </NameAndTopicContainer>
+          </BadgeAndTextContainer>
+
+          <ActionsArea>
+            <UserCountContainer>
+              <UserCountIcon icon='groups' size={20} />
```
tec27, line 236: You already specified the icon in the style declaration above, so really you're just using the style declaration to declare a color now, so why is it named `UserCountIcon`? Why does it exist at all? You can just use `MaterialIcon` here, `UserCountContainer` can provide a color for it and the text (which you can also just use `Caption` for)

tec27 (approved):

client/chat/channel-header.tsx
```diff
@@ -0,0 +1,252 @@
...
+        key='leave-channel'
+        text={t('chat.channelHeader.actionItems.leaveChannel', 'Leave channel')}
+        onClick={onLeaveChannelClick}
+      />,
+    )
+  }
+
+  return (
+    <ChannelHeaderRoot $hasRightPadding={actions.length === 0}>
+      {!basicChannelInfo ? (
```
tec27, line 199: basicChannelInfo is a required prop so I'm not sure why you're doing this check in the first place tbh

PR #1008, opened 2023-08-19
Fix the channel banner card trying to display a name of the deleted

2Pacalypse-:
channel.

When the last person leaves a non-official channel, it gets deleted. However, there might've been some channel banners which were available in this now-deleted channel, so we can't display its name in the channel banner card.

Note that this still leaves the deleted channel in the banners table in DB, so an ideal fix would be where each `channelId` in the `available_in` column in the `channel_banners` table is a foreign key to the `channels` table. Is that even possible? :/
Although this also might leave things in a weird state, where a channel banner is marked as `limited`, but doesn't have any channel that it's available in.

### DIFF
--- a/client/chat/admin/channel-content.tsx
+++ b/client/chat/admin/channel-content.tsx
@@ -216,7 +216,7 @@ function ChannelBannerCard({
             <ChannelList>
               {banner.availableIn.map(channelId => (
                 <ChannelListEntry key={channelId}>
-                  {channelInfos.get(channelId)!.name}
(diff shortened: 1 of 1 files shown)


tec27:
> Note that this still leaves the deleted channel in the banners table in DB, so an ideal fix would be where each channelId in the available_in column in the channel_banners table is a foreign key to the channels table. Is that even possible? :/

No, you should probably move to a table that has a row per banner ID <-> channel ID pairing that can be properly FK'd. I don't really think a banner being limited but not having any channels its available in is a particularly weird/incorrect state.

tec27 (approved):

PR #1009, opened 2023-08-20
fix(landing): sign up and login buttons width

MikalaiLappo:
Landing buttons have fixed `width` which may squeeze inner text if it's too long (that happens with different locales)
[Current RU locale "sign up" button being broken](https://github-production-user-asset-6210df.s3.amazonaws.com/102466089/261858336-bd0e36fc-ad33-4fd7-a5fa-8ea426b3654e.png)

A possible solution could be changing `width` to `min-width` but in that case a bigger button may cause extra skewing from its side

This PR solution: 
1. Makes buttons responsive relative to its text (preserving `min-width`)
2. Ensures that a button will be positioned within its space only
3. Wraps buttons to new lines on 800px `media` corresponding to `BenefitContainer` wrap

### DIFF
--- a/client/landing/splash.tsx
+++ b/client/landing/splash.tsx
@@ -125,10 +125,21 @@ const ButtonsContainer = styled.div`
   flex-direction: row;
   flex-wrap: wrap;
   justify-content: center;
+  width: 100%;
+`
+
+const ContainerButtonAligner = styled.div.attrs(p => p)<{ align: 'left' | 'right' | 'center' }>`
+  flex: 1 1 0px;
+  display: flex;
+  justify-content: ${p => ({ center: 'center', left: 'flex-start', right: 'flex-end' })[p.align]};
+  @media (max-width: 800px) {
+    flex: 1 1 9999px;
+    justify-content: center;
+  }
 `
 
 const SplashButton = styled(RaisedButton)`
-  width: 200px;
+  min-width: 200px;
   height: 54px;
   margin: 16px 32px;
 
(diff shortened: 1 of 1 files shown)


client/landing/splash.tsx
```diff
@@ -125,10 +125,21 @@ const ButtonsContainer = styled.div`
   flex-direction: row;
   flex-wrap: wrap;
   justify-content: center;
+  width: 100%;
+`
+
+const ContainerButtonAligner = styled.div.attrs(p => p)<{ align: 'left' | 'right' | 'center' }>`
```
2Pacalypse-, line 131: I don't think you need the `.attrs(p => p)` here.

client/landing/splash.tsx
```diff
@@ -125,10 +125,21 @@ const ButtonsContainer = styled.div`
...
+  flex: 1 1 0px;
+  display: flex;
+  justify-content: ${p => ({ center: 'center', left: 'flex-start', right: 'flex-end' })[p.align]};
+  @media (max-width: 800px) {
```
2Pacalypse-, line 135: What's the difference between this and `@media screen and (max-width: 800px)` that we use in the rest of the file?

Also, please add a blank space before this line.

2Pacalypse- (changes requested):
I'm not sure if all of this is necessary in this particular case. I would imagine that it looks okay-ish with just removing the fixed width constraint on these buttons.

Can you please post screenshots on how it looks with the min-width approach instead, and also how it looks with your approach (please post full page screenshots)?

Also, this might be broken in other places (and other languages), so we would probably need to generalize the whatever solution we decide to use. The login page in particular I think suffers from this problem as well.

client/landing/splash.tsx
MikalaiLappo: Yeah that's just redundant. Haven't used `styled-components` for a while

client/landing/splash.tsx
MikalaiLappo: 1. It was related to the proposed solution specifically — nvm
2. Atm how buttons wrap is controlled by their flex container so they may get a bit too stacked horizontally at some point ([like that](https://github.com/ShieldBattery/ShieldBattery/assets/102466089/ac58d7fa-c338-4807-a417-8468ba97ac2a)) — not really a problem

MikalaiLappo:
I didn't want to PR only a single loc and was looking for some flex/grid workaround that would align the buttons perfectly but couldn't find yet a solution generic and idiomatic enough

I guess someone better familiar with the project should commit `width` to `min-width` fix which should be enough

This PR may be closed. I'll make another or open an Issue if anything

2Pacalypse-:
A single LOC PRs are perfectly fine FWIW :d

PR #1034, opened 2023-09-02
Add the dialog to change channel settings.

2Pacalypse-:
For now, the only editable channel settings are:
- description
- topic
- banner

NOTE: This currently only allows channel owners to edit their own
channels. This means that official channels can't be edited since they
don't have owners. In the next PR I'll allow server admins to edit any
channel settings.


client/chat/channel-banner-reducer.ts
```diff
@@ -0,0 +1,56 @@
+import { Immutable } from 'immer'
+import { ChannelBannerId, ChannelBannerJson } from '../../common/chat-channels/channel-banners'
+import { immerKeyedReducer } from '../reducers/keyed-reducer'
+
+export interface ChannelBannerState {
+  /** An array of channel banners that are available to us. */
+  availableChannelBanners: ChannelBannerId[]
```
tec27, line 7: I think this stuff is likely to change now anyway, but wasn't the whole idea that banners would be available to a *channel* not a *user*? Thus you would need to key this list per channel

client/chat/channel-banner-reducer.ts
```diff
@@ -0,0 +1,56 @@
...
+  /** An array of default channel banners. */
+  defaultChannelBanners: ChannelBannerId[]
+  /** A map of channel banner ID -> channel banner info (used in channel info card, etc.) */
```
tec27, line 10: nit: Comments that just repeat the type info are probably not that useful, we only used to make these comments because we didn't have types

client/chat/channel-banner-reducer.ts
```diff
@@ -0,0 +1,56 @@
...
+  idToInfo: Map<ChannelBannerId, ChannelBannerJson>
+}
+
+const DEFAULT_CHANNEL_BANNER_STATE: Immutable<ChannelBannerState> = {
```
tec27, line 14: nit: prefer ReadonlyDeep to Immutable

client/chat/channel-info-card.tsx
```diff
@@ -166,6 +171,12 @@ export function ConnectedChannelInfoCard({
     dispatch(getBatchChannelInfo(channelId))
   }, [dispatch, channelId])
 
+  useEffect(() => {
+    if (bannerId) {
+      dispatch(getBatchChannelBanner(bannerId))
```
tec27, line 176: Can we not assume that `getBatchChannelInfo` already has this info? Seems like we're just making 2 requests for the same thing here

client/chat/channel-settings-banner.tsx
tec27, line 1: This filename seems weirdly ordered, I think `channel-banner-settings` would make a lot more sense (apply to component name as well)

client/chat/channel-settings-banner.tsx
```diff
@@ -0,0 +1,302 @@
...
+  bannerId?: ChannelBannerId
+}
+
+export function ChannelSettingsBanner({
+  channelId,
+  bannerId: originalBannerId,
+}: ChannelSettingsBannerProps) {
+  const { t } = useTranslation()
+  const dispatch = useAppDispatch()
+  const availableBanners = useAppSelector(s => s.channelBanners.availableChannelBanners)
```
tec27, line 72: I really do wonder whether we need to be storing these in a reducer at all, this info doesn't seem useful outside of this UI so caching it in a reducer feels unnecessary, and it complicates things significantly (since banners are particular to where you are trying to use them). Caching banner info makes sense, the rest seems like something you'd just retrieve when you need and use in the component locally

client/chat/channel-settings-general.tsx
tec27, line 1: Same as the other one, `channel-general-settings` or `general-channel-settings`

tec27: Or just make a folder for the settings stuff and call this `general.tsx`

server/lib/chat/chat-api.ts
```diff
@@ -184,6 +199,28 @@ export class ChatApi {
...
+    const channelId = getValidatedChannelId(ctx)
+    const { body } = validateRequest(ctx, {
+      body: Joi.object<EditChannelRequest>({
+        description: Joi.string().allow(null),
+        topic: Joi.string().allow(null),
+        bannerId: Joi.string().uuid(),
+      }),
+    })
+
+    if (body.description === 'null') {
```
tec27, line 214: We probably need to do something better about this. I was fine with this approach with leagues because it was admin-only but this is not, and thus we're blocking people from setting the string 'null' as a topic or description, which seems pretty dumb

tec27 (changes requested):

2Pacalypse-:
Rebased onto the #1047 so shouldn't be merged until that one is merged.

Also removed the ability to change banners from this PR since that's gonna work completely different now. Will implement new system for changing banners in a separate PR.

client/chat/settings/general.tsx
```diff
@@ -0,0 +1,127 @@
...
+
+export function ChannelSettingsGeneral({
+  channelId,
+  channelDescription: originalChannelDescription,
+  channelTopic: originalChannelTopic,
+}: ChannelSettingsGeneralProps) {
+  const { t } = useTranslation()
+  const dispatch = useAppDispatch()
+  const [error, setError] = useState<Error>()
+  const [isUpdatingDescription, setIsUpdatingDescription] = useState(false)
```
tec27, line 44: feels like you should use an enum for these instead of 2 booleans per thing

server/lib/validation/joi-validator.ts
```diff
@@ -58,7 +58,13 @@ export type ValidatedRequestData<T extends JoiValidationDescriptor> = {
   body: ValidatedBodyType<T>
 }
 
-/** Validates a Koa request using Joi, returning typed values that have been normalized by Joi. */
+/**
+ * Validates a Koa request using Joi, returning typed values that have been normalized by Joi.
+ *
+ * POST and PATCH requests are using a convention where if they recieve a `jsonBody` field in a
```
tec27, line 64: Really not a fan of this, the right way to handle this is via custom Joi extensions, not by making a magic key name.

https://github.com/hapijs/joi/issues/2037#:~:text=Array%20and%20object%20string%20coercion has an example of an extension to object to do this, but it should probably be a separate type (you'll have to look into Joi stuff, I don't know how any of this works)

server/migrations/20230826123551-add-description-and-banner-to-channels.js
```diff
@@ -0,0 +1,19 @@
+exports.up = async function (db) {
+  await db.runSql(`
+    ALTER TABLE channels
+    ADD COLUMN description text,
+    ADD COLUMN banner_id uuid;
```
tec27, line 5: This shouldn't have a banner ID any more, yeah?

tec27 (changes requested):

server/migrations/20230826123551-add-description-and-banner-to-channels.js
2Pacalypse-: I actually left it in this migration and removed it in the following migration because the order in which my migrations were executed locally was fucked. But anyway, I've managed to unfuck them so this is not needed anymore.

server/lib/json/json-validator.ts
tec27, line 1: I really don't think this needs its own package, just put it in `validation`?

server/lib/json/json-validator.ts
```diff
@@ -0,0 +1,24 @@
+import Joi from 'joi'
+
+export const json = Joi.extend({
+  type: 'object',
+  base: Joi.object(),
+  messages: {
+    'json.invalid': 'The field {{#label}} is not a valid JSON',
```
tec27, line 7: "The field {{#label}} is not valid JSON"

server/lib/json/json-validator.ts
```diff
@@ -0,0 +1,24 @@
...
+  },
+  coerce: {
+    from: 'string',
+    method: (value, helpers) => {
+      try {
+        const parsedValue = JSON.parse(value)
+        return {
+          value: parsedValue,
```
tec27, line 15: nit: you can just inline this

server/lib/chat/chat-service.ts
```diff
@@ -381,6 +385,42 @@ export default class ChatService {
...
+    }
+
+    if (originalChannel.ownerId !== userId) {
+      throw new ChatServiceError(
+        ChatServiceErrorCode.CannotModerateChannelModerator,
+        'Only channel owner can edit the channel',
+      )
+    }
+
+    const updatedChannel: Patch<EditChannelRequest> = { ...updates }
```
tec27, line 406: This does not seem like it's doing anything? Why do we need a new object here?

server/lib/chat/chat-models.ts
```diff
@@ -199,6 +204,58 @@ export async function createChannel(
...
+  withClient?: DbClient,
+): Promise<FullChannelInfo> {
+  const { client, done } = await db(withClient)
+  try {
+    let query = sql`
+      UPDATE channels
+      SET
+    `
+
+    const updateEntries = Object.entries(updates).filter(([_, value]) => value !== undefined)
```
tec27, line 223: Pull this above grabbing the db client so things fail faster without tying up the DB pool

server/lib/chat/chat-models.ts
```diff
@@ -199,6 +204,58 @@ export async function createChannel(
...
+ * make sure to only call this method when you update one of the channel's fields.
+ */
+export async function updateChannel(
+  channelId: SbChannelId,
+  updates: Patch<EditChannelRequest>,
+  withClient?: DbClient,
+): Promise<FullChannelInfo> {
+  const { client, done } = await db(withClient)
+  try {
+    let query = sql`
```
tec27, line 218: After you rearrange `updateEntries`, you can construct the whole query in one go instead of appending multiple times

server/lib/chat/chat-api.ts
```diff
@@ -156,6 +168,27 @@ export class ChatApi {
...
+  @httpBefore(
+    throttleMiddleware(editThrottle, ctx => String(ctx.session!.user!.id)),
+    handleMultipartFiles(MAX_IMAGE_SIZE),
+  )
+  async editChannel(ctx: RouterContext): Promise<EditChannelResponse> {
+    const channelId = getValidatedChannelId(ctx)
+    const {
+      body: { jsonBody },
+    } = validateRequest(ctx, {
+      body: Joi.object<{ jsonBody: EditChannelRequest }>({
```
tec27, line 181: `jsonBody` could probably be named better for this request now that it's not a magic string

client/settings/settings.tsx
```diff
@@ -223,6 +223,7 @@ const NavEntryRoot = styled.div<{ $isActive: boolean }>`
   align-items: center;
 
   border-radius: 4px;
+  overflow: hidden;
```
tec27, line 226: Is this really desirable?

client/chat/settings/general.tsx
```diff
@@ -0,0 +1,129 @@
...
+  color: ${colorError};
+  margin-bottom: 8px;
+`
+
+const SuccessIcon = styled(MaterialIcon).attrs({ icon: 'check_circle' })`
+  color: ${colorSuccess};
+`
+
+enum UpdateStatus {
+  Updating = 'Updating',
```
tec27, line 26: zero need for these to be strings, just set no value on the enum. We only use string enums for stuff that needs to be maintained across versions or communicated across processes, these are not that

tec27 (changes requested):

client/settings/settings.tsx
2Pacalypse-: It's the only way I've found on how to have the border-radius show up on the child Ripple component. Maybe there's a better way to make child components inherit border-radius from their parents?

tec27: `contain: content` (or one of the other options) will likely work

client/chat/channel-settings-dialog.tsx
```diff
@@ -0,0 +1,148 @@
...
+
+        <Divider />
+
+        <ContentContainer>{settingsContent}</ContentContainer>
+      </Container>
+    </Dialog>
+  )
+}
+
+const NavEntryRoot = styled.div<{ $isActive: boolean }>`
```
tec27, line 88: This should probably either be a button or a link, the div won't read well to accessibility tools (this also probably isn't keyboard-navigable as-is, so that should probably be fixed)

server/lib/chat/chat-service.test.ts
```diff
@@ -719,6 +725,102 @@ describe('chat/chat-service', () => {
     })
   })
 
+  describe('editChannel', () => {
+    const expectItWorks = (updates: EditChannelRequest, result: EditChannelResponse) => {
```
tec27, line 729: would vastly prefer that you put in the object to expect to see in this method, rather than constructing that dynamically. As is, it is hard to verify that this method actually gives the output you expect (you really need a test for this test utility), but if you just plug in the actual final object you expect directly it will be obvious and this method will be far simpler

tec27: The general test structure should always be:
1) Do a thing
2) Compare the result(s) of that thing to *simple* values to ensure they match

If you do a bunch of dynamic stuff in part 2, it becomes very hard to know whether the tests are working or not. Them passing or failing is not a great indication of whether something is wrong in the code or just wrong in the complex test code you've written, and you greatly increase the risk that your steps to generate a result are just reproducing the code under test in the first place (including all its bugs/mistakes).

We want:

```ts
const result = 2 + 2
expect(result).toEqual(4)
```

**not**

```ts
const result = 2 + 2
expect(result).toEqual(2 + 2)
```

server/lib/validation/json-validator.test.ts
```diff
@@ -0,0 +1,34 @@
+import Joi from 'joi'
+import { json } from './json-validator'
+
+describe('server/lib/json/json-validator', () => {
+  test('throws an error if the property is not a valid JSON', () => {
```
tec27, line 5: Can you add a test for when the invalid json is inside of a normal Joi object or something? To verify that the field label actually gets applied in the error message. Since the validation message here is using `value` for the root (I guess), it's hard to say whether that is working correctly

tec27 (changes requested):

client/chat/channel-settings-dialog.tsx
```diff
@@ -49,6 +50,14 @@ export function ChannelSettingsDialog({
   const joinedChannelInfo = useAppSelector(s => s.chat.idToJoinedInfo.get(channelId))
   const [section, setSection] = useState<ChannelSettingsSection>(ChannelSettingsSection.General)
 
+  useEffect(() => {
```
tec27, line 53: Don't think you meant to leave this in

tec27 (approved):

PR #1053, opened 2023-10-14
Implement our own version of the language detector.

### LINKED ISSUE #1051: Write new language detection logic
i18next's doesn't seem configurable enough to do what we want, so we should replace it since the logic isn't that hard. Desired logic:

- For languages we provide without dialects, prefer the earliest matching language in `navigator.languages`
- For languages we provide with dialects, prefer only if exact match

The result should always be an entry of our available languages (even if the language we matched to it included/didn't include a dialect)

2Pacalypse-:
i18n's version had some very weird rules, so we just implement our own now. Fairly rudimentary at the moment, and I'm sure it doesn't cover *all* of the possible language codes, but should work for our supported languages.

Fixes #1051

### DIFF
--- a/client/i18n/language-detector.test.ts
+++ b/client/i18n/language-detector.test.ts
@@ -1,0 +1,22 @@
+import { TranslationLanguage } from '../../common/i18n'
+import { getBestLanguage } from './language-detector'
+
+describe('client/i18n/language-detector', () => {
+  test('returns the english language if none of the given languages are supported', () => {
+    const lng = getBestLanguage(['INVALID_LANGUAGE'])
+
+    expect(lng).toBe(TranslationLanguage.English)
+  })
+
+  test('returns the earliest matching language without dialect', () => {
+    const lng = getBestLanguage(['en-US', 'kr'])
+
+    expect(lng).toBe(TranslationLanguage.English)
+  })
+
+  test('returns the exact match for the language with dialect', () => {
+    const lng = getBestLanguage(['zh-Hans', 'en-US'])
+
+    expect(lng).toBe(TranslationLanguage.ChineseSimplified)
+  })
+})
--- a/client/i18n/language-detector.test.ts
+++ b/client/i18n/language-detector.test.ts
@@ -1,22 +1,30 @@
 import { TranslationLanguage } from '../../common/i18n'
 import { getBestLanguage } from './language-detector'
 
+const SUPPORTED_LANGUAGES = ['zh-Hans', 'en', 'ko', 'ru', 'es']
+
 describe('client/i18n/language-detector', () => {
   test('returns the english language if none of the given languages are supported', () => {
-    const lng = getBestLanguage(['INVALID_LANGUAGE'])
+    const lng = getBestLanguage(['INVALID_LANGUAGE'], SUPPORTED_LANGUAGES)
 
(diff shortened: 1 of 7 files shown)


client/i18n/language-detector.test.ts
```diff
@@ -0,0 +1,22 @@
...
+
+describe('client/i18n/language-detector', () => {
+  test('returns the english language if none of the given languages are supported', () => {
+    const lng = getBestLanguage(['INVALID_LANGUAGE'])
+
+    expect(lng).toBe(TranslationLanguage.English)
+  })
+
+  test('returns the earliest matching language without dialect', () => {
+    const lng = getBestLanguage(['en-US', 'kr'])
```
tec27, line 12: One day I am going to get you to stop writing `'kr'` for Korean :P It's `'ko'` (and you probably want `'ko-KR'` for this to be a realistic test)

client/i18n/language-detector.test.ts
```diff
@@ -0,0 +1,22 @@
...
+
+    expect(lng).toBe(TranslationLanguage.English)
+  })
+
+  test('returns the exact match for the language with dialect', () => {
+    const lng = getBestLanguage(['zh-Hans', 'en-US'])
+
+    expect(lng).toBe(TranslationLanguage.ChineseSimplified)
+  })
+})
```
tec27, line 22: This is missing a pretty important test, which is that a language with a dialect will match a non-dialect code (e.g. `['es-MX', 'en-US']` should pick `'es'`)

client/i18n/language-detector.ts
```diff
@@ -0,0 +1,31 @@
...
+
+/**
+ * Choose the preferred language from the list of given languages. If no languages are given, we
+ * default to using `navigator.languages` as a list of available languages. The logic for choosing
+ * the preferred language is:
+ *  - for languages we support without dialects, prefer the earliest matching language in the list
+ *    of given languages.
+ *  - for languages we support with dialects, prefer only if exact match.
+ */
+export function getBestLanguage(languages = navigator.languages) {
```
tec27, line 13: Not making the available languages a parameter means that the unit tests you've written a brittle to changes with that (and it's hard to test cases that may not exist yet just because we don't support some particular language), so I would pass that in as a parameter as well (and update your tests to use some constant list)

client/i18n/language-detector.ts
```diff
@@ -0,0 +1,31 @@
...
+  for (let i = 0; i < languages.length; i++) {
```
tec27, line 14: You don't need this index, just do `for (const lang of languages)` or similar

client/i18n/language-detector.ts
```diff
@@ -0,0 +1,31 @@
...
+    const lng = languages[i].toLowerCase() as TranslationLanguage
+
+    let lngIndex = lowerCaseSupportedLanguages.indexOf(lng)
+    if (lngIndex > -1) {
+      return ALL_TRANSLATION_LANGUAGES[lngIndex]
+    }
+
+    const lngWithoutDialect = lng.split('-')[0]
```
tec27, line 22: if you're ever splitting and just taking the first entry, you should pass a limit to the `split` call to avoid extra work

tec27 (changes requested):

PR #1055, opened 2023-10-28
Add a way for users to upload their channel banner and badge.

2Pacalypse-:


client/chat/channel-settings-dialog.tsx
```diff
@@ -45,106 +76,178 @@ export function ChannelSettingsDialog({
...
 
-      <Ripple ref={rippleRef} />
-    </NavEntryRoot>
+  return (
+    <StyledDialog dialogRef={dialogRef} title={`#${basicChannelInfo?.name}`} buttons={buttons}>
+      <Root>
+        <FormContainer>
+          <StyledForm noValidate={true} onSubmit={onSubmit}>
+            <BannerButtonsContainer>
+              <HiddenFileInput
```
tec27, line 158: I don't think this is really how this should be handled. The file input component should provide a button or otherwise, but still be visible (e.g. definitely *not* `display: none`) so that things like file dragging work on it. If you deal with things properly then implementing those behaviors is free

client/dom/use-object-url.tsx
```diff
@@ -0,0 +1,28 @@
+import { useEffect, useState } from 'react'
+import { useValueAsRef } from '../state-hooks'
+
+/**
+ * A hook which creates an object URL out of the received file, while also making sure that stale
```
tec27, line 5: This hook does not make sure that object URLs are revoked, it only revokes if the URL *changes*. So you need to fix that

client/dom/use-object-url.tsx
```diff
@@ -0,0 +1,28 @@
...
+ * object URLs are revoked.
+ *
+ * If the received argument is already a string, this hook is basically a noop which just returns
+ * the received string.
+ */
+export function useObjectUrl(file?: Blob | string): string | undefined {
+  const [objectUrl, setObjectUrl] = useState<string>()
+  const objectUrlRef = useValueAsRef(objectUrl)
```
tec27, line 13: I would vastly prefer you just set the ref at the same time you set the state with something like this, otherwise you're depending on a render having run 100% of the time in order to revoke things properly (which is not guaranteed). Given that you're just using the state to trigger re-renders, it might be better to just *only* store the URL in a ref and use the stateless rerender hook we already have (to prevent inconsistency)

client/forms/file-input.tsx
```diff
@@ -19,59 +19,90 @@ const ClearButton = styled(IconButton)`
...
-}) {
-  const { t } = useTranslation()
-  const inputRef = useRef<HTMLInputElement>(null)
+}
 
-  const onInputChange = (event: React.ChangeEvent<HTMLInputElement>) => {
-    const files = event.target.files ? Array.from(event.target.files) : []
-    onChange?.(inputProps.multiple ? files : files[0])
-  }
+// Quick and ugly wrap for <input type='file'> that works with rest of the form stuff
```
tec27, line 37: Now that this has gone through like 4 separate modifications I am not sure this comment is really true

tec27 (changes requested):
The deal was that uploaded files were going to go through a moderation service, which you have not implemented at all here. This is not suitable to commit without that functionality. If you would like to commit this piece and work on that after, you need to feature flag this.

tec27 (approved):

PR #1056, opened 2023-10-30
Allow server admins to change channel settings.

2Pacalypse-:
This enables server admins to change channel settings of official channels too.


2Pacalypse-:
Depends on #1055

tec27 (approved):

PR #1057, opened 2023-10-30
Display the channel banner in the channel (behind the messages).

2Pacalypse-:
Also adds an option to users to allow them to hide the banner in case they find it distracting.


2Pacalypse-:
Depends on #1056

tec27 (approved):

PR #1059, opened 2023-11-04
Add the option to make Tooltip interactive.

### LINKED ISSUE #1007: Make the tooltip content interactable
Right now there are two issues that prevent users from interacting with the tooltip content (this makes it hard to have stuff like links inside a tooltip):
- A couple of elements in there have `pointer-events: none` set
- Tooltip goes away when you move the mouse off of the trigger element (unless you've actually focused the trigger element somehow)

For the first point, I can see that the tooltip contents are rendered inside a PopoverContent component, which is inside a Portal component, both of which have `pointer-events: none` set. I'm not sure yet if this is something we can simply remove, or instead have a separate interactable version of the Tooltip component which can be configured through props or something.

For the second point, I assume we're fine with making all tooltips stay open if the mouse moves on top of them, with some padded area between the tooltip trigger element and the tooltip contents.

2Pacalypse-:
Interactive tooltips allow users to interact with the content inside it, e.g. pressing buttons and links.

Also, we're utilizing this interactive tooltip for the topic in channel header.

Closes #1007

### DIFF
--- a/client/material/tooltip.tsx
+++ b/client/material/tooltip.tsx
@@ -57,7 +57,7 @@ const arrowStyle: Record<TooltipPosition, FlattenSimpleInterpolation> = {
   `,
 }
 
-export const TooltipContent = styled.div<{ $position: TooltipPosition }>`
+export const TooltipContent = styled.div<{ $position: TooltipPosition; $interactive?: boolean }>`
   ${caption};
   ${shadow2dp};
 
@@ -72,7 +72,7 @@ export const TooltipContent = styled.div<{ $position: TooltipPosition }>`
   border: 1px solid rgba(255, 255, 255, 0.36);
   border-radius: 4px;
   background-color: ${background900};
-  pointer-events: none;
+  pointer-events: ${props => (props.$interactive ? 'auto' : 'none')};
 
   &::before {
     content: '';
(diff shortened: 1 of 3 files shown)


client/material/tooltip.tsx
```diff
@@ -146,6 +156,10 @@ export function Tooltip({
   const [mouseAnchorElem, setMouseAnchorElem] = useState<HTMLElement>()
   const [focusAnchorElem, setFocusAnchorElem] = useState<HTMLElement>()
   const anchorElem = focusAnchorElem ?? mouseAnchorElem
+  // NOTE(2Pac): We store the tooltip element here to indicate the mousing over the tooltip content
+  // so we can keep the tooltip open in case it's interactive.
+  const [tooltipElem, setTooltipElem] = useState<HTMLElement>()
```
tec27, line 161: if you're just going to use this value as a boolean, then it seems better to just set a boolean instead of storing this

client/material/tooltip.tsx
```diff
@@ -188,10 +208,23 @@ export function Tooltip({
         }
       }
     } else {
-      setOpen(false)
-      return () => {}
+      if (interactive) {
+        let timeout: ReturnType<typeof setTimeout> | undefined = setTimeout(() => {
```
tec27, line 212: I feel like it would probably make more sense to have padding around the tooltip that would keep it open if you were still in the vicinity, rather than delaying the close (I assume that's what this is for anyway, can't really see another purpose for it). Delay on closing can make the interface feel laggy/unresponsive

tec27 (approved):

PR #1061, opened 2023-11-15
Implement a service which checks if the given image is safe.

2Pacalypse-:
The safety of images is checked on following categories, with varying strictness: adult, spoof, medical, violence, and racy.

Images with adult content have the strictest rules, followed by images with medical and violence content, which have a bit less strictness. Finally, images with spoof and racy content have no strictness rules.


sample.env
```diff
@@ -87,3 +87,7 @@ SB_MAP_PARSER_MAX_CONCURRENT=1
 
 # Datadog configuration for log monitoring. If not specified, logs will only be stored locally
 #SB_DATADOG_KEY=DEADBEEF
+
+# Path to the Google Cloud credentials file. Usually downloaded as a JSON file from Google Cloud
+# representing the service account key.
+#GOOGLE_APPLICATION_CREDENTIALS=/path/to/the/credentials.json
```
tec27, line 93: Paths to files won't work for production deployment unless you map the file into the container, so this PR needs to have updated the various docker-compose files to do so. It would probably be good to also provide a mock API for integration tests (I assume there is a way to configure the server this client is talking to, which you will need to implement this) + enforce that this service is always configured in production, as silently not moderating images seems bad

tec27 (changes requested):

Dockerfile
```diff
@@ -85,6 +85,7 @@ COPY --chown=node:node --from=builder /shieldbattery/server/deployment_files/ent
 # Allow the various scripts to be run (necessary when building on Linux)
 RUN chmod +x ./server/update_server.sh
 RUN chmod +x ./server/testing/run_mailgun.sh
+RUN chmod +x ./server/testing/run_google_cloud.sh
```
tec27, line 88: Can you combine all these `chmod`s into a single line (you should just be able to append the other file paths onto the original command). Docker generates new layers every time you run a command so ideally we want to minimize that

deployment/appserver/docker-compose.yml
```diff
@@ -53,11 +53,13 @@ services:
       - SB_ANALYTICS_ID
       - SB_DATADOG_KEY
       - SB_DISABLE_CRAWLERS
+      - GOOGLE_APPLICATION_CREDENTIALS=/google-credentials.json
     networks:
       - frontend
       - backend
     volumes:
       - app_logs:/home/node/shieldbattery/server/logs:rw
+      - ./google-credentials.json:/google-credentials.json
```
tec27, line 62: add `:ro` after this, don't want the container to be able to modify the file

integration/docker-compose.yml
```diff
@@ -37,6 +37,13 @@ services:
       - 'SB_MAILGUN_FROM=ShieldBattery <shieldbattery@example.org>'
       - 'SB_MAILGUN_URL=http://mailgun:5528'
       - SB_JWT_SECRET=SuperSecretJwtKey
+      - GOOGLE_APPLICATION_CREDENTIALS=/google-fake-credentials.json
+      - SB_USE_FAKE_GOOGLE_CLOUD=true
+      - 'SB_GOOGLE_CLOUD_DOMAIN=google-cloud'
+      - SB_GOOGLE_CLOUD_PORT=5530
+    volumes:
+      - ./google-fake-credentials.json:/google-fake-credentials.json
```
tec27, line 45: Same as the other one, make it read-only

integration/tests/chat/channel-settings.spec.ts
```diff
@@ -0,0 +1,91 @@
...
+
+let loginPage: LoginPage
+let chatPage: ChatPage
+
+test.beforeEach(async ({ page }) => {
+  loginPage = new LoginPage(page)
+  chatPage = new ChatPage(page)
+})
+
+test('changing channel banner', async ({ page }) => {
```
tec27, line 15: Part of the point of these tests is that they should also test what happens when an innappropriate image is uploaded, so you need a test that does that too. You'll need to flesh out the fake google API to do this, probably just hash the image or check its filesize or something to decide whether it's inappropriate, don't think it really matters too much.

deployment/appserver/docker-compose.yml
```diff
@@ -53,11 +53,13 @@ services:
       - SB_ANALYTICS_ID
       - SB_DATADOG_KEY
       - SB_DISABLE_CRAWLERS
+      - GOOGLE_APPLICATION_CREDENTIALS=/google-credentials.json
```
tec27, line 56: What you wanted was "a file named `google-credentials.json` next to this file", what you wrote was "a file named `google-credentials.json` at the root of the filesystem"

tec27 (changes requested):

deployment/appserver/docker-compose.yml
tec27: Oh nevermind, I forgot this is going to be where it is inside the container, this is fine

integration/tests/chat/channel-settings.spec.ts
```diff
@@ -25,13 +33,30 @@ test('changing channel banner', async ({ page }) => {
...
+  }
+  await loginPage.navigateTo()
+  await loginPage.loginWith('admin', 'admin1234')
+
+  await chatPage.openChannelSettings()
+
+  await chatPage.setChannelBanner(path.join(TEST_IMAGE_PATH_INAPPROPRIATE))
+  await chatPage.clickChannelSettingsSaveButton()
+
+  const errorMessage = await chatPage.getChannelSettingsSnackbarError()
```
tec27, line 58: Wouldn't it make a lot more sense to show this error within the dialog? Using a snackbar for this seems very weird

tec27 (changes requested):

client/chat/channel-settings-dialog.tsx
```diff
@@ -149,114 +168,145 @@ export function ChannelSettingsDialog({
...
+  if (error) {
+    if (isFetchError(error) && error.code === ChatServiceErrorCode.InappropriateImage) {
+      errorMessage = t(
+        'chat.channelSettings.general.inappropriateImageErrorMessage',
+        'The selected image is inappropriate. Please select a different image.',
+      )
+    } else {
+      errorMessage = t(
+        'chat.channelSettings.general.saveErrorMessage',
+        'Something went wrong saving the settings',
```
tec27, line 187: nit: `while saving`

tec27 (approved):

PR #1067, opened 2023-12-10
Add a player's full match history to their public profiles.

2Pacalypse-:
There's no way to filter this list yet. That will come in a future commit.


tec27 (approved):
Would really prefer we implement the list query in the gql server but I guess this is fine for now if you're not comfortable doing that

2Pacalypse-:
I would love to do that, but I feel like it would make me unproductive for way too long if I start going down that path now.

But once I finally start learning Rust, my first contributions will most likely be rewriting existing things (like this API) in Rust.

PR #1072, opened 2024-01-20
Update SB version in local storage as soon as changelog dialog is opened.

2Pacalypse-:
Doing it when the dialog was about to be closed complicated things because we would have to handle closing the dialog through Dialog or through Scrim separately.

Also, moves the changelog dialog to TS/hooks.


client/changelog/changelog-dialog.tsx
```diff
@@ -62,45 +63,22 @@ const ChangelogLoadable = React.lazy(async () => {
...
-  }
-
-  onStorageChange = e => {
-    if (!this._setting && e.key === KEY) {
-      if (!shouldShowChangelog()) {
-        this.onDismiss()
-      }
+export function ChangelogDialog(props: CommonDialogProps) {
+  useEffect(() => {
+    if (shouldShowChangelog()) {
```
tec27, line 68: this check should be unnecessary I think

tec27 (approved):

client/changelog/changelog-dialog.tsx
2Pacalypse-: Yeah, the only reason I left it is because people can open the changelog manually whenever; but if writing to the local storage is not an expensive operation, might as well always do it.

tec27: It's not like this is happening every frame, it happens once when the dialog opens

tec27: This also involves a local storage read to even do this check, so it's not really "free" if trying to avoid expensive operations was even necessary here

PR #1073, opened 2024-01-20
Fix the spacing/padding issues in multiline TextFields.

### LINKED ISSUE #1068: Fix multi-line text-field and textarea spacing/padding
This is how it currently looks (I think it's even worse when there are trailing icons):

image

2Pacalypse-:
Also fixes a bug which wouldn't render the correct size of textarea if the initial value was causing it to have multiple rows.

Fixes #1068

### DIFF
--- a/client/material/text-field.tsx
+++ b/client/material/text-field.tsx
@@ -1,4 +1,4 @@
-import React, { useCallback, useEffect, useId, useState } from 'react'
+import React, { useCallback, useId, useLayoutEffect, useState } from 'react'
 import styled, { css } from 'styled-components'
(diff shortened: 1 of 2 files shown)


client/material/text-field.tsx
```diff
@@ -212,6 +218,12 @@ export const TextField = React.forwardRef<HTMLInputElement, TextFieldProps>(
       [dense, rows],
     )
 
+    useEffect(() => {
```
tec27, line 221: You want this in a `useLayoutEffect` so the resize happens before the browser paints, otherwise you'll get a flash of the smaller size

tec27 (approved):

PR #1087, opened 2024-07-28
Properly clean up state when users leave during countdown/loading.

### LINKED ISSUE #663: Unable to join lobby after the countdown gets cancelled
Trying to join a lobby after its countdown gets cancelled returns a following error message:
```
Uncaught (in promise) Error: lobby has already started
```

2Pacalypse-:
So there are 2 issues that were happening in the previous code when user would leave the lobby:
- We would publish the `cancelLoading` event only to the user who left, instead of the whole lobby.
- We would delete the lobby from the loading state only in case everyone left, instead of always.

I made the loading cleanup a bit more consistent now and similar to how/when cancel countdown happens. The only slightly weird thing here is that we publish the event to the lobby list to add the lobby back to the list when the loading gets canceled. However this is also done when cleaning up the countdown state as well, so we have two events fire each time, because those two cleanup functions are called in unison. We could perhaps combine these two functions into one, but I don't think it's a big deal to publish the event twice either?

I also fixed a rare bug that would happen if the last user would leave the lobby during countdown/loading (this is possible if you start the game with a computer and leave). The lobby would get deleted, but the cancel countdown/loading cleanup functions would publish the "add" event back to the lobby list, making the clients show a non-existent lobby. This wasn't a huge issue since it would only show the non-existent lobby if someone else already had the lobby list open. As soon as they close it and open it again they would get a full lobby list without the non-existent lobby.

Fixes #663

### DIFF
--- a/server/lib/wsapi/lobbies.js
+++ b/server/lib/wsapi/lobbies.js
@@ -632,9 +632,10 @@ export class LobbyApi {
   _removeClientFromLobby(lobby, client, removalType = REMOVAL_TYPE_NORMAL) {
     const [teamIndex, slotIndex, player] = findSlotByName(lobby, client.name)
     const updatedLobby = Lobbies.removePlayer(lobby, teamIndex, slotIndex, player)
+    const isLobbyEmpty = !updatedLobby
 
-    if (!updatedLobby) {
-      // The lobby is now empty and needs to be removed the this list
+    if (isLobbyEmpty) {
+      // The lobby is now empty and needs to be removed from the list
 
       // Ensure the client's local state gets updated to confirm the leave
       this._publishTo(lobby, {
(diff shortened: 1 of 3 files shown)


server/lib/wsapi/lobbies.js
```diff
@@ -816,7 +806,7 @@ export class LobbyApi {
     })
   }
 
-  _onLoadingCanceled(lobby) {
+  _maybeCancelLoading(lobby, isLobbyEmpty) {
```
tec27, line 809: probably default this new param to false since it isn't always passed? End result is probably the same but it's a bit more correct and will prevent some de-optimizations or something (doubt they matter very much here but whatever :) )

server/lib/wsapi/lobbies.js
```diff
@@ -852,7 +844,7 @@ export class LobbyApi {
   }
 
   // Cancels the countdown if one was occurring (no-op if it was not)
-  _maybeCancelCountdown(lobby) {
+  _maybeCancelCountdown(lobby, isLobbyEmpty) {
```
tec27, line 847: also add default of false here

tec27 (approved):

PR #1088, opened 2024-09-08
Add a better throttle for image uploads that require safety checks.

### LINKED ISSUE #1070: Add a better throttle for image uploads that require validation
Users should generally be able to change these far less than the text content of related settings (e.g. channel settings) because they are much more expensive to validate. Currently however, we use the same throttle for both, so we have to leave it throttled higher than we would probably like to.

2Pacalypse-:
Closes #1070


server/lib/chat/chat-api.ts
```diff
@@ -174,10 +180,12 @@ export class ChatApi {
   }
 
   @httpPatch('/:channelId')
-  @httpBefore(
-    throttleMiddleware(editThrottle, ctx => String(ctx.session!.user!.id)),
-    handleMultipartFiles(MAX_IMAGE_SIZE),
-  )
+  @httpBefore(handleMultipartFiles(MAX_IMAGE_SIZE), async (ctx: ExtendableContext, next: Next) => {
```
tec27, line 183: can you pull this into a named function instead of doing this inline?

tec27 (approved):

PR #1093, opened 2024-09-10
Update aws-sdk to v3.

2Pacalypse-:
Aws sdk v2 has been deprecated:
https://aws.amazon.com/blogs/developer/announcing-end-of-support-for-aws-sdk-for-javascript-v2/

Even though there's still a year until the end-of-support, I thought I'd be proactive and update to v3 now.

This was mostly done by running their codemod and fixing a couple of issues manually, e.g.

`npx aws-sdk-js-codemod -t v2-to-v3 ./server/lib/file-upload/aws.ts`

One thing I have *not* done yet is setup DO spaces to be use locally which would actually test this code. That seems like a pretty important step tho :d. However, in my defense, setting up that locally is a hassle and I forgot how to do it, so we can just leave this in open PR now until that is done.

### DIFF
--- a/sample.env
+++ b/sample.env
@@ -67,7 +67,7 @@ SB_RALLY_POINT_LOCAL_PORT=14098
 # two options, storing to the filesystem and storing to the DigitalOcean Spaces.
 SB_FILE_STORE={"filesystem":{"path":"server/uploaded_files"}}
 # Example DigitalOcean Spaces configuration
-#SB_FILE_STORE={"doSpaces":{"endpoint":"region.digitaloceanspaces.com","accessKeyId":"ACCESS_KEY_ID","secretAccessKey":"SUPER_SECRET_ACCESS_KEY","bucket":"shieldbattery","cdnHost":"cdn.example.org"}}
+#SB_FILE_STORE={"doSpaces":{"endpoint":"https://region.digitaloceanspaces.com","accessKeyId":"ACCESS_KEY_ID","secretAccessKey":"SUPER_SECRET_ACCESS_KEY","bucket":"shieldbattery","cdnHost":"cdn.example.org"}}
 
 
 # Configuration for BW sprite data used for map thumbnail generation
--- a/server/lib/file-upload/aws.ts
+++ b/server/lib/file-upload/aws.ts
@@ -60,7 +60,10 @@ export default class Aws implements FileStore {
 
     // DO Spaces use endpoints
     if (endpoint) {
+      // AWS SDK v3 constructs a new URL() out of this, so it needs to have a "https://" prefix.
       options.endpoint = endpoint
+      // AWS SDK v3 requires `region` to be set, even if it's not used.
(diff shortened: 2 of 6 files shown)


sample.env
```diff
@@ -67,7 +67,7 @@ SB_RALLY_POINT_LOCAL_PORT=14098
 # two options, storing to the filesystem and storing to the DigitalOcean Spaces.
 SB_FILE_STORE={"filesystem":{"path":"server/uploaded_files"}}
 # Example DigitalOcean Spaces configuration
-#SB_FILE_STORE={"doSpaces":{"endpoint":"region.digitaloceanspaces.com","accessKeyId":"ACCESS_KEY_ID","secretAccessKey":"SUPER_SECRET_ACCESS_KEY","bucket":"shieldbattery","cdnHost":"cdn.example.org"}}
+#SB_FILE_STORE={"doSpaces":{"endpoint":"https://region.digitaloceanspaces.com","accessKeyId":"ACCESS_KEY_ID","secretAccessKey":"SUPER_SECRET_ACCESS_KEY","bucket":"shieldbattery","cdnHost":"cdn.example.org"}}
```
tec27, line 70: Please check if the provided endpoint has a protocol and if not, add `https://` to the front of it. I don't want to have to remember to update all the existing configs we have to add this

server/lib/file-upload/aws.ts
```diff
@@ -101,33 +111,34 @@ export default class Aws implements FileStore {
...
+    const files = await this.client.listObjectsV2({ Prefix: normalized, Bucket: this.bucket })
     const keys = (files.Contents ?? []).map(file => ({ Key: file.Key }))
     const params = { Delete: { Objects: keys }, Bucket: this.bucket, ...formatOptions(options) }
-    return this.client.deleteObjects(params).promise()
+    return this.client.deleteObjects(params)
   }
 
   url(filename: string, options: any = {}): string {
-    const url = `https://${this.bucket}.${this.endpoint}/${filename}`
+    const hostname = new URL(this.endpoint).hostname
```
tec27, line 140: Given that the endpoint never changes, it feels nicer to just calculate this when this object is constructed and not re-parse the URL every time we need a url

tec27 (changes requested):

server/lib/file-upload/aws.ts
```diff
@@ -61,16 +61,21 @@ export default class Aws implements FileStore {
     // DO Spaces use endpoints
     if (endpoint) {
       // AWS SDK v3 constructs a new URL() out of this, so it needs to have a "https://" prefix.
-      options.endpoint = endpoint
+      const endpointWithPrefix = endpoint.startsWith('https://') ? endpoint : `https://${endpoint}`
```
tec27, line 64: I think you should make this more general in case we need to set a different endpoint for testing, etc. Just check if it matches `/^[^:/]+:\/\//`

tec27 (approved):

PR #1103, opened 2024-10-10
Add a helper class to simplify the usage of the `BatchedAbortSignals`.

### LINKED ISSUE #1101: Reproduce implementation of request coalescing from `viewUserProfile` wherever we coalesce requests with potential AbortSignals
All the other places we do that are likely bugged in the same way as #828 (even if we haven't seen the bug yet).

It would probably be nice to create a helper class to make the implementation easier, as the logic is a bit easy to mess up (especially around not adding a new AbortSignal to a BatchedAbortSignals that has already been aborted). Some kind of `RequestCoalescer` class that lets you register a request/signal based on a given ID.

e.g.
```ts
await requestCoalescer.makeRequest(id, spec.signal, async () => {
  dispatch({
    /* ... */
  })
})

class RequestCoalescer {
  /* ... */
  async makeRequest(id, signal, requestFn): Promise<void> {
    let batchedSignals = this.inProgress.get(id)
    if (batchedSignals && !batchedSignals.aborted) {
      batchedSignals.add(signal)
      return
    }
    
    batchedSignals = new BatchedAbortSignals(signal)
    this.inProgress.set(id, batchedSignals)
    try {
      await requestFn()
    } finally {
      if (this.inProgress.get(id) === batchedSignals) {
        this.inProgress.delete(id)
      }
    }
}
```

2Pacalypse-:
I think I got all of our usages with this? Ended up being fewer than I initially thought.

Also, while searching for these usages, I've deleted a few unused states from the user's reducer.

Closes #1101

### DIFF
--- a/client/chat/action-creators.ts
+++ b/client/chat/action-creators.ts
@@ -279,19 +279,17 @@ export function getChatUserProfile(
   spec: RequestHandlingSpec<void>,
 ): ThunkAction {
   return abortableThunk(spec, async (dispatch, getStore) => {
-    const channelTargetId: `${SbChannelId}|${SbUserId}` = `${channelId}|${targetId}`
-
     await getChatUserProfileRequestCoalescer.makeRequest(
-      channelTargetId,
+      `${channelId}|${targetId}`,
       spec.signal,
-      async (batchedSignal: AbortSignal) => {
+      async (signal: AbortSignal) => {
         dispatch({
           type: '@chat/getChatUserProfile',
           payload: await fetchJson<GetChatUserProfileResponse>(
             apiUrl`chat/${channelId}/users/${targetId}`,
             {
               method: 'GET',
-              signal: batchedSignal,
+              signal,
             },
           ),
         })
--- a/client/chat/action-creators.ts
+++ b/client/chat/action-creators.ts
@@ -1,3 +1,4 @@
+import { RequestCoalescer } from '../../common/async/abort-signals'
 import {
   ChannelModerationAction,
   ChatServiceErrorCode,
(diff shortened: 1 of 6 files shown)


client/chat/action-creators.ts
```diff
@@ -270,34 +271,32 @@ export function retrieveUserList(channelId: SbChannelId): ThunkAction {
...
 
-const chatUserProfileLoadsInProgress = new Set<`${SbChannelId}|${SbUserId}`>()
-
 export function getChatUserProfile(
   channelId: SbChannelId,
   targetId: SbUserId,
   spec: RequestHandlingSpec<void>,
 ): ThunkAction {
   return abortableThunk(spec, async (dispatch, getStore) => {
+    const requestCoalescer = new RequestCoalescer()
```
tec27, line 280: This doesn't work, you're making a new coalescer for every request so it will coalesce nothing. It needs to be global

tec27 (changes requested):

client/chat/action-creators.ts
2Pacalypse-: Ah, that's a silly mistake.

You know, when I first started implementing this, I actually didn't even use a class. I just wrote a function and had the in progress requests saved in a global Map. But then I realized that it would be hard to ensure that the IDs we use are always unique for different APIs. So I deleted it and went with the class approach.

But maybe using a function is not a bad idea after all? And for the in progress Map keys, we just use whatever we call in `apiUrl`, e.g. `users/${userId}/profile`, `chat/${channelId}/users/${targetId}`, and `games/${gameId}`.

tec27: No, I think you should use the class as you have an create a global instance for each particular action that needs to be coalesced

common/async/abort-signals.ts
```diff
@@ -74,3 +74,34 @@ export class BatchedAbortSignals {
     }
   }
 }
+
+/**
```
tec27, line 78: This should be elsewhere, this doesn't really have much to do with abort signals

common/async/abort-signals.ts
```diff
@@ -74,3 +74,34 @@ export class BatchedAbortSignals {
...
+ * A helper class which allows you to use `BatchedAbortSignals` more easily. See the description
+ * of that class for more info.
+ */
+export class RequestCoalescer {
+  private requestsInProgress = new Map<unknown, BatchedAbortSignals>()
+
+  async makeRequest(
+    id: unknown,
```
tec27, line 86: probably this class should be made generic and the generic type should be the ID type, no reason not to have type safety around that

client/chat/action-creators.ts
```diff
@@ -279,25 +280,23 @@ export function getChatUserProfile(
 ): ThunkAction {
   return abortableThunk(spec, async (dispatch, getStore) => {
     const channelTargetId: `${SbChannelId}|${SbUserId}` = `${channelId}|${targetId}`
```
tec27, line 282: nit: this is only used in one place now so you can just inline it and avoid needing to declare the type explicitly

client/network/request-coalescer.ts
```diff
@@ -0,0 +1,32 @@
+import { BatchedAbortSignals } from '../../common/async/abort-signals'
+
+/**
+ * A helper class which allows you to use `BatchedAbortSignals` more easily. See the description
```
tec27, line 4: Please make a comment that stands on its own, this class is for coalescing requests that take AbortSignals, the BatchedAbortSignals piece is an implementation detail and not something the consumer of this API needs to worry about

client/network/request-coalescer.ts
```diff
@@ -0,0 +1,32 @@
...
+ * of that class for more info.
+ */
+export class RequestCoalescer<T> {
+  private requestsInProgress = new Map<T, BatchedAbortSignals>()
+
+  async makeRequest(
+    id: T,
+    signal: AbortSignal | undefined,
+    requestFn: (batchedSignal: AbortSignal) => Promise<void>,
```
tec27, line 13: nit: renamed `batchedSignal` to `signal`

tec27 (approved):

PR #1117, opened 2024-12-28
Display user's ranking history in their profiles.

### LINKED ISSUE #1107: Add historical rank information to profile pages
`matchmaking_finalized_ranks` joined with `matchmaking_ratings` will give you this info. I think it probably makes sense to have a dropdown near the current rank info to switch seasons (defaulting to the current season). I don't think this needs to be part of the URL/route structure. Could probably always return all the relevant seasons with the user profile data, don't think it would end up being that much data overall (and this would let you filter seasons in the dropdown to just the ones that are relevant for that user).

2Pacalypse-:
The ranking history is displayed under the "Seasons" tab and only includes past seasons in which user had a rank. Doesn't include the current season.

Closes #1107

### DIFF
--- a/client/users/user-profile-seasons.tsx
+++ b/client/users/user-profile-seasons.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useRef, useState } from 'react'
+import React, { useEffect, useState } from 'react'
 import { useTranslation } from 'react-i18next'
 import styled from 'styled-components'
 import { appendToMultimap } from '../../common/data-structures/maps'
@@ -40,19 +40,18 @@ export function UserProfileSeasons({ user }: { user: SbUser }) {
 
   const [isLoading, setIsLoading] = useState(false)
   const [error, setError] = useState<Error>()
-  const abortControllerRef = useRef<AbortController>()
 
   useEffect(() => {
     setIsLoading(true)
 
-    abortControllerRef.current?.abort()
-    abortControllerRef.current = new AbortController()
+    const abortController = new AbortController()
 
     dispatch(
       getUserRankingHistory(user.id, {
-        signal: abortControllerRef.current.signal,
+        signal: abortController.signal,
         onSuccess: data => {
           setIsLoading(false)
+          setError(undefined)
           setRanks(data.history)
         },
         onError: err => {
@@ -61,11 +60,11 @@ export function UserProfileSeasons({ user }: { user: SbUser }) {
         },
       }),
     )
-  }, [dispatch, user.id])
 
-  useEffect(() => {
-    return () => abortControllerRef.current?.abort()
-  }, [])
+    return () => {
+      abortController.abort()
+    }
+  }, [dispatch, user.id])
 
   if (isLoading) {
     return <LoadingDotsArea />
--- a/client/users/user-profile-seasons.tsx
+++ b/client/users/user-profile-seasons.tsx
@@ -1,0 +1,146 @@
+import React, { useEffect, useRef, useState } from 'react'
+import { useTranslation } from 'react-i18next'
+import styled from 'styled-components'
+import { appendToMultimap } from '../../common/data-structures/maps'
+import { LadderPlayer } from '../../common/ladder/ladder'
+import { SeasonId } from '../../common/matchmaking'
+import { SbUser } from '../../common/users/sb-user'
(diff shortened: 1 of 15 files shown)


client/users/user-profile-seasons.tsx
```diff
@@ -0,0 +1,146 @@
...
+        onError: err => {
+          setIsLoading(false)
+          setError(err)
+        },
+      }),
+    )
+  }, [dispatch, user.id])
+
+  useEffect(() => {
+    return () => abortControllerRef.current?.abort()
```
tec27, line 67: Think this will cause issues, it should either be part of the previous effect or not included at all

client/users/user-profile-seasons.tsx
```diff
@@ -0,0 +1,146 @@
...
+  useEffect(() => {
+    setIsLoading(true)
+
+    abortControllerRef.current?.abort()
+    abortControllerRef.current = new AbortController()
+
+    dispatch(
+      getUserRankingHistory(user.id, {
+        signal: abortControllerRef.current.signal,
+        onSuccess: data => {
```
tec27, line 54: Should probably be clearing the error here as well

server/lib/matchmaking/models.ts
```diff
@@ -433,6 +447,45 @@ export async function getMatchmakingRatingsForUser(
...
+      FROM
+        matchmaking_finalized_ranks mfr
+        INNER JOIN matchmaking_ratings mr ON mfr.season_id = mr.season_id
+          AND mfr.user_id = mr.user_id
+          AND mfr.matchmaking_type = mr.matchmaking_type
+        INNER JOIN matchmaking_seasons ms ON mfr.season_id = ms.id
+      WHERE
+        mfr.user_id = ${userId}
+      ORDER BY
+        ms.start_date DESC
```
tec27, line 480: Can we have an explicit ordering for the matchmaking types within the same season as well?

server/lib/users/user-api.ts
```diff
@@ -132,6 +137,12 @@ const matchHistoryRetrievalThrottle = createThrottle('matchhistoryretrieval', {
   window: 60000,
 })
 
+const rankingsHistoryRetrievalThrottle = createThrottle('rankingshistoryretrieval', {
+  rate: 50,
```
tec27, line 141: I don't think 50 requests per minute is at all a reasonable rate for this API, this is very much in "someone is scraping the entire userbase" territory, not at all what the general usage would be for real users

tec27 (changes requested):

server/lib/matchmaking/models.ts
```diff
@@ -477,7 +477,7 @@ export async function getMatchmakingFinalizedRanksForUser(
       WHERE
         mfr.user_id = ${userId}
       ORDER BY
-        ms.start_date DESC
+        ms.start_date DESC, mfr.matchmaking_type
```
tec27, line 480: We may need to re-sort the enum (if we can?), I'm not sure how it is ordered currently but we weren't ordering it based on an expectation of it sorting the UIs and such as well

tec27 (approved):

server/lib/matchmaking/models.ts
2Pacalypse-: PostgreSQL [documentation](https://www.postgresql.org/docs/current/datatype-enum.html#DATATYPE-ENUM-ORDERING) says this about the ordering of the enum type:
```
The ordering of the values in an enum type is the order in which the values were listed when the type was
created. All standard comparison operators and related aggregate functions are supported for enums.
```

In our case that makes it `1v1, 2v2, 1v1fastest` so far. Which seems fine to me? Although might get a bit tricky to think about that once we add more matchmaking types.

Also, maybe we should sort by games count in each matchmaking type?

tec27: I think I'd prefer it to keep the 1v1 types grouped, especially going forward. Sorting by game count could be okay but I think it might make it hard to read different people's profiles (but maybe it'll be hard anyway since it hides empty ones)

So I guess maybe sorting by game count or sorting by top points (in many cases these are probably the same sort order because of how points work) would be fine, and not require us to do whatever (probably annoying) work is necessary to reorder the enum.

PR #1118, opened 2024-12-30
Add historical rankings to the ladder page.

### LINKED ISSUE #1108: Add historical rankings to ladder page
We can add a new dropdown on this page to select seasons (defaulting to the current one). Probably this should be part of the URL/route structure, which might mean it has to be at the end now?

`matchmaking_finalized_ranks` will give you the ordering + ranks to apply to `matchmaking_ratings`

2Pacalypse-:
This adds a select for past seasons on the ladder page, defaulting to the current season.

The action creators and reducers for all of this state got a bit messy, but not sure if there's a better way to deal with it.

Closes #1108

### DIFF
--- a/client/ladder/action-creators.ts
+++ b/client/ladder/action-creators.ts
@@ -22,7 +22,7 @@ export function getCurrentSeasonRankings(
 ): ThunkAction {
   return abortableThunk(spec, async dispatch => {
     const fetchTime = performance.now()
-    const cacheKey = `${matchmakingType}|undefined` as MatchmakingTypeAndSeasonId
+    const cacheKey: MatchmakingTypeAndSeasonId = `${matchmakingType}|undefined`
     const lastFetchTime = lastFetchTimeByMatchmakingTypeAndSeasonId.get(cacheKey)
 
     if (lastFetchTime !== undefined && fetchTime - lastFetchTime < LADDER_RANKINGS_CACHE_TIME_MS) {
@@ -58,7 +58,7 @@ export function getPreviousSeasonRankings(
 ): ThunkAction {
   return abortableThunk(spec, async dispatch => {
     const fetchTime = performance.now()
-    const cacheKey = `${matchmakingType}|${seasonId}` as MatchmakingTypeAndSeasonId
+    const cacheKey: MatchmakingTypeAndSeasonId = `${matchmakingType}|${seasonId}`
     const lastFetchTime = lastFetchTimeByMatchmakingTypeAndSeasonId.get(cacheKey)
 
     if (lastFetchTime !== undefined && fetchTime - lastFetchTime < LADDER_RANKINGS_CACHE_TIME_MS) {
@@ -92,7 +92,7 @@ export function searchCurrentSeasonRankings(
 ): ThunkAction {
   return abortableThunk(spec, async dispatch => {
     const fetchTime = performance.now()
-    const cacheKey = `${matchmakingType}|undefined` as MatchmakingTypeAndSeasonId
+    const cacheKey: MatchmakingTypeAndSeasonId = `${matchmakingType}|undefined`
     let { fetchTime: lastFetchTime, searchQuery: lastSearchQuery } =
       lastSearchInfoByMatchmakingTypeAndSeasonId.get(cacheKey) ?? {}
 
@@ -143,7 +143,7 @@ export function searchPreviousSeasonRankings(
 ): ThunkAction {
   return abortableThunk(spec, async dispatch => {
     const fetchTime = performance.now()
-    const cacheKey = `${matchmakingType}|${seasonId}` as MatchmakingTypeAndSeasonId
+    const cacheKey: MatchmakingTypeAndSeasonId = `${matchmakingType}|${seasonId}`
     let { fetchTime: lastFetchTime, searchQuery: lastSearchQuery } =
       lastSearchInfoByMatchmakingTypeAndSeasonId.get(cacheKey) ?? {}
 
--- a/client/ladder/action-creators.ts
+++ b/client/ladder/action-creators.ts
@@ -1,10 +1,5 @@
 import { GetRankForUserResponse, GetRankingsResponse } from '../../common/ladder/ladder'
-import {
-  makeMatchmakingTypeAndSeasonId,
-  MatchmakingType,
-  MatchmakingTypeAndSeasonId,
-  SeasonId,
-} from '../../common/matchmaking'
+import { MatchmakingType, SeasonId } from '../../common/matchmaking'
 import { apiUrl, urlPath } from '../../common/urls'
 import { ThunkAction } from '../dispatch-registry'
 import { push } from '../navigation/routing'
@@ -13,9 +8,13 @@ import { fetchJson } from '../network/fetch'
 
 const LADDER_RANKINGS_CACHE_TIME_MS = 60 * 1000
 
+type MatchmakingTypeAndSeasonId = `${MatchmakingType}|${SeasonId | undefined}`
+
 const lastFetchTimeByMatchmakingTypeAndSeasonId = new Map<MatchmakingTypeAndSeasonId, number>()
-const lastSearchTimeByMatchmakingTypeAndSeasonId = new Map<MatchmakingTypeAndSeasonId, number>()
-const lastSearchQueryByMatchmakingTypeAndSeasonId = new Map<MatchmakingTypeAndSeasonId, string>()
+const lastSearchInfoByMatchmakingTypeAndSeasonId = new Map<
+  MatchmakingTypeAndSeasonId,
+  { fetchTime: number; searchQuery: string }
+>()
 
 export function getCurrentSeasonRankings(
   matchmakingType: MatchmakingType,
@@ -23,7 +22,7 @@ export function getCurrentSeasonRankings(
 ): ThunkAction {
   return abortableThunk(spec, async dispatch => {
     const fetchTime = performance.now()
-    const cacheKey = makeMatchmakingTypeAndSeasonId(matchmakingType)
+    const cacheKey = `${matchmakingType}|undefined` as MatchmakingTypeAndSeasonId
     const lastFetchTime = lastFetchTimeByMatchmakingTypeAndSeasonId.get(cacheKey)
 
     if (lastFetchTime !== undefined && fetchTime - lastFetchTime < LADDER_RANKINGS_CACHE_TIME_MS) {
@@ -53,13 +52,13 @@ export function getCurrentSeasonRankings(
 }
 
 export function getPreviousSeasonRankings(
-  seasonId: SeasonId,
   matchmakingType: MatchmakingType,
+  seasonId: SeasonId,
   spec: RequestHandlingSpec<void>,
 ): ThunkAction {
   return abortableThunk(spec, async dispatch => {
     const fetchTime = performance.now()
-    const cacheKey = makeMatchmakingTypeAndSeasonId(matchmakingType, seasonId)
+    const cacheKey = `${matchmakingType}|${seasonId}` as MatchmakingTypeAndSeasonId
     const lastFetchTime = lastFetchTimeByMatchmakingTypeAndSeasonId.get(cacheKey)
 
     if (lastFetchTime !== undefined && fetchTime - lastFetchTime < LADDER_RANKINGS_CACHE_TIME_MS) {
@@ -93,13 +92,13 @@ export function searchCurrentSeasonRankings(
 ): ThunkAction {
   return abortableThunk(spec, async dispatch => {
     const fetchTime = performance.now()
-    const cacheKey = makeMatchmakingTypeAndSeasonId(matchmakingType)
-    const lastSearchTime = lastSearchTimeByMatchmakingTypeAndSeasonId.get(cacheKey)
-    const lastSearchQuery = lastSearchQueryByMatchmakingTypeAndSeasonId.get(cacheKey)
+    const cacheKey = `${matchmakingType}|undefined` as MatchmakingTypeAndSeasonId
+    let { fetchTime: lastFetchTime, searchQuery: lastSearchQuery } =
+      lastSearchInfoByMatchmakingTypeAndSeasonId.get(cacheKey) ?? {}
 
     if (
-      lastSearchTime !== undefined &&
-      fetchTime - lastSearchTime < LADDER_RANKINGS_CACHE_TIME_MS &&
+      lastFetchTime !== undefined &&
+      fetchTime - lastFetchTime < LADDER_RANKINGS_CACHE_TIME_MS &&
       lastSearchQuery === searchQuery
     ) {
       return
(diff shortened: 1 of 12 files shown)


client/ladder/action-creators.ts
```diff
@@ -12,6 +12,13 @@ const lastFetchTimeByMatchmakingType = new Map<MatchmakingType, number>()
 const lastSearchTimeByMatchmakingType = new Map<MatchmakingType, number>()
 const lastSearchQueryByMatchmakingType = new Map<MatchmakingType, string>()
 
+const lastFetchTimeBySeasonIdAndMatchmakingType = new Map<SeasonId, Map<MatchmakingType, number>>()
```
tec27, line 15: These could both be converted to strings so you don't need to keep a nested Map, e.g. `${seasonId}|${matchmakingType}`

tec27: I also don't really get why the ones above here (that don't include a season ID) still exist?

client/ladder/action-creators.ts
```diff
@@ -97,6 +208,6 @@ export function getInstantaneousSelfRank(spec: RequestHandlingSpec<void>): Thunk
 }
 
 /** Navigates to the ladder standings (optionally for a given matchmaking type). */
-export function navigateToLadder(type?: MatchmakingType, transitionFn = push) {
-  transitionFn(urlPath`/ladder/${type ?? ''}`)
+export function navigateToLadder(type?: MatchmakingType, seasonId?: SeasonId, transitionFn = push) {
+  transitionFn(urlPath`/ladder/${type ?? ''}` + (seasonId ? urlPath`/${seasonId}` : ''))
```
tec27, line 212: This doesn't really seem correct, it is possible to not specify a matchmaking type but specify a season ID with this API and that generates URLs that are nonsensical

client/ladder/ladder-reducer.ts
```diff
@@ -15,14 +15,33 @@ export interface SearchResults extends RetrievedRankings {
...
+  /** A map of matchmaking type -> rankings. This is meant to be used for the current season. */
   typeToRankings: Map<MatchmakingType, RetrievedRankings>
+  /**
+   * A map of matchmaking type -> search results. This is meant to be used for the current season.
+   */
   typeToSearchResults: Map<MatchmakingType, SearchResults>
+  /**
+   * A map of season ID -> matchmaking type -> rankings. This is meant to be used for past seasons.
+   */
+  seasonToTypeToRankings: Map<SeasonId, Map<MatchmakingType, RetrievedRankings>>
```
tec27, line 32: I would again prefer you convert these keys to strings rather than maintain nested maps

client/ladder/ladder-reducer.ts
```diff
@@ -15,14 +15,33 @@ export interface SearchResults extends RetrievedRankings {
   searchQuery: string
 }
 
+// NOTE(2Pac): The reason why we have two separate maps for rankings and search results is so we
+// we don't have to re-fetch rankings when user clears the search query. Also, we keep the rankings
+// for current season separately from previous seasons because we don't immediately know the current
+// season ID.
 export interface LadderState {
+  /** A map of matchmaking type -> rankings. This is meant to be used for the current season. */
   typeToRankings: Map<MatchmakingType, RetrievedRankings>
```
tec27, line 24: these should probably be renamed to indicate they're for the current season

client/ladder/ladder.tsx
```diff
@@ -120,38 +130,54 @@ export function LadderRouteComponent(props: { params: any }) {
...
   const { t } = useTranslation()
   const dispatch = useAppDispatch()
-  const rankings = useAppSelector(s => s.ladder.typeToRankings.get(matchmakingType))
-  const searchResults = useAppSelector(s => s.ladder.typeToSearchResults.get(matchmakingType))
-  const usersById = useAppSelector(s => s.users.byId)
-
-  const rankingsSeason = useAppSelector(s =>
-    rankings?.seasonId ? s.matchmakingSeasons.byId.get(rankings.seasonId) : undefined,
+  const seasons = useAppSelector(s => s.matchmakingSeasons.byId)
+  const currentSeasonId = useAppSelector(s => s.matchmakingSeasons.currentSeasonId)
```
tec27, line 152: If you have the current season ID here, does the reducer even need a separate non-season-based storage of any of this? Seems useless and makes all the logic after this far more complex when you could just retrieve the same data from the same place always

client/ladder/ladder.tsx
```diff
@@ -183,35 +209,72 @@ export function Ladder({ matchmakingType: routeType }: LadderProps) {
...
+        onSuccess: () => setLastError(undefined),
+        onError: err => setLastError(err),
+      }),
+    )
+  }, [dispatch])
+
   useEffect(() => {
     const getRankingsAbortController = new AbortController()
     const searchRankingsAbortController = new AbortController()
+    const getPastRankingsAbortController = new AbortController()
```
tec27, line 224: This again just seems very weird to me, why are we treating the current season as different from the historical ones?

server/lib/matchmaking/matchmaking-api.ts
```diff
@@ -134,19 +139,27 @@ export class MatchmakingApi {
...
+    throttleMiddleware(matchmakingThrottle, ctx => String(ctx.session!.user!.id)),
+  )
   async acceptMatch(ctx: RouterContext): Promise<void> {
     await this.matchmakingService.accept(ctx.session!.user!.id)
   }
 
   @httpGet('/seasons')
-  @httpBefore(checkAllPermissions('manageMatchmakingSeasons'))
+  @httpBefore(
+    throttleMiddleware(seasonsRetrievalThrottle, ctx => String(ctx.session?.user?.id) ?? ctx.ip),
```
tec27, line 161: can you file an issue to pull these throttle identifier functions out into something common? Copy-pasting these everywhere is not great and it's not entirely clear what their intention is without deeper thought

server/lib/matchmaking/models.ts
```diff
@@ -418,6 +418,52 @@ export async function getManyMatchmakingRatings(
   }
 }
 
+/**
+ * Returns a list of players's finalized matchmaking ranks, and optionally filtered by a search
+ * query, for a particular matchmaking type and season.
+ */
+export async function getManyMatchmakingFinalizedRanks(
```
tec27, line 425: This is a kind of weird naming for this, we typically only use this pattern when we're passing in a bunch of IDs to retrieve, and that's not what this method is doing. I'd probably call this `getFinalizedRanksForSeason` or something

client/ladder/ladder.tsx
2Pacalypse-: Yeah, I was debating this with myself as well. We _don't_ have the current season ID here actually, we fetch it in the effect below which fetches all the matchmaking seasons (and returns the ID of the current season). But you're right, it seems like a whole bunch of stuff would be simplified if we simply fetched the matchmaking seasons (and the current season ID) in the component above this one and show a loader (or nothing at all) until that request is finished?

tec27: I think you could also achieve this without needing to add a waterfall of requests if they're looking at the current season, but yes, I would prefer simplifying all of this code.

I don't think you really need to change anything about the component above this either, if we don't have the current season ID and we are looking at the current season, we know we're in a loading state anyway (one that should be resolved by retrieving the ranks for the current season, as that *also* gives us a season ID and is retrieved from a route we know to be only for the current season, so it can update all this stuff in the reducer anyway).

2Pacalypse-: Man, I haven't looked at this code for a few days and I already forgot most of it 😅. But anyway, after refreshing my memory, I believe there are three main issues that cause this code to be more complex than it needs:
- When the Ladder component gets first rendered, we don't have the `currentSeasonId` yet. And also, unless the user clicked on a particular link which includes the season ID in the route params, the URL won't contain the season ID either. Which means that we can't really write the selector function to get the ranking information from the store (this is why I kept the state for the current season separately, which didn't need the season ID to access it).
- Similarly, I had to do some annoying things with `currentSeasonId` and using it as a ref, because on first render it was `undefined` and it only got initialized once we actually got that info from the backend causing the effects that depended on it to re-run, making additional requests since both of those cases are valid.
- And finally, I had to create separate action creators for retrieving the rankings/search in past seasons versus the current season because there were too many differences created by the fact that they were using different Maps to save the caching time (again caused by the fact that season ID could potentially be `undefined`).

All of these problems would be solved by making sure we always have _a_ season ID in the Ladder component; either it's the one from the route params, or if that's `undefined` we use the `currentSeasonId` that we get from backend when we retrieve all the seasons (to prefill the seasons Select component). But yeah, the downside is that it would add a 1 additional waterfall request, which is mostly the reason I haven't done it and went with this more complex approach.

Btw, one more idea that popped into my head, but wasn't sure if it would work so I haven't tried it yet: what if we still just used `${SeasonId}|${MatchmakingType}` as the Map key for all these things, and just let it be `undefined|1v1` for current season in case we don't have the season ID?

tec27: It's not an additional waterfall request *if* you make the decision about which route to hit for the rankings at the same time you are requesting the season info for the selection control. Because once the rankings come back you could know the current season ID *and* populate the rankings just from that data, and thus both of those things would become defined at that point and the loading state would end (even if the total season info hasn't come back yet).

I don't really care much where those requests happen, I'd just like them not to be chained (they should happen in parallel)

tec27: undefined|1v1 could be fine except that it adds a weird case where we can have the same data for the current season under `undefined|1v1` and e.g. `5|1v1` (but since we request the current season from a different API, maybe we are safe from this occurring). I do think it probably makes the reducer logic between current+past seasons harder to share, though.

tec27: I also understand the need for different action creators, I think this is fine given that they use different API methods on the server as well (and those API methods also hit different data sources, etc.). I think the fact that the action creators are different is actually somewhat helpful here, because it can let you respond to receiving the current season's rankings by also updating the reducer's `currentSeasonId` value (and not do that in the historical season retrieval case)

2Pacalypse-: >undefined|1v1 could be fine except that it adds a weird case where we can have the same data for the current season under undefined|1v1 and e.g. 5|1v1 (but since we request the current season from a different API, maybe we are safe from this occurring). I do think it probably makes the reducer logic between current+past seasons harder to share, though.

Yeah, I thought about this case too. It's technically still possible, because we show the current season in the Select as well, so if the user selects a past season and then goes back to the current one, we'll update the URL with the current season and we'll get duplicated data in the reducer (and also a cache miss in the action creator). I was actually thinking of detecting when the user selects the current season in Select and then navigate to the URL without the season ID in the route params so we re-use the same key (with `undefined` for the season ID). But decided against it since it seemed to complicate things even more than I already had 😅.

But yeah, I can simplify my current approach somewhat by just using the `undefined|MatchmakingType` as the key for the current season, and if we ever get current season ID in the init data on the homepage or something for whatever reason, we can simplify this approach even further.

2Pacalypse-: Ok, only after actually trying to implement this did I finally understand what you were talking about 😅. Namely that we don't have to actually _store_ the rankings in an `undefined|1v1` key in the reducer since we'll always get the season ID in the response. We only need to use the `undefined` for the season ID in the action creator for the cache keys. So the scenario I described in my last message can't really happen and we will not have the duplicated state in the store (we will also not get the cache miss because we will _always_ use the `undefined` value for the current season ID).

Also, just a side note, I've reversed the order in the key so it's `${MatchmakingType}|${SeasonId}` just to be more consistent with our URLs and APIs.

common/matchmaking.ts
```diff
@@ -936,3 +936,21 @@ export interface AddMatchmakingSeasonResponse {
 export interface GetCurrentMatchmakingSeasonResponse {
   season: MatchmakingSeasonJson
 }
+
+export type MatchmakingTypeAndSeasonId = Tagged<
```
tec27, line 940: I don't think you really need this to be tagged? Doesn't the string typing stuff in TS just handle this properly now?

(I would prefer this type doesn't exist here, this is an implementation detail of the things that do this, not something we should be making into a common type we use a ton of different places)

client/ladder/action-creators.ts
```diff
@@ -70,7 +113,57 @@ export function searchRankings(
...
+  seasonId: SeasonId,
+  matchmakingType: MatchmakingType,
+  searchQuery: string,
+  spec: RequestHandlingSpec<void>,
+): ThunkAction {
+  return abortableThunk(spec, async dispatch => {
+    const fetchTime = performance.now()
+    const cacheKey = makeMatchmakingTypeAndSeasonId(matchmakingType, seasonId)
+    const lastSearchTime = lastSearchTimeByMatchmakingTypeAndSeasonId.get(cacheKey)
+    const lastSearchQuery = lastSearchQueryByMatchmakingTypeAndSeasonId.get(cacheKey)
```
tec27, line 144: If search time and query are always set together, they should probably be stored in the same map instead of separately

client/ladder/action-creators.ts
```diff
@@ -96,7 +189,18 @@ export function getInstantaneousSelfRank(spec: RequestHandlingSpec<void>): Thunk
   })
 }
 
-/** Navigates to the ladder standings (optionally for a given matchmaking type). */
-export function navigateToLadder(type?: MatchmakingType, transitionFn = push) {
-  transitionFn(urlPath`/ladder/${type ?? ''}`)
+/** Navigates to the ladder rankings page (optionally for a given matchmaking type and a season). */
+export function navigateToLadder(
+  matchmakingType?: MatchmakingType,
```
tec27, line 194: You could also set a default value here instead of doing the if check below (would result in the URL not being as simple sometimes but I don't think that really matters tbh)

client/ladder/ladder.tsx
```diff
@@ -183,35 +210,75 @@ export function Ladder({ matchmakingType: routeType }: LadderProps) {
...
-        }),
-      )
+    // NOTE(2Pac): Since we want to initiate the requests for retrieving the rankings and retrieving
+    // all of the seasons, including the `currentSeasonId`, at the same time (without having the
+    // waterfall requests), we're using the ref here for the `currentSeasonId` to avoid running this
+    // effect again when the `currentSeasonId` changes (e.g. from `undefined` -> actual value). It
+    // also wouldn't really make sense to run this effect again in case the `currentSeasonId` gets
+    // updated through websockets or something since that would change the rankings the user is
+    // looking at without their input.
+    if (!seasonId || !currentSeasonIdRef.current || seasonId === currentSeasonIdRef.current) {
```
tec27, line 236: can you pull this condition out into a variable like `isForCurrentSeason` or something? Would make it easier to understand what this if statement is trying to do

client/matchmaking/action-creators.ts
```diff
@@ -236,6 +237,7 @@ export function updateLastQueuedMatchmakingType(
   }
 }
 
+// TODO(2Pac): I Don't think this is used anymore, remove it?
```
tec27, line 240: Can probably figure that out before this PR gets merged :)

tec27 (changes requested):

common/matchmaking.ts
2Pacalypse-: Hmm, if I define a map like this
```ts
const map = new Map<`${MatchmakingType}|${SeasonId}`, number>()
```

And then try to use it like this:
```ts
const key = `${matchmakingType}|${seasonId}`
const value = map.get(key)
```

I get the following error:

image

But if I use it like this:
```ts
const value = map.get(`${matchmakingType}|${seasonId}`)
```

Then it works fine. Weird.

tec27: Would guess this is just a limitation of type inference, and if you explicitly defined the type of `key` it would work the other way too. Or just wait a few typescript versions and maybe they'll fix this :P

client/ladder/action-creators.ts
```diff
@@ -13,17 +8,21 @@ import { fetchJson } from '../network/fetch'
...
+>()
 
 export function getCurrentSeasonRankings(
   matchmakingType: MatchmakingType,
   spec: RequestHandlingSpec<void>,
 ): ThunkAction {
   return abortableThunk(spec, async dispatch => {
     const fetchTime = performance.now()
-    const cacheKey = makeMatchmakingTypeAndSeasonId(matchmakingType)
+    const cacheKey = `${matchmakingType}|undefined` as MatchmakingTypeAndSeasonId
```
tec27, line 25: Does this work if you do `cacheKey: MatchmakingTypeAndSeasonId = ...` instead? Would prefer that, as I worry the cast may not actually require the string to be in that form. (Apply throughout if so)

tec27 (approved):

PR #1131, opened 2025-02-07
Move the `chatReady` and `whispersReady` socket events to an HTTP

### LINKED ISSUE #1126: Move `chatReady` and `whispersReady` socket events to an HTTP request that happens on load instead
These are a remainder of the loading filter stuff we did initially but don't really use any more, and it'd be nice to move the rest of the large-ish data loads to HTTP instead of being done through sockets on connect

2Pacalypse-:
request that happens on load instead.

This also simplifies a little bit the way we're sending the joined
channels data (previously we had an unnecessary DB call).

We also had to send the active users data separately from the init data
for the channel, since now the HTTP request that returns the joined
channel info might run *before* we update the state for active users.

Also, this refactors the whisper service a little bit so that `init`
data is not sent when the user connects. Now it's only sent when a new
whisper session is started. This is more inline with how chat service
works as well, since we had a bunch of issues with channels ordering due
to that there.

Closes #1126


tec27:
This will need to be redone after the new design stuff is merged, there's too much stuff that has changed that this work is based off of

client/whispers/route.tsx
```diff
@@ -20,5 +20,5 @@ export function WhisperRouteComponent(props: { params: any }) {
     return null
   }
```
tec27, line 22: Can you remove the `queueMicrotask` piece above this? (Just call `replace('/')` directly, not inside a microtask)

I had to deal with this in chat too where it caused weird issues, guess I didn't see that whispers was doing it too

common/chat.ts
```diff
@@ -296,6 +288,12 @@ export interface ChatMessageDeletedEvent {
   messageId: string
 }
 
+export interface ChatInitActiveUsersEvent {
+  action: 'initActiveUsers'
+  /** A list of IDs of active users that are in a chat channel. */
```
tec27, line 293: nit: `in the chat channel`

server/lib/chat/chat-service.ts
```diff
@@ -161,7 +181,10 @@ export default class ChatService {
     // is allowed in some cases (e.g. during account creation)
     const userSockets = this.userSocketsManager.getById(userInfo.id)
     if (userSockets) {
-      userSockets.subscribe(getChannelPath(channelId))
+      userSockets.subscribe<ChatInitActiveUsersEvent>(getChannelPath(channelId), () => ({
+        action: 'initActiveUsers',
+        activeUserIds: this.state.channels.get(channelId)!.toArray(),
```
tec27, line 186: For safety can you please make this handle the undefined case instead of doing the non-null coercion? `this.state.channels.get(channelId)?.toArray() ?? []`

server/lib/chat/chat-service.ts
```diff
@@ -1128,45 +1150,15 @@ export default class ChatService {
         action: 'userActive2',
         userId: userSockets.userId,
       })
-      userSockets.subscribe(getChannelPath(userChannel.channelId))
+      userSockets.subscribe<ChatInitActiveUsersEvent>(
+        getChannelPath(userChannel.channelId),
+        () => ({
+          action: 'initActiveUsers',
+          activeUserIds: this.state.channels.get(userChannel.channelId)!.toArray(),
```
tec27, line 1157: Same as above

tec27: And perhaps these 2 subscribes should be pulled out into a common method or something so this code isn't duplicated?

server/lib/whispers/whisper-api.ts
```diff
@@ -68,6 +74,12 @@ export class WhisperApi {
     container.resolve(WhisperService)
   }
 
+  @httpGet('/sessions')
+  @httpBefore(throttleMiddleware(getWhisperSessionsThrottle, ctx => String(ctx.session!.user!.id)))
+  async getWhisperSessions(ctx: RouterContext): Promise<SbUserId[]> {
```
tec27, line 79: I think we should have learned by this point that you should basically never return a raw array from API methods, always put it within an object so we can add new fields when necessary.

Additionally, this should almost certainly return the info for the users they have sessions with so that it doesn't have to be retrieved separately.

tec27 (changes requested):

client/whispers/route.tsx
2Pacalypse-: Should it be removed from here as well?
https://github.com/ShieldBattery/ShieldBattery/blob/7ee9a4fac8bd8a67bbfbe40efa885b3f1f356f55/client/chat/channel-route.tsx#L17-L19

client/social/social-sidebar.tsx
```diff
@@ -332,10 +342,11 @@ function ChatContent() {
         signal: abortController.signal,
         onSuccess: () => {
           setIsLoadingJoinedChannels(false)
+          setLoadingJoinedChannelsError(undefined)
         },
-        onError: () => {
+        onError: err => {
           setIsLoadingJoinedChannels(false)
-          // TODO(2Pac): Do something with the error?
+          setLoadingJoinedChannelsError(err)
```
tec27, line 349: log this too

client/social/social-sidebar.tsx
```diff
@@ -355,10 +366,11 @@ function ChatContent() {
         signal: abortController.signal,
         onSuccess: () => {
           setIsLoadingWhisperSessions(false)
+          setLoadingWhisperSessionsError(undefined)
         },
-        onError: () => {
+        onError: err => {
           setIsLoadingWhisperSessions(false)
-          // TODO(2Pac): Do something with the error?
+          setLoadingWhisperSessionsError(err)
```
tec27, line 373: log this too

client/whispers/whisper-reducer.ts
```diff
@@ -68,9 +68,9 @@ function updateMessages(
 
 export default immerKeyedReducer(DEFAULT_STATE, {
   ['@whispers/getWhisperSessions'](state, action) {
-    state.sessions = new Set(action.payload)
+    state.sessions = new Set(action.payload.sessions)
 
-    for (const session of action.payload) {
+    for (const session of action.payload.sessions) {
```
tec27, line 73: might need to merge these instead of assuming they are completely unset?

tec27 (approved):

client/whispers/whisper-reducer.ts
2Pacalypse-: You mean something like this?
```ts
for (const session of action.payload.sessions) {
  if (state.byId.has(session)) {
    continue
  }

  state.byId.set(session, defaultWhisperSession(session))
}
```

tec27: sure

PR #1132, opened 2025-02-09
Fix wrap issue

### LINKED ISSUE #885: Flex-wrap issue in mini-match-results
image

GrantJenk:
Fixes #885

Preview: 
Screenshot From 2025-02-08 20-20-45

In the issues thread, 2pac suggested CSS Grid may be the way to do this, but I figured just fixing the placeholders would be less effort


2Pacalypse- (approved):
Thx!

PR #1155, opened 2025-04-13
Refactor the matchmaking times feature to use the new patterns.

2Pacalypse-:
Also simplifies how the pagination is done a bit, and adds slight changes to the UI.

Other than that, the feature remains mostly unchanged in its core functionality.

The next PR/commit will add to this work by making it more amenable to support many more matchmaking types.


client/matchmaking/admin-matchmaking-times.tsx
```diff
@@ -0,0 +1,658 @@
...
+    const result = await fetchJson<MatchmakingTimeJson>(
+      `/api/1/matchmaking-times/${encodeURIComponent(type)}`,
+      {
+        method: 'post',
+        body: encodeBodyAsParams<AddMatchmakingTimeRequest>({ startDate, enabled }),
+        signal: spec.signal,
+      },
+    )
+
+    externalShowSnackbar('New matchmaking time created')
```
tec27, line 94: You have a place to show this snackbar that is inside the component tree, so please don't use this method. Apply throughout

client/matchmaking/admin-matchmaking-times.tsx
```diff
@@ -0,0 +1,658 @@
...
+import { encodeBodyAsParams, fetchJson } from '../network/fetch'
+import { LoadingDotsArea } from '../progress/dots'
+import { useImmerState } from '../react/state-hooks'
+import { useAppDispatch } from '../redux-hooks'
+import { externalShowSnackbar } from '../snackbars/snackbar-controller-registry'
+import { CenteredContentContainer } from '../styles/centered-container'
+import { styledWithAttrs } from '../styles/styled-with-attrs'
+import { bodyLarge, bodyMedium, titleLarge } from '../styles/typography'
+
+function getMatchmakingTimesHistory(
```
tec27, line 34: This naming seems weird since we also have `getMatchmakingTimesPast`, maybe it should just be called `getMatchmakingTimes`?

tec27 (changes requested):
Would generally prefer that you rethink the client components instead of just inlining the old reducer into the new component. This can be a lot simpler if you divide up the component a bit more

client/matchmaking/admin-matchmaking-times.tsx
```diff
@@ -0,0 +1,658 @@
...
+
+  const [activeTab, setActiveTab] = useState<MatchmakingType>(MatchmakingType.Match1v1)
+
+  const [startDate, setStartDate] = useState('')
+  const [invalidDate, setInvalidDate] = useState(false)
+  const [enabled, setEnabled] = useState(false)
+
+  const [addError, setAddError] = useState<Error>()
+
+  const [matchmakingTimesState, setMatchmakingTimesState] = useImmerState<MatchmakingTimesState>({
```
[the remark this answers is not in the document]
tec27: I don't get why you need to combine these things at all? Even if you wanted them in the same Card without a wrapper element, you can easily just return a JSX fragment with all the things you want? But I don't see why you even need/want that, you can just make a historical times card and a current time card and a future times card or whatever. Adding a new time indeed updates the future state, but that just means only the component dealing with future times needs to deal with that? We have APIs for retrieving future/past/current that are all separate, why not just delete the one that returns all 3 and make 3 separate requests, one per component? What is the benefit of having the combined request here?

If you ever have 10 different state variables that you're passing down to the same component, you should probably reconsider how you've organized the code/responsibilities.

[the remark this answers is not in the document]
2Pacalypse-: Well, I didn't know rewriting APIs was allowed :P

client/matchmaking/admin-matchmaking-times.tsx
```diff
@@ -0,0 +1,690 @@
...
+
+function addMatchmakingTime(
+  type: MatchmakingType,
+  startDate: number,
+  enabled: boolean,
+  spec: RequestHandlingSpec<MatchmakingTimeJson>,
+): ThunkAction {
+  return abortableThunk(spec, async dispatch => {
+    return await fetchJson<MatchmakingTimeJson>(
+      `/api/1/matchmaking-times/${encodeURIComponent(type)}`,
```
tec27, line 85: use `apiUrl`

client/matchmaking/admin-matchmaking-times.tsx
```diff
@@ -0,0 +1,690 @@
...
+  })
+}
+
+function deleteMatchmakingTime(
+  type: MatchmakingType,
+  id: number,
+  spec: RequestHandlingSpec,
+): ThunkAction {
+  return abortableThunk(spec, async dispatch => {
+    await fetchJson(`/api/1/matchmaking-times/${encodeURIComponent(id)}`, {
```
tec27, line 101: use `apiUrl`

client/matchmaking/admin-matchmaking-times.tsx
```diff
@@ -0,0 +1,690 @@
...
+import { styledWithAttrs } from '../styles/styled-with-attrs'
+import { bodyLarge, bodyMedium, titleLarge } from '../styles/typography'
+
+function getCurrentMatchmakingTime(
+  type: MatchmakingType,
+  spec: RequestHandlingSpec<MatchmakingTimeJson | undefined>,
+): ThunkAction {
+  return abortableThunk(spec, async () => {
+    return await fetchJson<MatchmakingTimeJson | undefined>(
+      apiUrl`matchmaking-times/${encodeURIComponent(type)}/current`,
```
tec27, line 39: `apiUrl` already encodes, you can remove those calls provided you're using it

client/matchmaking/admin-matchmaking-times.tsx
```diff
@@ -0,0 +1,690 @@
...
+  ${bodyLarge};
+  color: var(--theme-error);
+`
+
+const AddNewCard = styled.div`
+  display: flex;
+  flex-direction: column;
+  gap: 8px;
+  padding: 16px;
+  background-color: rgba(255, 255, 255, 0.03);
```
tec27, line 129: nit: use `containerStyles`

client/matchmaking/admin-matchmaking-times.tsx
```diff
@@ -0,0 +1,690 @@
...
+  padding: 8px 12px;
+  border-radius: 4px;
+  border: 1px solid rgba(255, 255, 255, 0.1);
+  background-color: rgba(255, 255, 255, 0.05);
+  font-size: 14px;
+  min-width: 200px;
+
+  &:focus {
+    outline: none;
+    border-color: rgba(255, 255, 255, 0.2);
```
tec27, line 151: I know visuals don't really matter as much in admin tools but it would be nice to use the common CSS variable stuff, I have a todo thing to clean up all the places we do `rgba(255, 255, 255, ###)` because it looks much worse than using a specific shade of blue/grey-blue, and it would be nice to not have to come clean up this file that you just created

client/matchmaking/admin-matchmaking-times.tsx
```diff
@@ -0,0 +1,690 @@
...
+  text-transform: uppercase;
+  letter-spacing: 0.5px;
+`
+
+const HistoryCard = styled.div`
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  padding: 16px;
+  background-color: rgba(255, 255, 255, 0.05);
```
tec27, line 207: `containerStyles` here too

client/matchmaking/admin-matchmaking-times.tsx
```diff
@@ -0,0 +1,690 @@
...
+  border-radius: 8px;
+  transition: background-color 0.2s ease;
+
+  &:hover {
+    background-color: rgba(255, 255, 255, 0.08);
+  }
+`
+
+const CurrentMatchmakingCard = styled(HistoryCard)`
+  background-color: rgba(255, 193, 7, 0.1);
```
tec27, line 217: I don't know what these colors are but please use stuff from the theme and/or `containerStyles`

client/matchmaking/admin-matchmaking-times.tsx
```diff
@@ -0,0 +1,690 @@
...
+  color: var(--theme-on-surface-variant);
+`
+
+const EmptyStateContainer = styled.div`
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  justify-content: center;
+  padding: 32px;
+  background-color: rgba(255, 255, 255, 0.03);
```
tec27, line 275: `containerStyles`

client/matchmaking/admin-matchmaking-times.tsx
```diff
@@ -0,0 +1,690 @@
...
+  const dispatch = useAppDispatch()
+  const snackbarController = useSnackbarController()
+
+  const [startDate, setStartDate] = useState('')
+  const [invalidDate, setInvalidDate] = useState(false)
+  const [enabled, setEnabled] = useState(false)
+
+  const [addError, setAddError] = useState<Error>()
+
+  const [futureTimesIds, setFutureTimesIds] = useImmerState<Set<number>>(new Set())
```
tec27, line 326: I think this is overcomplicating things, just keep an array of the objects in a `useState`? What does a Set/Map provide us here? We're just iterating them to render anyway. Iterating 10s of items to find the right one to delete is straightforward/fast

client/matchmaking/admin-matchmaking-times.tsx
```diff
@@ -0,0 +1,690 @@
...
+
+function AddAndDisplayMatchmakingTimes({ activeTab }: { activeTab: MatchmakingType }) {
+  const dispatch = useAppDispatch()
+  const snackbarController = useSnackbarController()
+
+  const [startDate, setStartDate] = useState('')
+  const [invalidDate, setInvalidDate] = useState(false)
+  const [enabled, setEnabled] = useState(false)
+
+  const [addError, setAddError] = useState<Error>()
```
tec27, line 324: `addError` is a very strange name for this, makes it seem like this is a function for adding errors

client/matchmaking/admin-matchmaking-times.tsx
```diff
@@ -0,0 +1,690 @@
...
+
+  const [futureTimesIds, setFutureTimesIds] = useImmerState<Set<number>>(new Set())
+  const [futureTimesById, setFutureTimesById] = useImmerState<Map<number, MatchmakingTimeJson>>(
+    new Map(),
+  )
+
+  const handleStartDateChange = (event: React.ChangeEvent<HTMLInputElement>) => {
```
tec27, line 331: inline, then you won't need the parameter type

client/matchmaking/admin-matchmaking-times.tsx
```diff
@@ -0,0 +1,690 @@
...
+        </HistoryInfo>
+      </CurrentMatchmakingCard>
+    </>
+  )
+}
+
+function PastMatchmakingTimes({ activeTab }: { activeTab: MatchmakingType }) {
+  const dispatch = useAppDispatch()
+
+  const [pastTimesIds, setPastTimesIds] = useImmerState<Set<number>>(new Set())
```
tec27, line 604: Same as for the future ones, don't really see the value of this here

client/matchmaking/admin-matchmaking-times.tsx
```diff
@@ -0,0 +1,690 @@
...
+  }
+
+  if (pastTimesIds.size === 0) {
+    return null
+  }
+
+  return (
+    <>
+      <SectionTitle>Past Matchmaking Times</SectionTitle>
+      {Array.from(pastTimesIds).map(id => {
```
tec27, line 661: `Array.from`'s second parameter is a `map` function, use it instead of calling `map`  after

client/matchmaking/admin-matchmaking-times.tsx
```diff
@@ -0,0 +1,690 @@
...
+      {hasMoreFutureTimes && (
+        <LoadMoreButton
+          label={isLoadingFutureTimes ? 'Loading...' : 'Load more upcoming times'}
+          color='accent'
+          disabled={isLoadingFutureTimes}
+          onClick={() => onLoadMoreFutureTimes(futureTimesIds.size)}
+        />
+      )}
+      <FutureTimesContainer>
+        {Array.from(futureTimesIds).map(id => {
```
tec27, line 509: Array.from's second parameter is a map function, use it instead of calling map after

server/lib/matchmaking/matchmaking-times-api.ts
```diff
@@ -0,0 +1,147 @@
...
+    return current ? toMatchmakingTimeJson(current) : undefined
+  }
+
+  @httpGet('/:matchmakingType/future')
+  async getFutureTimes(ctx: RouterContext): Promise<GetFutureMatchmakingTimesResponse> {
+    const matchmakingType = getValidatedMatchmakingType(ctx)
+    const {
+      query: { offset },
+    } = validateRequest(ctx, {
+      query: Joi.object<{ offset: number }>({
```
tec27, line 59: This typing isn't correct for what you've written here, you never made the parameter required. If it is required, mark it as such. If it isn't, mark it optional in this typing

server/lib/matchmaking/matchmaking-times-api.ts
```diff
@@ -0,0 +1,147 @@
...
+    }
+  }
+
+  @httpGet('/:matchmakingType/past')
+  async getPastTimes(ctx: RouterContext): Promise<GetPastMatchmakingTimesResponse> {
+    const matchmakingType = getValidatedMatchmakingType(ctx)
+    const {
+      query: { offset },
+    } = validateRequest(ctx, {
+      query: Joi.object<{ offset: number }>({
```
tec27, line 85: Same as above, optional or `required()`

tec27 (changes requested):

client/matchmaking/admin-matchmaking-times.tsx
2Pacalypse-: Actually I didn't do this primarily for the performance. But rather, to get around the annoying edge case of sometimes possibly having duplicates in the future times (e.g. we insert a new time into the array when adding it, and then also insert it when loading a new page if the added time is in that page). We were handling this manually previously (also by utilizing Sets), but thought separating the states like this makes it a bit "cleaner".
https://github.com/ShieldBattery/ShieldBattery/blob/2e120c81127c0d2b171ccc3ffad7d961aadad936/client/admin/matchmaking-times-reducer.js#L89-L90


I guess I could do the same thing again and just use a Set of objects? Or use an array and just do `Array.from(new Set(...))` to deduplicate?

client/matchmaking/admin-matchmaking-times.tsx
2Pacalypse-: I was moving the past times to an array since we don't have the same issue we have with future times here, but I've noticed another issue with using arrays or sets of objects. Which is that I get duplicated first page of past times on initial load, since the effect that's being used to load the first page of past times is being called twice in strict mode (only on dev I guess).

Now I know that React does this to surface any bugs in effects caused by calling the effect twice, but I really don't know how we're supposed to do this very basic functionality of _load some data when component mounts_ otherwise?

tec27: Strict mode does this to surface bugs that happen when effects get called twice *which can happen in production*. There is no concept of "effects that only run on mount", all effects can run multiple times. If your code assumes it will only run on mount ever, the code is broken

client/matchmaking/admin-matchmaking-times.tsx
tec27: Probably just:
```ts
const seen = new Set()
return existingTimes.concat(futureTimes).filter(t => {
  const result = seen.has(t.id)
  seen.add(t.id)
  return result
})
```

but this isn't really correct either (and your old code also wasn't). If you retrieve an ordering from the server then that ordering should be assumed to be the correct one, not whatever you had locally. You can of course just sort by date or whatever in t his case, but yeah.

tec27 (changes requested):

client/matchmaking/admin-matchmaking-times.tsx
```diff
@@ -275,46 +268,110 @@ export function AdminMatchmakingTimes() {
...
+  const [addingTimeError, setAddingTimeError] = useState<Error>()
 
   const [futureTimes, setFutureTimes] = useState<MatchmakingTimeJson[]>([])
+  const [hasMoreFutureTimes, setHasMoreFutureTimes] = useState(true)
+  const [isLoadingFutureTimes, setIsLoadingFutureTimes] = useState(false)
+  const [loadingFutureTimesError, setLoadingFutureTimesError] = useState<Error | undefined>(
+    undefined,
+  )
+
+  const futureTimesRef = useValueAsRef(futureTimes)
```
tec27, line 300: I dunno if this is deprecated or not but it should be, so please stop writing new code using it. To accomplish what you're doing here without it, just use the callback form of the state setter. Apply throughout

common/data-structures/arrays.ts
```diff
@@ -52,17 +52,24 @@ export function binarySearch<T, V>(
...
+ * @param array2 The second array that will be appended to the first one.
+ * @param keyFn A function that returns the value based on which the uniqueness will be checked.
+ *   Defaults to the identity function (to support arrays of non-objects).
  */
-export function concatWithoutDuplicates<T extends Record<string, unknown>>(
+export function concatWithoutDuplicates<T, U>(
   array1: ReadonlyArray<T>,
   array2: ReadonlyArray<T>,
-  key: keyof T,
+  keyFn: (value: T) => U | T = (value: T) => value,
```
tec27, line 67: This is not correct, you don't want a function that can return a `U` or `T`, you want one that can return a `U`. By doing it this way you make it possible to accidentally write a function that returns the wrong thing. In the case that your default method works, `U = T`. Make the `keyFn` optional (with no default value in the parameter list) and assign a variable in the function, i.e. `const getKey = keyFn ?? identity`

common/data-structures/arrays.ts
```diff
@@ -52,17 +52,24 @@ export function binarySearch<T, V>(
...
 ): Array<T> {
   const seen = new Set()
   return array1.concat(array2).filter(t => {
-    const result = !seen.has(t[key])
-    seen.add(t[key])
+    const result = !seen.has(keyFn(t))
```
tec27, line 71: only call `keyFn` on `t` once here, store the value. We don't know how expensive getting the key will be

tec27 (approved):

PR #1157, opened 2025-04-20
Make sure that the content width doesn't extend beyond viewport width.

2Pacalypse-:


tec27:
Done a different way in a5fc5bc40282a3c52b6f2da0fa45b88badc20c3c

PR #1158, opened 2025-04-21
Add a little bit of vertical padding to usernames in chat messages.

2Pacalypse-:
This should make it less likely for right-clicking the username to accidentally pop up chat message context menu if you were a couple of pixels off.

### DIFF
--- a/client/messaging/common-message-layout.tsx
+++ b/client/messaging/common-message-layout.tsx
@@ -33,7 +33,7 @@ const Username = styled(ConnectedUsername)`
   ${titleSmall};
 
   margin-right: 8px;
-  padding: 4px 0;
+  padding-block: 4px;
 
   color: var(--color-amber90);
   line-height: inherit;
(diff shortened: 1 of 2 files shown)


client/messaging/common-message-layout.tsx
```diff
@@ -29,10 +29,11 @@ const newDayFormat = new Intl.DateTimeFormat(navigator.language, {
   day: '2-digit',
 })
 
-const Username = styled.span`
+const Username = styled(ConnectedUsername)`
   ${titleSmall};
 
   margin-right: 8px;
+  padding: 4px 0;
```
tec27, line 36: nit: use `padding-block` for this

I feel like `padding-block`/`padding-inline` are a better choice for stuff like this because it doesn't override the value we don't care about (so it's more resilient to changes in the thing we're modifying)

tec27 (approved):

PR #1162, opened 2025-05-05
Open a popover for autocompleting user mentions when @ is typed in chat.

2Pacalypse-:


client/messaging/message-input.tsx
```diff
@@ -148,19 +277,75 @@ export const MessageInput = React.forwardRef<MessageInputHandle, MessageInputPro
...
+          message.slice(0, userMentionStartIndex) +
+            `@${user.name} ` +
+            message.slice(userMentionStartIndex + userMentionMatchedText.length),
+        )
+      }
+
+      inputRef.current?.focus()
+      // Setting the caret position immediately after the focus doesn't work for some reason, so we
+      // need to wait a tick first.
+      Promise.resolve()
```
tec27, line 294: Use `queueMicrotask`?

client/messaging/message-input.tsx
```diff
@@ -148,19 +277,75 @@ export const MessageInput = React.forwardRef<MessageInputHandle, MessageInputPro
...
+        .then(() => {
+          const newCaretPosition = userMentionStartIndex + user.name.length + 2
+          inputRef.current?.setSelectionRange(newCaretPosition, newCaretPosition)
+        })
+        .catch(err => logger.warning(`Error while setting caret position: ${getErrorStack(err)}`))
```
tec27, line 299: I don't think this can fail anyway? Just `swallowNonBuiltins`

client/messaging/message-input.tsx
tec27: There's also no point in doing this if `inputRef.current` isn't set

client/messaging/message-input.tsx
```diff
@@ -148,19 +277,75 @@ export const MessageInput = React.forwardRef<MessageInputHandle, MessageInputPro
       }),
     })
 
+    const onUserMentionSelect = useStableCallback((user: SbUser) => {
```
tec27, line 280: Stop writing new `useCallback`/`useStableCallback`. The only time to `useCallback` now is if you get the lint warning about it making a `useMemo` useless (and hopefully React fixes that warning soon)

tec27: Also inline this

client/messaging/message-input.tsx
```diff
@@ -69,6 +91,17 @@ export interface MessageInputProps {
...
+   * An optional list of users that can be mentioned in the message input. If provided, the message
+   * input will display a popover with all matching users when the user starts typing something
+   * *after* the @ character and there's a match.
+   */
+  mentionableUsers?: SbUser[]
+  /**
+   * Similar to the `mentionableUsers` property above, except this list will be used when the user
+   * has only typed the @ character and nothing else after it.
+   */
+  defaultMentionableUsers?: SbUser[]
```
tec27, line 104: I think this naming is somewhat weird since `defaultFoo` is usually for uncontrolled values. Come up with something else

client/dom/focus-trap.tsx
```diff
@@ -11,30 +10,35 @@ export interface FocusTrapProps {
    * keyboard-navigable, this can be `-1`, and should probably be `0` otherwise.
    */
   focusableRef: React.RefObject<HTMLElement | null>
+  /**
+   * Whether the focus trap should steal the focus when it is mounted.
```
tec27, line 14: Mention the default

tec27 (changes requested):

client/messaging/message-input.tsx
```diff
@@ -107,12 +160,88 @@ export const MessageInput = React.forwardRef<MessageInputHandle, MessageInputPro
...
+
+          // TODO(2Pac): Handle channel mentions as well.
+
+          if (mentionableUsers) {
+            // Special case the single @ character (Discord displays last 10 people who posted in
+            // chat here).
+            const singleAtCharacterIndex = message.slice(0, selectionStart).search(/(?<=^|\s)@$/)
+            if (
+              singleAtCharacterIndex !== -1 &&
+              defaultMentionableUsers &&
```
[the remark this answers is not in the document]
2Pacalypse-: This actually is not allowed by TypeScript it seems:

image

client/messaging/message-input.tsx
```diff
@@ -107,12 +160,88 @@ export const MessageInput = React.forwardRef<MessageInputHandle, MessageInputPro
...
+          if (selectionStart === null || selectionStart !== selectionEnd) {
+            return
+          }
+
+          // TODO(2Pac): Handle channel mentions as well.
+
+          if (mentionableUsers) {
+            // Special case the single @ character (Discord displays last 10 people who posted in
+            // chat here).
+            const singleAtCharacterIndex = message.slice(0, selectionStart).search(/(?<=^|\s)@$/)
```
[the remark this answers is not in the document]
2Pacalypse-: Hmm I actually need to get the index here so I can insert the selected user at that place. Searching by regex also allows me to cover two cases I need to check succinctly (whether the @ is at the beginning of the message, or somewhere after a whitespace).

client/messaging/message-input.tsx
```diff
@@ -107,12 +160,88 @@ export const MessageInput = React.forwardRef<MessageInputHandle, MessageInputPro
...
+              openUserMentions(event)
+            } else {
+              closeUserMentions()
+            }
+          }
+        }
+      }
+
+      const inputRefCopy = inputRef.current
+      inputRefCopy?.addEventListener('selectionchange', onSelectionChange)
```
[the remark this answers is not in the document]
2Pacalypse-: For some reason React doesn't support `onSelectionChange` event yet, only `onSelect` which works differently:
https://github.com/facebook/react/issues/5785

client/messaging/message-input.tsx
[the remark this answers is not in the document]
tec27: remove the `> 0` part?

client/messaging/message-input.tsx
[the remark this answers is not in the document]
tec27: You already know all of these things easily though? It's the last 2 characters of the message, so
1) It's at length - 1
2) If the first character of the last 2 characters is a whitespace character, it's not at the beginning of the string, otherwise it is.

Please avoid running a regex over 100s of characters for no reason, especially when it happens on every character typed or selection change

client/users/sorted-user-ids.ts
```diff
@@ -35,7 +35,6 @@ export function useUserEntriesSelector(
         id,
         state.users.byId.get(id)?.name,
       ])
-      result.sort((a, b) => a[0] - b[0])
```
tec27, line 38: We were sorting these because `areUserEntriesEqual` expects a consistent/stable ordering or it doesn't work. If you need some special behavior for your implementation I would prefer that you just leave this intact and write something for your one-off

tec27 (changes requested):

client/users/sorted-user-ids.ts
tec27: I honestly don't really get why you're passing user entries down at all, you should just be passing user IDs and the selection popover can connect to the store to get the username?

2Pacalypse-: > We were sorting these because areUserEntriesEqual expects a consistent/stable ordering or it doesn't work.

Ah that's right, completely forgot about that. Thought it was an unnecessary sort since we're sorting the users later again separately.

> I honestly don't really get why you're passing user entries down at all, you should just be passing user IDs and the selection popover can connect to the store to get the username?

Yeah, I tried doing something like this initially, but then I realized I actually need usernames in the message-input too, since that's where I'm doing the fuzzy matching. I thought about connecting the message-input to the store as well, but seemed kinda cleaner to just have the chat component where message-input is used to just pass down the user entries. Especially since I would still need to use the same user entries selectors to avoid unnecessary re-renders.

client/users/friends-list.tsx
```diff
@@ -327,36 +327,42 @@ function VirtualizedFriendRequestsList({ height }: { height: number }) {
     useUserEntriesSelector(outgoingRequests),
     areUserEntriesEqual,
   )
-  const sortedIncoming = useMemo(() => sortUserEntries(incomingUserEntries), [incomingUserEntries])
-  const sortedOutgoing = useMemo(() => sortUserEntries(outgoingUserEntries), [outgoingUserEntries])
+  const incomingUserIds = useMemo(
```
tec27, line 330: Why do this work at all? You're already mapping over this resulting array anyway, why not just make that map work on the entries instead of IDs?

client/users/user-entries.ts
```diff
@@ -10,6 +10,10 @@ export type UserEntry = [userId: SbUserId, username: string | undefined]
  * Returns a function for use with `useAppSelector` that maps the given `userIds` to their
  * associated names. Should be used with `areUserEntriesEqual` when passed to `useAppSelector`.
  *
+ * The user entries are sorted because the `areUserEntriesEqual` equality function this selector is
```
tec27, line 13: Just say that this function returns the user entries sorted alphabetically, and that the sort is stable.

tec27 (changes requested):

Issue #1161, opened 2025-05-04
Keep the consistent order of teams/players when displaying them in user's match history


2Pacalypse-:
Currently the teams in user's match history are displayed like this:

![Image](https://github.com/user-attachments/assets/c4c25e4b-20a2-4337-91da-351ea89b361f)

As you can see, the order of teams/players can be a bit random because I think it depends on how the players got slotted into when creating the game?

It would be nice if we could keep this order consistent and predictable, and I believe we can achieve this by doing just two things:
- Put the team/player of the user whos profile we're currently viewing on the left side
- Order the players within team alphabetically

tec27:
The order is based on how slots were randomized, not anything to do with creating the game/the lobby (not that it matters that much for doing anything about this issue :) )

2Pacalypse-:
Fixed in https://github.com/ShieldBattery/ShieldBattery/commit/67e62bd5433fd6d11b553022410322db52b205f1

PR #1167, opened 2025-05-16
Refactor the game players display to use flexbox instead of CSS columns.

### LINKED ISSUE #1166: Fix player alignment in match history
Players in match history are aligned weirdly because they always have bottom margin regardless of if there is anything underneath them. This means that when they get "vertically centered" they're actually "vertically centered if they had an extra 8px at the bottom".

I don't know what layout type they're using but probably this should just be using `gap` or something?

Image

I think it may be reusing the stuff from the mini-match-history, but mini-match-history doesn't ever try to align these elements relative to any other elements vertically, so the extra margin doesn't matter there

2Pacalypse-:
This has the benefit of being able to use gaps instead of margins, and also not having to insert fake elements to keep the teams ordered.

This commit also changes the order in which the teams/players are ordered so they always keep the consistent order. Basically the ordering follows two rules:
- Put the team/player of the user whose profile we're currently viewing on the left side
- Order the players within team alphabetically

Fixes #1166, #1161

### DIFF
--- a/client/games/game-players-display.tsx
+++ b/client/games/game-players-display.tsx
@@ -20,6 +20,7 @@ const GamePreviewPlayers = styled.div`
 
 const GamePreviewTeam = styled.div`
   min-width: 0;
+  width: calc(50% - 8px);
 
   display: flex;
   flex-direction: column;
--- a/client/games/game-players-display.tsx
+++ b/client/games/game-players-display.tsx
@@ -3,18 +3,27 @@ import React, { useCallback, useMemo } from 'react'
 import { useTranslation } from 'react-i18next'
 import styled from 'styled-components'
 import { ReadonlyDeep } from 'type-fest'
+import { GameConfigPlayer } from '../../common/games/configuration'
 import { GameRecordJson } from '../../common/games/games'
 import { RaceChar } from '../../common/races'
 import { SbUser } from '../../common/users/sb-user'
+import { SbUserId } from '../../common/users/sb-user-id'
 import { RaceIcon } from '../lobbies/race-icon'
 import { useAppSelector } from '../redux-hooks'
 import { RootState } from '../root-reducer'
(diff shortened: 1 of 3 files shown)


client/games/game-players-display.tsx
```diff
@@ -3,38 +3,46 @@ import React, { useCallback, useMemo } from 'react'
...
 import { RaceIcon } from '../lobbies/race-icon'
 import { useAppSelector } from '../redux-hooks'
 import { RootState } from '../root-reducer'
 import { labelMedium, singleLine, titleSmall } from '../styles/typography'
 
 const GamePreviewPlayers = styled.div`
-  column-count: 2;
-  column-gap: 16px;
-  text-align: start;
+  display: flex;
```
tec27, line 17: I think the flex change is fine, but FWIW the things you "fixed" by moving to flex work fine with grid too. You could have used `row-gap` instead of the bottom margin, and could have explicitly assigned the players to specific columns instead of adding blank elements

tec27 (approved):

client/games/game-players-display.tsx
2Pacalypse-: I actually did try using grid for this initially, while all the players were in a single array, and I couldn't get it to fill columns first. Which I later learned is not possible with grids :d:
https://stackoverflow.com/questions/44092529/make-grid-container-fill-columns-not-rows

I did think that it actually might be possible to do that with pre-defined grid areas and then assigning players to each grid area, but honestly, using flexbox seemed more straightforward to me and a bit easier to reason about to have a separate element for each team.

If you want, I can try reworking this with grids.

tec27: Yes, I was talking about either using predfined grid areas with 1 area per column, or explicitly setting the 1-column ranges on the player elements. I don't really care for this particular PR, you can leave it as-is, I was just saying for future reference

PR #1172, opened 2025-05-24
Improve error handling in dealing with files (e.g. uploading).

### LINKED ISSUE #1164: Improve error handling when map is over filesize limit
Currently this is a 500 (and so probably just like an "unknown error" snackbar for the user). Ideally this should be a dialog + have an actual error code (both so we don't log it as a server error and also so the user knows what is going on).

2Pacalypse-:
For now we only handle the error when the max file size exceeds its limits, but it should be fairly straightforward to add additional error codes if we care about them.

Closes #1164


client/maps/action-creators.ts
```diff
@@ -38,10 +40,16 @@ export function uploadLocalMap(path: string, onMapSelect: (map: MapInfoJson) =>
...
         onMapSelect(map)
       },
-      () => {
-        externalShowSnackbar(
-          i18n.t('maps.local.uploadMapError', 'An error occurred while uploading the map'),
-        )
+      err => {
+        if (isFetchError(err) && err.code) {
+          if (err.code === FilesErrorCode.MaxFileSizeExceeded) {
+            externalShowSnackbar(i18n.t('maps.local.mapTooLarge', 'The map is too large'))
```
tec27, line 46: "Map is too large" can mean a lot of things other than file size, please be specific with these messages.

I'd prefer both of these errors to be a in a dialog instead of a snackbar as well

server/lib/files/handle-multipart-files.ts
```diff
@@ -0,0 +1,69 @@
...
+import koaBody from 'koa-body'
+import { FilesErrorCode } from '../../../common/files'
+import { asHttpError } from '../errors/error-with-payload'
+import logger from '../logging/logger'
+import { FilesError } from './files-error'
+
+function convertFormidableError(err: Error) {
+  // TODO(2Pac): Can't do a much better check here since koa-body uses an old version of formidable.
+  // Looks like it might get updated soon tho: https://github.com/koajs/koa-body/issues/241
+  if ('code' in err === false) {
```
tec27, line 12: nit: `!('code' in err)` feels like a preferable way to express this

server/lib/files/handle-multipart-files.ts
```diff
@@ -0,0 +1,69 @@
...
+    throw err
+  }
+
+  switch ((err as any).code) {
+    // TODO(2Pac): koa-body seems to be using an ancient version of formidable internally; in the
```
tec27, line 17: We need a test for this or something, no load-bearing comments please

tec27 (changes requested):

client/maps/action-creators.ts
2Pacalypse-: Do you think we should show the dialog as soon as the upload starts, or just show a dialog when/if the error happens?

If former, should the dialog be a modal one so the user _has_ to wait until the map finishes uploading?

tec27: I have no preferences

server/lib/files/handle-multipart-files.ts
2Pacalypse-: I was actually wrong in my initial assumption and they didn't change the meaning of the error code. They just renamed it to be more accurate, and added a new error code. So now there are error codes if the _total_ files size exceeds the limit, or if any of the files exceed it.

But anyway, I'm handling both here even though only one is used in this old version of koa-body.

PR #1173, opened 2025-05-31
Hide the map favorite button if the user is not logged in.

2Pacalypse-:
Initially I thought we should still show the button, but redirect them to the login page if they click it. But then I thought would the user expect for the map to be favorited once they log in? Unsure, so this seemed simpler :d


client/maps/map-thumbnail.tsx
```diff
@@ -261,7 +265,7 @@ export function MapThumbnail({
           onClick={onPreview}
         />
       ) : null}
-      {onToggleFavorite ? (
+      {onToggleFavorite && isLoggedIn ? (
```
tec27, line 268: Wouldn't it make more sense to deal with this at the place passing `onToggleFavorite`? This component is not connected to the store currently and that was an intentional design

tec27 (changes requested):

PR #1208, opened 2025-07-19
Add a way to zoom in and pan around map image preview.

2Pacalypse-:
Not a perfect implementation by all means, but might be enough for now?

Things that could be improved:
- add pinch/touch gestures for mobile support
- make the pan speed smoother on all zoom levels
- add the "slide" effect when panning fast and releasing a mouse (similar to what happens on phones / google maps)
- make the image quality even better than 2048x2048 on the highest zoom level?
- pull this logic into a common hook / control component that can be used on any element? (might come in handy if/when we add tournament brackets)

Also fixes one issue with buttons where if you release a mouse on a disabled button, a `mouseup` event on the document would not run (this conflicted with the slider), and adds a prop to render the slider component without a balloon that shows the value.

### DIFF
--- a/client/maps/map-preview.tsx
+++ b/client/maps/map-preview.tsx
@@ -1,11 +1,25 @@
-import { useEffect, useMemo } from 'react'
+import { useEffect, useMemo, useRef, useState } from 'react'
+import { useTranslation } from 'react-i18next'
 import styled from 'styled-components'
 import { SbMapId } from '../../common/maps'
 import { CommonDialogProps } from '../dialogs/common-dialog-props'
+import { MaterialIcon } from '../icons/material/material-icon'
+import { IconButton } from '../material/button'
+import { fastOutSlowInShort } from '../material/curves'
 import { Dialog } from '../material/dialog'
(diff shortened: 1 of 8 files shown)


tec27 (changes requested):

client/maps/map-preview.tsx
```diff
@@ -77,16 +78,16 @@ const ZoomableMapImageContainer = styled.div`
   border-radius: 4px;
 `
 
-const StyledMapInfoImage = styled(MapInfoImage)<{ $zoom: number }>`
+const StyledMapInfoImage = styled(MapInfoImage)`
   ${fastOutSlowInShort};
 
-  transform: scale(${props => props.$zoom});
+  transform: scale(var(--sb-map-image-zoom));
```
tec27, line 84: For future reference, the `--sb-map-image-` type naming is generally for things that are meant to be customizable by outside components, if you're just passing values internally you can do `--_foo` and stuff (but I don't really care that much, it just makes the code a bit less annoying to write)

PR #1223, opened 2025-07-28
Display the badge in the league entries of the home feed, if the league has it.

2Pacalypse-:


server-rs/src/leagues.rs
```diff
@@ -75,8 +76,32 @@ pub struct League {
...
+    #[graphql(skip)]
+    pub image_path_internal: Option<String>,
     pub rules_and_info: Option<String>,
     pub link: Option<String>,
 }
+
+#[ComplexObject]
+impl League {
+    async fn image_path(&self, ctx: &async_graphql::Context<'_>) -> async_graphql::Result<Option<String>> {
+        let Some(image_path_internal) = &self.image_path_internal else {
```
tec27, line 90: I think you need to embrace `match` and `Option.map` more but this is fine for now :P

server-rs/src/leagues.rs
```diff
@@ -17,8 +17,8 @@ impl LeaguesQuery {
             League,
             r#"
                 SELECT id, name, matchmaking_type as "matchmaking_type: _", description,
-                    signups_after, start_at, end_at, badge_path, image_path, rules_and_info,
-                    link
+                    signups_after, start_at, end_at, badge_path as badge_path_internal,
```
tec27, line 20: My preference would be to leave these as `_path` (still `graphql(skip)`) and add `_url` computed fields instead. We don't really need to maintain the same structure as the older APIs, as common components should generally have GQL wrapper versions that expose fragments to request the data they need (and thus would know how to map their own fragment into whatever underlying UI components)

PR #1229, opened 2025-08-02
Refactor matchmaking map pools manager in admin panel.

2Pacalypse-:
Now it uses all the newest patterns that we use elsewhere and it looks a bit nicer as well.

Also rewrote Carousel component in TypeScript and React functional components.


client/lists/devonly/carousel-test.tsx
```diff
@@ -0,0 +1,94 @@
...
+const ITEMS_PER_PAGE = 10
+
+export default function CarouselTest() {
+  const [page, setPage] = useState(1)
+  const [isLoading, setIsLoading] = useState(false)
+
+  const timerRef = useRef<NodeJS.Timeout | null>(null)
+
+  useEffect(() => {
+    if (timerRef.current) {
```
tec27, line 54: Seems like the intention was to do this in a cleanup function?

client/lists/carousel.tsx
```diff
@@ -0,0 +1,190 @@
...
+    setTranslateWidth(translateWidth => translateWidth + delta)
+  }
+
+  const contentStyle = { transform: `translateX(${translateWidth}px)` }
+  const showPrevButton = hasPrevItems
+  const showNextButton = hasNextItems && !infiniteListProps.isLoadingNext
+
+  return (
+    <CarouselContainer ref={carouselRef}>
+      <WindowListener event='resize' listener={calcCarouselWidth} />
```
tec27, line 145: We should probably move to a resizeobserver instead of this (I know this PR isn't to rewrite this component so it doesn't need to be part of this, but if we ever plan to use this going forward we should fix it up)

client/maps/maps-reducer.ts
```diff
@@ -120,6 +120,30 @@ export default immerKeyedReducer(DEFAULT_STATE, {
     }
   },
 
+  ['@matchmaking/getMatchmakingMapPoolsHistory'](state, action) {
```
tec27, line 123: imo, when we're making pages that don't use redux to store their data, it makes more sense to just make (or use) a more general action for loading this data into reducers. Those actions are basically just "here's the data" (and I think one might already exist for maps, I can't recall)

client/material/datetime-text-field.tsx
```diff
@@ -0,0 +1,21 @@
+import * as React from 'react'
+import { useCallback } from 'react'
+import { TextField, TextFieldProps } from './text-field'
+
+export function DateTimeTextField({
+  value,
+  onChange,
+  ...rest
+}: Omit<TextFieldProps, 'value' | 'onChange'> & {
+  value: string
```
tec27, line 10: Can you command what format this is expected to be in

client/material/datetime-text-field.tsx
```diff
@@ -0,0 +1,21 @@
...
+  onChange: (value: string) => void
+}) {
+  const onInputChange = useCallback(
```
tec27, line 13: We don't need useCallback any more, inline this

client/material/text-field.tsx
```diff
@@ -352,7 +352,7 @@ export function TextField({
     renderLabel = (
       <FloatingLabel
         htmlFor={id}
-        $hasValue={!!value}
+        $hasValue={!!value || type === 'datetime-local' || type === 'date'}
```
tec27, line 355: not a huge fan of this, would prefer a separate prop

common/matchmaking/matchmaking-map-pools.ts
```diff
@@ -0,0 +1,57 @@
...
+  return {
+    id: pool.id,
+    matchmakingType: pool.matchmakingType,
+    maps: pool.maps,
+    maxVetoCount: pool.maxVetoCount,
+    startDate: Number(pool.startDate),
+  }
+}
+
+export function fromMatchmakingMapPoolJson(pool: MatchmakingMapPoolJson): MatchmakingMapPool {
```
tec27, line 27: Do we need to do this? It's extremely uncommon that we have functions for this, we typically just leave them in JSON form

tec27 (changes requested):

client/maps/actions.ts
```diff
@@ -90,3 +91,12 @@ export interface LoadMapInfo {
   type: '@maps/loadMapInfo'
   payload: MapInfoJson
 }
+
+/**
+ * Loads multiple map infos directly. Useful if map infos are received from a source that doesn't
+ * dispatch through redux.
+ */
+export interface LoadMapInfos {
```
tec27, line 99: dunno how many places actually use `LoadMapInfo` but it doesn't really seem like we need both of these, those places could just use this

tec27 (approved):

PR #1234, opened 2025-08-12
Don't render the context menu contents if it doesn't have any items.

2Pacalypse-:
Can be reproduced by trying to right-click a chat message in the whispers session; or for users who don't have the "Delete message" action, by right-clicking any message in the chat.

This keeps the Popover still open, but figuring out how to close/not open the Popover in that case seems much harder :d, and I think just not rendering the Popover contents might be fine?

### DIFF
--- a/client/messaging/message-context-menu.tsx
+++ b/client/messaging/message-context-menu.tsx
@@ -154,14 +154,14 @@ function MessageContextMenuContents({
     )
   }
 
-  return (
+  return items.size > 0 ? (
     <MessageMenu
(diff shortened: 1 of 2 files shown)


tec27:
I fixed Popovers for chat before so that they don't render anything if they have no contents, so I think it should be fine (I'm not sure why whispers in particular need a special fix on top of that though?)

tec27:
Does seem likely it regressed due to always rendering a context inside the popover even if there were no items

PR #1242, opened 2025-08-14
Reorder the whisper sessions so the one with the latest message is on top.

2Pacalypse-:


server/lib/whispers/whisper-models.ts
```diff
@@ -10,10 +10,14 @@ export async function getWhisperSessionsForUser(userId: SbUserId): Promise<SbUse
   const { client, done } = await db()
   try {
     const result = await client.query<{ id: SbUserId }>(sql`
-      SELECT target_user_id AS id
-      FROM whisper_sessions
-      WHERE user_id = ${userId}
-      ORDER BY start_date;
+      SELECT ws.target_user_id as id, MAX(wm.sent) AS last_message_date
```
tec27, line 13: Selecting every message in the whisper to find the max one seems... pretty inefficient. Seems better to order them and limit 1? Additionally you need to handle whisper sessions that have no messages, so you probably need to select the max of the message time + the session creation time (if we don't store the latter, we probably need to for this to work)

tec27 (changes requested):

server/lib/whispers/whisper-models.ts
2Pacalypse-: > Selecting every message in the whisper to find the max one seems... pretty inefficient. Seems better to order them and limit 1?

My SQL-fu is obviously lacking, but I have no idea how to do this in an optimal way. I only made it work by using `CROSS JOIN LATERAL`, but that didn't give me better performance. If anything, it actually gave me worse from what I can tell. Also, that solution ignored the sessions with no messages for some reason.

This is stackoverflow post where I got the `CROSS JOIN LATERAL` solution from:
https://stackoverflow.com/questions/34246403/performance-of-max-vs-order-by-desc-limit-1

> Additionally you need to handle whisper sessions that have no messages, so you probably need to select the max of the message time + the session creation time (if we don't store the latter, we probably need to for this to work)

I actually did _try_ to handle whisper sessions that have no messages by just ordering them at the bottom with `NULLS LAST` :d. But I see now that we actually want to use the session start date (which we do actually store!) in case the last message date is `NULL`, so I fixed that now.

tec27: Anyway all of the queries involved here give absolutely godawful query plans, we're going to need more indexes to be able to do this feature

tec27: I think we can probably also remove the `sent DESC` index on `whisper_messages`, doesn't seem like it'd ever really be used properly. We can just adjust the message retrieval query to be using this new index as well

client/whispers/whisper-reducer.ts
```diff
@@ -104,6 +104,10 @@ export default immerKeyedReducer(DEFAULT_STATE, {
       text,
     })
 
+    // Reorder the sessions to put the one that got the message on top of the list
+    state.sessions.delete(target)
```
tec27, line 108: Unnecessary, by putting it first below it'll keep it first in the order and since it's a set it won't have a duplicate

server/lib/whispers/whisper-models.ts
```diff
@@ -154,10 +162,8 @@ export async function getMessagesForWhisperSession(
       FROM whisper_messages AS m
       INNER JOIN users AS u_from ON m.from_id = u_from.id
       INNER JOIN users AS u_to ON m.to_id = u_to.id
-      WHERE (
-        (m.from_id = ${userId1} AND m.to_id = ${userId2}) OR
-        (m.from_id = ${userId2} AND m.to_id = ${userId1})
-      ) `
+      WHERE m.user_low  = LEAST(${userId1}, ${userId2})::int4
```
tec27, line 165: You're passing both of these in, you can just do this in JS

tec27 (changes requested):

PR #1250, opened 2025-08-25
Replacing wait-for-it.sh with something we can provide multiple endpo…

WillSkates:
…ints to check at the same time.

Hey again!

This one is a very quick wait script that allows you to specify multiple TCP endpoints that you want to be checked.

The script essentially checks each of the endpoints one after the other, with a different timeout each time.

I think the _expected_ behaviour is that the script spawns each check at the same time. This one is purposely simpler than that just because I didn't want to give you anything more complicated than this in case I'm not available to help with any issues.

The reasons are the same as [https://github.com/ShieldBattery/ShieldBattery/pull/1249](the s3cmd PR) where pulling a script without any supply chain mitigations leaves you open to someone being an asshat.

This one doesn't protect you from those either except that the script itself is actually inside the repository.

Hope you're having a good week!

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,13 +1,6 @@
 # Get sqlx-cli so we can run migrations (Linux version needs to match what we run on below)
 FROM ghcr.io/shieldbattery/shieldbattery/base:sqlx-tools AS rust-tools
 
-# Clone external tools in a separate stage for better caching
-FROM node:22-alpine AS external-tools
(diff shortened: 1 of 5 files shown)


WillSkates:
Adding [WIP] as integration tests are failing.

WillSkates:
Integration tests now passing, removing [WIP].

Issue #1214, opened 2025-07-21
Change "private" map visibility to "unlisted"


tec27:
This is a more accurate description of its effects/security

PR #1256, opened 2025-09-28
Use ReduxMapThumbnail in all places that currently use MapThumbnail.

### LINKED ISSUE #1191: Use ConnectedMapThumbnail component in all places that currently use MapThumbnail
The list of files that still use the MapThumbnail component directly:

- [x] client/admin/map-pools.jsx
- [x] client/games/results.tsx
- [x] client/lobbies/active-lobby.jsx
- [x] client/lobbies/join-lobby.tsx
- [x] client/lobbies/lobby.jsx
- [x] client/maps/map-details.tsx
- [x] client/matchmaking/find-match-forms.tsx
- [x] client/replays/replay-info-display.tsx
- [x] client/users/match-history.tsx
- [x] client/users/mini-match-history.tsx

Using the ConnectedMapThumbnail will allow some of these places to not have to use a full map info object (like LobbySummary), and instead can just use the map ID.

2Pacalypse-:
As part of this work, I also changed what we show on the map thumbnail in certain places (screenshots below). I used my best judgement for this, but it can easily be changed.

Also, the MapThumbnail component is a bit weird because it uses one prop (`showMapName`; false by default) to decide whether to not only show the map name, but also the map actions. And map actions (view details, download, etc.) are all visible by default in ReduxMapThumbnail. But unless you set the `showMapName` to `true`, the actions won't be visible regardless. Not sure if we would even want to show map actions without the name, but yeah, something to think about.

Screenshots:
client/games/results.tsx (Removed the separate map name and just displayed it on the thumbnail):
image

client/lobbies/lobby.jsx (Also removed the separate map name and just displayed it on the thumbnail):
image

client/lobbies/join-lobby.tsx (Left this unchanged to what it was before, but maybe it could use an update?):
image

client/maps/map-details.tsx (Also left this unchanged, but maybe we could add preview/favorite actions here?):
image

client/matchmaking/find-match-forms.tsx (Added all actions here, but maybe should be left unchanged?):
image

client/replays/replay-info-display.tsx (Added all actions here, maybe could remove map name right below it now?):
image

client/users/match-history.tsx (Left this unchanged, not much we can display on a map thumbnail this small I think :d):
image

client/users/mini-match-history.tsx (Added all actions here):
image

Closes #1191

### DIFF
--- a/client/lobbies/lobby.tsx
+++ b/client/lobbies/lobby.tsx
@@ -1,4 +1,3 @@
-import { Immutable } from 'immer'
 import { List } from 'immutable'
 import React from 'react'
 import { WithTranslation, withTranslation } from 'react-i18next'
@@ -16,7 +15,6 @@ import {
   Team,
 } from '../../common/lobbies'
 import { Slot, SlotType } from '../../common/lobbies/slot'
-import { MapInfoJson } from '../../common/maps'
 import { BwTurnRate } from '../../common/network'
 import { RaceChar } from '../../common/races'
 import { SelfUserJson } from '../../common/users/sb-user'
@@ -370,10 +368,7 @@ class LobbyComponent extends React.Component<LobbyProps & WithTranslation> {
             label={t('lobbies.lobby.leaveLobby', 'Leave lobby')}
             onClick={onLeaveLobbyClick}
           />
-          <StyledMapThumbnail
-            mapId={(lobby.map as unknown as Immutable<MapInfoJson>).id}
-            showInfoLayer
-          />
+          <StyledMapThumbnail mapId={lobby.map!.id} showInfoLayer />
           <InfoItem>
             <InfoLabel as='span'>{t('lobbies.lobby.gameType', 'Game type')}</InfoLabel>
             <InfoValue as='span'>{gameTypeToLabel(lobby.gameType, t)}</InfoValue>


tec27:
Maybe `showMapName` could just be renamed to match its actual behavior? `showInfoLayer` or something

tec27:
Didn't we decide that having the favorite action in some places like these was weird since it could be for private (well, "unlisted") maps that were not your own? I guess if they really are just unlisted it is maybe fine

2Pacalypse-:
> Didn't we decide that having the favorite action in some places like these was weird since it could be for private (well, "unlisted") maps that were not your own? I guess if they really are just unlisted it is maybe fine

Yeah I was gonna tackle this issue next: https://github.com/ShieldBattery/ShieldBattery/issues/1214
I think treating them as "unlisted" makes more sense (and much easier to handle on our side :d) than having true "private" maps.

client/lobbies/lobby.tsx
```diff
@@ -375,10 +370,9 @@ class LobbyComponent extends React.Component<LobbyProps & WithTranslation> {
             label={t('lobbies.lobby.leaveLobby', 'Leave lobby')}
             onClick={onLeaveLobbyClick}
           />
-          <MapName>{(lobby.map as unknown as Immutable<MapInfoJson>).name}</MapName>
           <StyledMapThumbnail
-            map={lobby.map as unknown as Immutable<MapInfoJson>}
-            onPreview={onMapPreview}
+            mapId={(lobby.map as unknown as Immutable<MapInfoJson>).id}
```
tec27, line 374: I *think* this cast is unnecessary now

tec27 (approved):

PR #1277, opened 2026-01-07
Add ability to "ban" users from a particular league.

2Pacalypse-:
Users can be banned by admins with "manageLeague" permission. Banned user are still left in a league, but just marked as banned and their league points are not updated anymore. This serves as a built-in mechanism to prevent them from re-joining a league after being banned.

For now, we're just banning the single account of a user, and not all of their accounts.

### DIFF
--- a/client/leagues/action-creators.ts
+++ b/client/leagues/action-creators.ts
@@ -92,7 +92,10 @@ export function joinLeague(id: LeagueId, spec: RequestHandlingSpec<void>): Thunk
   })
 }
 
-export function getLeagueLeaderboard(id: LeagueId, spec: RequestHandlingSpec<void>): ThunkAction {
+export function getLeagueLeaderboard(
+  id: LeagueId,
+  spec: RequestHandlingSpec<GetLeagueLeaderboardResponse>,
+): ThunkAction {
   return abortableThunk(spec, async dispatch => {
     const result = await fetchJson<GetLeagueLeaderboardResponse>(
       apiUrl`leagues/${id}/leaderboard/`,
@@ -105,6 +108,8 @@ export function getLeagueLeaderboard(id: LeagueId, spec: RequestHandlingSpec<voi
       type: '@leagues/getLeaderboard',
       payload: result,
     })
+
+    return result
   })
 }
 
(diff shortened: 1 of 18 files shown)


tec27 (commented):
Did a deep review pass (Claude, on Travis's request). The core mechanics here are solid — the ban check sits at exactly the right choke point (the single path that writes league points and feeds the Redis leaderboard), and leaving the `league_users` row in place is a clean way to make bans survive re-join attempts. A few things to fix before merge:

**Must fix**

- `client/leagues/league-ban-notification-ui.tsx:35` — leftover debug `console.log('leagueName', leagueName)` (also flagged by the bot review, still present at head).

**Should fix**

- `server/lib/leagues/league-api.ts` (`banUserFromLeague`/`unbanUserFromLeague`) — neither endpoint checks current ban state before acting. Banning an already-banned user (double-click, two admins, or a retry after a failed notification insert) sends a second "You have been banned" notification, and unbanning a never-banned user sends a spurious "Your ban has been lifted" notification. An `if (leagueUser.isBanned)` guard before the write + notification fixes both.
- `LeagueAdminApi` has `@httpBeforeAll(ensureLoggedIn, checkAllPermissions('manageLeagues'))` but no `convertLeagueApiErrors`, so the `LeagueApiError(NotFound)` throws in the new ban/unban handlers surface as 500s instead of 404s. (Pre-existing for the admin `getLeague`, but this PR adds four more throw sites — adding the converter to the class's `httpBeforeAll` fixes all of them.)

**Nits**

- Untranslated user-facing strings in the admin UI: `'User banned'`/`'Error banning user'` snackbars, `'Admin actions'` tooltip, `Ban ${user.name}` menu items. Some admin-only surfaces are already untranslated so maybe fine, but snackbars elsewhere go through `t()`.
- A banned user viewing the league details page still sees a disabled "Joined" button, while the league list card correctly shows "Banned" — inconsistent, though the notification covers informing them.

Verified non-issues for the record: `updateLeagueUser` in game-result-service is the only writer of league points anywhere (server-rs never touches `league_users`), the Redis leaderboard only updates via the same skipped path, the migration is a safe constant-default boolean, and the join endpoint's UNIQUE_VIOLATION fallback returns the still-banned row so re-joins can't clear a ban.

PR #1279, opened 2026-01-17
Refactor the channel settings dialog into a full screen overlay.

2Pacalypse-:
The channel settings are now rendered in the same way our regular settings are. In fact, I'm actually sharing a lot of the same UI components between the two.

Even though the channel settings only has one page for now, we'll soon add more: ability to change permissions of channel members, ability to see banned users and unban them, invitations in case of the private channels, etc.

Currently I haven't implemented the logic which would prevent user from navigating to a different channel settings page, or close the channel settings overlay, in case there are some changes. I wasn't exactly sure what's the best way to deal with that at the moment.


tec27 (commented):
Did a deep review pass (Claude, on Travis's request) — this looks ready, merging it. The extraction into `settings-content.tsx` is clean (regular app settings are behavior-identical — verified the styles/markup moved verbatim and the channel overlay's 840px override doesn't leak into app settings), all the old dialog's functionality is preserved (fields, validation, banner/badge upload/remove, error states, permission gating client and server side, integration-test hooks), and the overlay mirrors the proven `ConnectedSettings` structure including the history-state handling. Non-blocking notes for follow-up (fine to fold into #1280):

- `general-settings.tsx` — the old dialog passed `showFileName` to both banner/badge `SingleFileInput`s; the new code omits it, so after picking a file the user no longer sees its name. The live preview mostly covers it, but flagging in case it wasn't intentional.
- `client/material/file-input.tsx` — the added `flex-shrink: 0` on `FileName` defeats its `singleLine` ellipsis (truncation needs the flex item to be shrinkable), so long file names now overflow instead of ellipsizing. Only remaining consumer is the league admin page — and since channel settings no longer shows file names, this change looks vestigial and could just be dropped.
- `form-hook.ts` — the `hasChanges` doc says "changes since the last submission", but `dirtyFields` is never cleared on submit or when a field returns to its initial value, so it really means "ever touched". Harmless here (save closes the overlay), but worth aligning the comment/behavior before more consumers rely on it — especially if it ends up powering an unsaved-changes guard.
- Nit: the `<React.Suspense>` around `ConnectedChannelSettings` in `channel.tsx` does nothing — nothing inside is lazy and `SettingsContent` has its own Suspense; `ConnectedSettings` is mounted without one.

On the unsaved-changes question from the description: there's no reusable navigation-guard infra to borrow (regular settings sidestep it by auto-saving via `onValidatedChange`, which doesn't fit an image-moderated explicit-save flow). Cheapest consistent option: lift the `hasChanges` flag this PR already adds from the form up to `ChannelSettings`, and have `onCloseSettings`/page-switch open a confirm dialog when it's set — with the caveat above that `hasChanges` is currently over-eager.

PR #1280, opened 2026-01-24
Implement a way to change channel permissions for each user.

2Pacalypse-:


2Pacalypse-:
@claude review this please

tec27 (commented):
Did a deep review pass (Claude, on Travis's request), including a careful look at the unresolved security thread. Overall this is in good shape — small server delta following existing conventions, thorough unit tests, and the client page is a faithful copy of the proven channel-list infinite-scroll pattern. Needs #1279 to land first, then a rebase.

**On the bot's privilege-escalation thread** — mostly refuted, with one real gap:

- *Self-escalation*: technically possible via the API but confers nothing — `moderateUser` already treats `editPermissions` as implying kick/ban server-side, and `changeTopic`/`togglePrivate` aren't enforced anywhere server-side yet. The UI deliberately allows self-editing of everything except the `editPermissions` checkbox. Also worth noting the caller-only authz check is identical on master — not introduced here.
- *Modifying the owner's permissions*: the API does permit it, but it's inert — the owner's authority everywhere derives from `channelInfo.ownerId === userId` comparisons, never stored flags. Still, `moderateUser` rejects owner targets and the UI disables owner rows, so a one-line server-side guard (`targetId === channelInfo.ownerId` → reject) is cheap defense-in-depth and keeps API and UI consistent.
- *Granting `editPermissions` to others*: that's the feature itself — a deliberately flat delegation model.

**The one real policy gap (worth an explicit decision):** `moderateUser` enforces "only server moderators and channel owners can moderate channel moderators," but any `editPermissions` holder can first zero out a fellow moderator's permissions via `updateUserPermissions` (which has no target-side restriction), and *then* kick/ban them — including the person who granted them `editPermissions` in the first place. That launders around the moderation hierarchy. Either mirror `moderateUser`'s guard (require owner/server-admin to edit users who currently hold `editPermissions`), or document the flat model as accepted.

**Smaller items**

- `listUserChannelEntries` returns every member's `channelPreferences` (`hideBanner`) to channel-level owners/editors, while the new doc comment on `UserChannelEntry` says preferences "should be private to the user themselves and admins." The client never uses them in the listing — omit them (or fix the comment).
- `listUserChannelEntries` never uses `this` — per AGENTS.md it should be a module-level function (the existing bot thread on this is correct).
- Offset pagination orders by permission count, so editing a user's permissions between page loads shifts positions → duplicated/skipped entries on infinite scroll (and duplicate React keys). Add a stable tiebreaker (`cu.user_id`) and consider deduping on concat.
- The admin (`isAdmin`) bypass only skips the permission check, not the membership check — a server admin not in the channel gets `NotInChannel`, unlike the `getChannelUsers`/`deleteMessage` admin patterns. Unreachable via current UI, but inconsistent.
- Nit: `ChannelSettingsPayload.onSuccess` makes this the first dialog payload containing a function stored in Redux — non-serializable state, pattern deviation.
- Nit: `userPermissionsThrottle` (20/min) is shared between the debounced search listing and permission saves, so active searching can throttle-block saves; the analogous channel search uses a much higher limit.

Verified non-issues: search input parameterized + escaped like the existing channel search, users who left/were banned produce a clean `TargetNotInChannel` 4xx, and live updates flow correctly (`permissionsChanged` → target's self-permissions, `userProfileChanged` → loaded profiles so menu enablement reacts).

PR #1281, opened 2026-02-14
Add match history filtering and matchup support.

2Pacalypse-:
Add filter bar to match history with support for filtering by game type (ranked/custom), format, duration, matchup, map name, and player name, with sorting options. Filter state is persisted in URL search params.

This filter bar is currently used in user's match history, but we'll probably reuse it in our main Games page that shows all (matchmaking) games on the platform, as well as in our League details page where we can show a list of games particular to that league.

This also adds a couple of columns in our games table in DB, selected_matchup and assigned_matchup. Selected matchup is recorded as soon as the game record is created and it uses user's selected race. Additionally, this column is saved for the games including computers as well, and it's this column that is used to filter the games by "format" (e.g. 1v1, 2v2, 3v3, 4v4).

On the other hand, assigned matchup is saved after the game results are reconciled, and it uses user's assigned race (resolves random to actual race the user got in the game). Additionally, this column is *not* saved for games that have any computers in them since computers are not included in our games results (yet). Also, this column is used for filtering games by matchup.

The data for these 2 columns is being backfilled by an SQL function in the latest migration. Also, I'll provide a dry-run version of this function that can be run manually to verify its results before actually making any changes to the DB.

This PR also adds a Filter Chip component from Material Design 3.

And finally, I also renamed "search" match history terminology to "get" match history terminology to be a bit more in line with other places we do similar things.

### DIFF
--- a/client/games/game-filter-bar.tsx
+++ b/client/games/game-filter-bar.tsx
@@ -1,0 +1,381 @@
+import { Variants } from 'motion/react'
+import { useState } from 'react'
+import { useTranslation } from 'react-i18next'
+import styled from 'styled-components'
+import {
+  ALL_GAME_FORMATS,
+  createEmptyMatchup,
+  decodeMatchup,
+  EncodedMatchupString,
+  encodeMatchup,
+  GameDurationFilter,
+  GameFormat,
+  GameSortOption,
+  getDurationLabel,
+  getSortLabel,
+} from '../../common/games/game-filters'
+import { MaterialIcon } from '../icons/material/material-icon'
+import { useKeyListener } from '../keyboard/key-listener'
+import { TextButton } from '../material/button'
+import { FilterChip } from '../material/filter-chip'
+import { SelectableMenuItem } from '../material/menu/selectable-item'
+import { Popover, usePopoverController, useRefAnchorPosition } from '../material/popover'
+import { TextField } from '../material/text-field'
+import { ContainerLevel, containerStyles } from '../styles/colors'
+import { FlexSpacer } from '../styles/flex-spacer'
+import { labelLarge, labelMedium } from '../styles/typography'
+import { MatchupFilter } from './matchup-filter'
+
+const FilterBarContainer = styled.div`
+  display: flex;
+  flex-wrap: wrap;
+  align-items: center;
+  gap: 16px;
+  width: 100%;
+  min-height: 56px;
+`
+
+const FiltersLabel = styled.div`
+  ${labelLarge};
+  display: flex;
+  align-items: center;
+  gap: 4px;
+  color: var(--theme-on-surface-variant);
+`
+
+const popoverVariants: Variants = {
+  entering: { opacity: 0, y: -8 },
+  visible: { opacity: 1, y: 0 },
+  exiting: { opacity: 0, y: -8 },
+}
+
+const DURATION_OPTIONS = [
+  GameDurationFilter.All,
+  GameDurationFilter.Under10,
+  GameDurationFilter.From10To20,
+  GameDurationFilter.From20To30,
+  GameDurationFilter.Over30,
+] as const
+
+const SORT_OPTIONS = [
+  GameSortOption.LatestFirst,
+  GameSortOption.OldestFirst,
+  GameSortOption.ShortestFirst,
+  GameSortOption.LongestFirst,
+] as const
+
+export interface GameFilterBarProps {
+  ranked: boolean
+  setRanked: (v: boolean) => void
+  custom: boolean
+  setCustom: (v: boolean) => void
+  duration: GameDurationFilter
+  setDuration: (v: GameDurationFilter) => void
+  sort: GameSortOption
+  setSort: (v: GameSortOption) => void
+  mapName: string
+  setMapName: (v: string) => void
+  playerName: string
+  setPlayerName: (v: string) => void
+  format?: GameFormat
+  setFormat: (v?: GameFormat) => void
+  matchup?: EncodedMatchupString
+  setMatchup: (v?: EncodedMatchupString) => void
+  className?: string
+}
+
+/**
+ * A filter bar for game history with immediate-apply filters (chips, dropdowns)
+ * and an advanced filters panel with draft state.
+ */
+export function GameFilterBar({
+  ranked,
+  setRanked,
+  custom,
+  setCustom,
+  duration,
+  setDuration,
+  sort,
+  setSort,
+  mapName,
+  setMapName,
+  playerName,
+  setPlayerName,
+  format,
+  setFormat,
+  matchup,
+  setMatchup,
+  className,
+}: GameFilterBarProps) {
+  const { t } = useTranslation()
+  const [anchorRef, anchorX, anchorY, refreshAnchorPos] = useRefAnchorPosition('left', 'bottom')
+  const [opened, openPopover, closePopover] = usePopoverController({ refreshAnchorPos })
+
+  const hasAdvancedFilters = !!mapName || !!playerName || !!format || !!matchup
+  const hasActiveFilters =
+    ranked || custom || duration !== GameDurationFilter.All || hasAdvancedFilters
+
+  return (
+    <FilterBarContainer className={className}>
+      <FiltersLabel>
+        <MaterialIcon icon='filter_list' size={20} />
+        {t('game.filters.label', 'Filters')}
+      </FiltersLabel>
+
+      <FilterChip
+        label={t('game.filters.ranked', 'Ranked')}
+        selected={ranked}
+        onClick={() => setRanked(!ranked)}
+      />
+      <FilterChip
+        label={t('game.filters.custom', 'Custom')}
+        selected={custom}
+        onClick={() => setCustom(!custom)}
+      />
+
+      <FilterChip
+        label={getDurationLabel(duration, t)}
+        icon={<MaterialIcon icon='timer' filled={false} size={18} />}>
+        {DURATION_OPTIONS.map(d => (
+          <SelectableMenuItem
+            key={d}
+            text={getDurationLabel(d, t)}
+            selected={duration === d}
+            onClick={() => setDuration(d)}
+          />
+        ))}
+      </FilterChip>
+
+      <FilterChip
+        ref={anchorRef}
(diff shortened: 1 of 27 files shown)


2Pacalypse-:
Also, here's the dry-run version of the SQL function which should show what changes the migration will make without actually changing anything in the DB (NOTE: I can't guarantee 100% its correctness :d):
```sql
-- Dry-run version of backfill_matchups from migration 20260212000000_add_matchup_columns.sql
-- Returns a table of (game_id, selected_matchup, assigned_matchup) showing what the migration
-- would write, WITHOUT actually updating any rows.

CREATE OR REPLACE FUNCTION backfill_matchups_dry_run()
RETURNS TABLE(game_id text, config jsonb, results jsonb, computed_selected_matchup text, computed_assigned_matchup text) AS $$
DECLARE
  r RECORD;
  team_races text[];
  all_team_strings text[];
  team_str text;
  player_race text;
  player_id int;
  team jsonb;
  player jsonb;
  result_pair jsonb;
  teams_count int;
  team0_count int;
  selected text;
  assigned text;
  is_1v1_single_team boolean;
  has_computers boolean;
BEGIN
  FOR r IN SELECT g.id, g.config, g.results FROM games g LOOP
    teams_count := jsonb_array_length(r.config->'teams');
    selected := NULL;
    assigned := NULL;
    is_1v1_single_team := false;

    has_computers := EXISTS (
      SELECT 1
      FROM jsonb_array_elements(r.config->'teams') AS t,
           jsonb_array_elements(t) AS p
      WHERE (p->>'isComputer')::boolean = true
    );

    -- Determine if we can compute matchups
    IF teams_count = 1 THEN
      team0_count := jsonb_array_length(r.config->'teams'->0);
      IF team0_count = 2 THEN
        -- 1v1 stored as single team of 2 - split into two teams
        is_1v1_single_team := true;
      ELSE
        -- Melee with >2 players, can't determine teams
        CONTINUE;
      END IF;
    END IF;

    -- Compute selected_matchup
    IF is_1v1_single_team THEN
      -- Split single team into two 1-player teams
      all_team_strings := ARRAY[
        (r.config->'teams'->0->0->>'race'),
        (r.config->'teams'->0->1->>'race')
      ];
      -- Sort the team strings
      SELECT array_agg(s ORDER BY s) INTO all_team_strings FROM unnest(all_team_strings) AS s;
      selected := array_to_string(all_team_strings, '-');
    ELSE
      -- Multiple teams: sort races within each team, collect team strings
      all_team_strings := ARRAY[]::text[];
      FOR i IN 0..(teams_count - 1) LOOP
        team := r.config->'teams'->i;
        team_races := ARRAY[]::text[];
        FOR j IN 0..(jsonb_array_length(team) - 1) LOOP
          team_races := array_append(team_races, team->j->>'race');
        END LOOP;
        -- Sort races within the team
        SELECT array_agg(s ORDER BY s) INTO team_races FROM unnest(team_races) AS s;
        team_str := array_to_string(team_races, '');
        all_team_strings := array_append(all_team_strings, team_str);
      END LOOP;
      -- Sort team strings lexicographically
      SELECT array_agg(s ORDER BY s) INTO all_team_strings FROM unnest(all_team_strings) AS s;
      selected := array_to_string(all_team_strings, '-');
    END IF;

    -- Compute assigned_matchup (only if results exist and no computer players)
    IF r.results IS NOT NULL AND NOT has_computers THEN
      IF is_1v1_single_team THEN
        -- For 1v1, look up each player's assigned race from results
        all_team_strings := ARRAY[]::text[];
        FOR k IN 0..1 LOOP
          player := r.config->'teams'->0->k;
          player_id := (player->>'id')::int;
          player_race := NULL;
          FOR result_pair IN SELECT jsonb_array_elements(r.results) LOOP
            IF (result_pair->0)::int = player_id THEN
              player_race := result_pair->1->>'race';
              EXIT;
            END IF;
          END LOOP;
          IF player_race IS NULL THEN
            -- Player not found in results, skip this game
            all_team_strings := NULL;
            EXIT;
          END IF;
          all_team_strings := array_append(all_team_strings, player_race);
        END LOOP;

        IF all_team_strings IS NOT NULL THEN
          SELECT array_agg(s ORDER BY s) INTO all_team_strings FROM unnest(all_team_strings) AS s;
          assigned := array_to_string(all_team_strings, '-');
        END IF;
      ELSE
        -- Multiple teams
        all_team_strings := ARRAY[]::text[];
        FOR i IN 0..(teams_count - 1) LOOP
          team := r.config->'teams'->i;
          team_races := ARRAY[]::text[];
          FOR j IN 0..(jsonb_array_length(team) - 1) LOOP
            player := team->j;
            player_id := (player->>'id')::int;
            player_race := NULL;
            FOR result_pair IN SELECT jsonb_array_elements(r.results) LOOP
              IF (result_pair->0)::int = player_id THEN
                player_race := result_pair->1->>'race';
                EXIT;
              END IF;
            END LOOP;
            IF player_race IS NULL THEN
              team_races := NULL;
              EXIT;
            END IF;
            team_races := array_append(team_races, player_race);
          END LOOP;

          IF team_races IS NULL THEN
            all_team_strings := NULL;
            EXIT;
          END IF;

          SELECT array_agg(s ORDER BY s) INTO team_races FROM unnest(team_races) AS s;
          team_str := array_to_string(team_races, '');
          all_team_strings := array_append(all_team_strings, team_str);
        END LOOP;

        IF all_team_strings IS NOT NULL THEN
          SELECT array_agg(s ORDER BY s) INTO all_team_strings FROM unnest(all_team_strings) AS s;
          assigned := array_to_string(all_team_strings, '-');
        END IF;
      END IF;
    END IF;

    -- Instead of UPDATE, return the row
    IF selected IS NOT NULL OR assigned IS NOT NULL THEN
      game_id := r.id;
      config := r.config;
      results := r.results;
      computed_selected_matchup := selected;
      computed_assigned_matchup := assigned;
      RETURN NEXT;
    END IF;
  END LOOP;
END;
$$ LANGUAGE plpgsql;

-- Run the dry run and display results
SELECT * FROM backfill_matchups_dry_run();

-- Clean up
DROP FUNCTION backfill_matchups_dry_run();
```

tec27 (commented):
Did a deep review pass (Claude, on Travis's request). The feature code is in great shape — every interpolated filter value is parameterized, the canonical matchup design (selected vs assigned, wildcard expansion with canonical dedupe) is well thought out and unit-tested, and the URL-state handling follows existing patterns. Two significant issues need addressing before this merges, both around the migration/matchup computation:

**Blocking: the migration will lock the `games` table for its full duration**

`migrations/20260212000000_add_matchup_columns.sql` runs as a single sqlx transaction (no `-- no-transaction` marker). The initial `ALTER TABLE games ADD COLUMN` takes an ACCESS EXCLUSIVE lock that is then held until commit — i.e. through both full-table index builds *and* the entire row-by-row `backfill_matchups()` loop over every game since 2020. While that runs, **all** reads and writes to `games` block: game registration, result reporting, match history, game pages. Realistically minutes to tens of minutes of outage on production data, plus the backfill rewrites every row in one shot (~100% heap bloat until vacuum), and any single unexpected config shape aborts and rolls back everything.

Suggested split: (a) migration does only the `ALTER TABLE` + index creation (instant — nullable text columns are metadata-only); (b) the backfill runs as the manually-invoked function you already planned for the dry-run, batched (e.g. by id ranges) outside one long transaction. Your dry-run instinct was right — make that the actual mechanism. A set-based `UPDATE ... FROM (... lateral jsonb expansion)` would also be much faster than the plpgsql per-row loop, if you want to keep it simple.

**Major: empty observer teams break matchup computation (runtime and backfill disagree)**

Melee lobbies with `allowObservers` create an observers `Team`, and the game-config construction in `server/lib/wsapi/lobbies.ts` maps over all lobby teams, leaving the observer team as an **empty array** in `config.teams`. Consequences:

- Runtime: a 2-player melee lobby with observers has `config.teams = [[p1, p2], []]` → `getTeamsFromConfig` sees 2 teams → `computeMatchupString` hits the empty-team check and returns `null`. So melee 1v1s with observers enabled (common for casted games) silently never get a matchup and never match any format/matchup filter, while the identical lobby without observers produces `p-z` via the single-team split.
- Backfill: the same historical configs produce *garbage non-NULL* values instead — `array_to_string` silently drops the NULL element for the empty team, so `[[p,z],[]]` backfills `selected_matchup = 'pz'` (no dash, matches nothing), inconsistent with the runtime `null`.

Fix: filter out empty teams in `getTeamsFromConfig` before the length checks (which also makes 2-player-melee-with-observers correctly resolve via the split path), and mirror that in the backfill.

**Smaller items**

- `server/lib/games/game-models.ts` — `ORDER BY g.game_length ASC|DESC NULLS LAST` has no deterministic tiebreaker; all unreconciled games tie at NULL, so infinite scroll can duplicate/skip rows (and duplicate React keys, key = game.id). Add e.g. `, g.start_time DESC, g.id`.
- The format filter is a regex match (`g.selected_matchup ~ '^[prtz]{N}-[prtz]{N}$'`) which a btree index can never serve — `idx_games_selected_matchup` is dead weight for these queries (they're driven by the `games_users` PK anyway), but the planned global Games page would seq-scan on format filtering. Worth rethinking before that ships (expression index on team sizes, or a generated column).
- `client/users/match-history.tsx` `parseMatchup` — unlike `parseDuration`/`parseSort`/`parseFormat`, it passes any non-empty string through, so a hand-edited `?matchup=foo` fails server-side Joi and the user gets the generic error screen instead of the filter being ignored. Validate client-side like the others.
- Nit: selecting a format immediately encodes an all-wildcard matchup, so URLs carry a no-op `matchup=____-____` param the server ignores. Cleaner to leave it unset until a race is chosen.
- Nit: the backfill is a point-in-time snapshot — games reconciled by still-running old code during the deploy window keep NULL matchups forever. Worth a catch-up pass once new code is live (another argument for the manually-run batched backfill).

Verified non-issues: all dynamic SQL is parameterized (`sqlRaw` only used with compile-time constants, LIKE inputs escaped), matchup normalization is consistent client → Joi → `expandMatchupFilter` → canonical form, both game-creation paths funnel through the single `registerGame` call site, the search→get rename has no leftover call sites, and the results/config JSONB shapes the backfill assumes have been stable since 2020.

PR #1283, opened 2026-02-28
Implement the global games page which shows matchmaking games.

2Pacalypse-:
This page shows both the live games and the ones that already happened.

I also added a "watch replay" action directly from the games list; this is applied both in this games page, as well as the user's match history page.

The games page uses a similar filter as the user's match history page (without custom/ranked filter). I haven't attempted to share the SQL generation between the 2 model methods, but might do it when I add another page that uses game filters. That would be the league games page which is shows as a tab in league details.


tec27 (commented):
Did a deep review pass (Claude, on Travis's request). Solid, idiomatic work — the authorization story checks out (only matchmaking games with reconciled results are served, so private lobby games can't leak; replay access gated by `canUserAccessReplay`; watch-replay correctly Electron-only through typed IPC), and live-game exposure is no worse than the existing home feed. Currently blocked on conflicts with the `game-filters` base, and #1281 needs to land first. Things to address along with the rebase:

**Performance of the public list endpoint (most important)**

`getGames` in `server/lib/games/game-models.ts` runs `WHERE g.config->>'gameSource' = 'MATCHMAKING' AND g.results IS NOT NULL ORDER BY g.start_time DESC LIMIT 40 OFFSET n` — and no index on `games` can serve it (the only start_time index is partial on the *opposite* predicate, `WHERE game_length IS NULL`). Every page view of an unauthenticated, public endpoint does a sequential scan + top-N sort over the entire games table; the `ShortestFirst`/`LongestFirst` sorts and `playerName ILIKE '%…%'` are worse. The per-user/IP throttle doesn't bound aggregate public traffic. Suggest a partial index like `(start_time DESC) WHERE config->>'gameSource' = 'MATCHMAKING' AND results IS NOT NULL` (plus equivalents for the game_length sorts).

Related validation gaps on the same endpoint: `offset: Joi.number().min(0)` has no `.max()` (so `?offset=10000000` forces producing 10M sorted rows) and no `.integer()` — `offset=1.5` passes Joi, produces `OFFSET 1.5`, and 500s on the bigint cast. (Copied from the match-history validator, but that one sits behind a user's own history; this one is public.)

**Pagination drift on a moving window**

The list shows completed games sorted by `start_time DESC`, and games complete continuously — every game that finishes between page loads shifts the window, so page N+1 re-serves rows from page N. The client concatenates without dedupe and renders `key={game.id}` → duplicate keys and visibly repeated rows. Keyset pagination on `(start_time, id)` would fix it properly; at minimum add an `id` tiebreaker to the ORDER BY and dedupe on concat.

**Smaller items**

- `client/games/game-list.tsx` and `client/home/home.tsx` both define a byte-identical `fragment LiveGames_FeedFragment on Query` (and duplicate `LiveGamesFeed` components). Codegen only tolerates this because the strings are identical — the first edit to one copy breaks `pnpm gen-graphql` with a duplicate-fragment-name error. Export from one place.
- `useAppSelector(s => gameIds?.map(id => s.games.byId.get(id)!))` (game-list.tsx and match-history.tsx) returns a fresh array on every store update, so the whole list re-renders on any Redux action anywhere. Use a shallow-equal comparator or derive in render.
- The `useMemo` in `client/games/game-list-entry.tsx` from the existing bot thread is still present — redundant under react-compiler per AGENTS.md.
- The ~40-line replay-info-building block (signed URLs, error handling) is duplicated verbatim between `game-api.ts` and `user-api.ts` — easy extraction to do alongside the planned SQL unification.
- Informational: the "Live games" section uses a one-shot query (ttl 10s) with nothing re-executing it, so finished games linger and can appear in both sections. Matches the home-feed behavior so fine for now, but worth a TODO if this page is meant to feel live.

PR #1286, opened 2026-03-28
Implement new matchmaker in rust.

2Pacalypse-:


tec27 (commented):
Did a deep review pass (Claude, on Travis's request). First the good news: I audited all six previously-resolved review threads against the current head and every one is genuinely fixed (requeue-failure cleanup via `handleRsRestart`, team-size guards, the find/remove mutex scope, `urlPath`, dropped-match notification, and `total_cmp` replacing the NaN-panicking `partial_cmp().unwrap()`). The Rust core itself is clean, well-documented, panic-conscious, and well-tested.

The remaining weak spot is the Node↔Rust state-reconciliation layer — three related defects, the first of which will fire after routine deploys:

**1. Stale process-token baseline after an idle-period restart falsely kicks the next queuer and ghosts them in the Rust queue**

`matchmaking-service.ts`: (a) a player queues → `lastKnownProcessToken = X` (only fetched when `undefined`); (b) queue empties → watchdog stops, but `lastKnownProcessToken` is never cleared; (c) server-rs restarts while idle (token now Y) — a routine deploy; (d) the next player queues — `rsQueuePlayer` succeeds against the *new* process, but the token isn't refetched since it's non-undefined; (e) the watchdog fires, sees Y ≠ X, and `handleRsRestart`s a player who is validly queued. Worse, `handleRsRestart` never calls `rsCancelPlayer`, so they remain in the Rust queue — every subsequent `find()` 409s with `alreadyInQueue` (surfaced as a raw error, not a `MatchmakingServiceError`) until the ghost entry gets matched, which then ejects an innocent opponent via the drop path.

Fix: clear `lastKnownProcessToken` when the watchdog stops (or always refetch after a successful queue), and have `handleRsRestart` best-effort `rsCancelPlayer`.

**2. Restart detection kicks players who are mid-match, which can end in bans for innocent players**

`handleRsRestart([...this.playerQueueData.keys()])` includes players in the accept/draft/load phases (their `playerQueueData` lives until `runMatch`'s finally). A match in progress no longer depends on the Rust queue, but the restart path deletes their `queueEntries`, so `accept()` throws `NoActiveMatch` — they physically cannot accept — and the accept timeout then `markForBan`s them, ending in `matchmakingBanService.banUser` for players who did nothing wrong. Restart handling should be scoped to entries without a `matchId`.

**3. Orphaned Rust queue entries have no reconciliation path**

Three triggers, same outcome: (a) `rsGetProcessToken()` failing right after a successful `rsQueuePlayer` — the catch deletes only local state; (b) `rsCancelPlayer` failure on cancel/disconnect is only logged; (c) a Node restart loses all in-memory maps while Rust keeps its queue. In each case the user's next `find()` hard-fails with 409 `alreadyInQueue` and there's no TTL, startup flush, or admin escape hatch. A cheap mitigation: on `alreadyInQueue` in `queueSoloPlayer`, call `rsCancelPlayer` and retry once. Longer-term, a Rust-side entry TTL would bound all of these.

**Smaller items**

- A single failed token fetch nukes the entire queue (any transient server-rs hiccup ejects everyone with errors). Consider requiring 2 consecutive failures.
- When one player cancels in the tiny window before Node processes a `matchFound`, *all* players in the match get ejected with `matchmakingServiceError` even though Node holds valid requeue tickets for the innocent ones — requeueing them would be strictly better.
- `rsRequeuPlayer(ticket).catch(...)` is fire-and-forget, so a player who cancels immediately after a `requeue` event can have the DELETE reach Rust before the POST, resurrecting them as a ghost entry (feeds #3).
- Dead code: the old Node matchmaker wasn't removed — `matchmaker.ts` (only `calcEffectiveRating` is still used, by draft-state.ts), `matchmaker-queue.ts`, `lazy-scheduler.ts`, and their tests. AGENTS.md says to delete unused code during refactoring; the `shieldbattery_matchmaker_queue_size` gauge also silently disappears from metrics.
- If the spawned `search_loop` task ever panics it dies silently — `/matchmaker/token` keeps answering, so the Node watchdog never notices that matches stopped forming. Consider `unwrap_or_else(|e| e.into_inner())` on the locks or restarting the loop on panic.
- Nits: `rsRequeuPlayer` typo (→ `rsRequeuePlayer`); `RandomQueueSelector`'s comment claims front-bias but `random_range(0..=i)` with replacement is textbook reservoir sampling — uniform (your own regression test confirms it); both Criterion benches register the ID `"matches"`, so 2v2 results overwrite the 1v1 baseline.

Test-coverage note: the Rust core is well covered, but the rewritten Node service has zero tests — and findings 1–3 all live there (watchdog/restart logic, `handleMatchFound` guard paths, requeue flow). Worth adding before the multiqueue PR stacks more state on top of this layer.

PR #1288, opened 2026-06-11
Support queueing for multiple matchmaking types at once

2Pacalypse-:
Players can now search for several matchmaking types simultaneously, with a race selection per type. The matchmaker tracks per-mode ratings for each queued player instead of a single rating, and queue tickets preserve those ratings for requeues. The find match UI has been reworked to allow multi-type selection, with a new devonly test page for it.


tec27 (commented):
Did a deep review pass (Claude, on Travis's request), including digging into the CI failures. The core multiqueue design verified out nicely — a single queue entry per player with an `EnumSet` of modes plus a per-mode ratings map, with matched-player removal happening under one held mutex, makes the hard cross-queue races (double-match, cancel-vs-match) structurally impossible rather than carefully avoided. Findings below.

**CI diagnosis**

- `fmt` — **PR-caused.** `cargo fmt --check` fails on the PR's new code: single-line `PlayerModeRating { ... }` struct literals and `HashMap::from` tuples in `server-rs/benches/matchmaker.rs`, `src/matchmaking/api.rs`, and `src/matchmaking/matchmaker.rs`. Run `cargo fmt`.
- `unit-tests` — **not PR-caused.** `client/messaging/common-message-layout.test.tsx` fails with "Electron failed to install correctly"; master's CI run from today fails identically. Runner-environment issue. ⚠️ However, because the test step fails, **Typecheck and Lint never ran** — and prettier locally flags `server/lib/matchmaking/matchmaking-api.ts`, `matchmaking-service.ts`, and `client/games/action-creators.ts`, so Lint will fail once unit tests are unblocked. `pnpm run lint --fix` is needed too.
- `integration-tests` — **not PR-caused.** Hung inside `playwright install` (no output after the Chromium download hit 100%) until the 6-hour timeout; same binary-install failure class as the unit-test issue.
- `claude-review` — **branch drift.** 401: the workflow file on this branch differs from master (master got several claude-review workflow fixes after `matchmaker2` branched). Merging master into the stack fixes it.

**High: users with unset preferences silently cannot queue**

`find-match.tsx` `handleFindMatch` only includes a type if `prefs && 'matchmakingType' in prefs` — but for a user who has never saved preferences for a type, the server's preferences subscription deliberately sends `{}` (see the comment in `matchmaking-preferences-service.ts`). So: a fresh user selects 1v1, clicks Find match → `allPrefs.length === 0` → silent `return` before `setIsSubmitting(true)`; the button does nothing with no feedback. And a user selecting two types where one has `{}` prefs silently queues only the configured one while both chips look ready. The old path defended against exactly this — the previous `findMatch` action creator fell back to `defaultPreferences(...)` for empty prefs, and that fallback was deleted here with no replacement (`searchAgainFromGame` got a snackbar guard; `handleFindMatch` got nothing). Suggest restoring the per-type default-preferences fallback, or surfacing an error/disabling the button when a selected type has no usable prefs.

**Medium: missing per-mode rating silently becomes rating 0.0**

`matchmaker.rs` documents that `Player.ratings` "must contain an entry for every mode the player is queued for", but `effective_rating` does `.get(&mode).copied().unwrap_or_default()` and nothing validates the invariant at insert/requeue — a Node-side bug would silently match someone as a 0-MMR player instead of failing loudly. Notably the PR's own `finds_all_modes_in_order` test violates the invariant (players queue for `Match1v1 | Match1v1Fastest` with only a `Match1v1` rating). Suggest validating `modes ⊆ ratings.keys()` in `create_entry_and_update_modes` — or deriving `modes` from `mode_ratings` keys, which eliminates the inconsistency class entirely since `QueueRequest` carrying both is redundant.

**Low**

- `matchmaking-api.ts` — the Joi schema for the preferences array has no uniqueness or length cap. Duplicates mostly collapse harmlessly, but for a user with no MMR row, `Promise.all` over duplicate types runs two concurrent `createInitialMatchmakingRating` inserts for the same (user, type, season) — no `ON CONFLICT`, so the request 500s. Add `.unique((a, b) => a.matchmakingType === b.matchmakingType)` and `.max(ALL_MATCHMAKING_TYPES.length)`.
- `find-match.tsx` — a selected type that becomes disabled (matchmakingStatus update) can't be deselected (`onClick` is `undefined` when disabled) and `handleFindMatch` doesn't filter by enabled status, while the server rejects the entire request if *any* type is disabled — so every attempt errors until remount. Auto-prune disabled types from the selection, filter them at submit, or keep disabled rows deselectable.

Verified non-issues: decline semantics (decliner fully removed, innocent accepters requeued with all modes/ratings/original queue_time preserved), per-mode rating selection end-to-end (no cross-contamination; the new Rust test covers it), the `queueStatus` shape change (only client consumer checks truthiness), and protocol compat (the whole matchmaker2 stack is unreleased and ships as a unit). The devonly test page is a nice touch.

PR #1317, opened 2026-06-25
Compute matchmaker latency from rally-point pings

2Pacalypse-:
## Why

The matchmaker's quality formula penalizes a match by `WEIGHT_LATENCY * max_latency`, but the input was a per-player `latency_bucket` that the Node side always sent as `null` — so the term was inert. It also couldn't be made to work as designed: **latency isn't a property of a single player.** A game's latency depends on which rally-point relay gets chosen, and that's picked per *pair* by minimizing the players' combined pings (`RallyPointService.createBestRoute`). Two players who each ping their local servers great can still only share a far-away server once matched together.

This was flagged in review on the telemetry PR (#1316): *"you need to pass the pings from the rally-point servers and then calculate the server that would be chosen for a particular match."*

## What

Replace the bucket with the raw signal and move the computation into the matchmaker:

- **`Player.server_pings`** (round-trip ms per rally-point server) replaces `latency_bucket`. Node reads them from the new `RallyPointService.getPingsForClient` at queue time and passes them through; they're preserved in the requeue ticket.
- **`match_latency`** reproduces launch-time route selection: for each pair pick the server with the lowest combined ping, take that pair's one-way latency (`combined / 2`), and use the max across pairs (a lockstep game is bottlenecked by its slowest link). For 1v1 this is exactly the server that would be chosen.
- **`WEIGHT_LATENCY`** switches from per-bucket to per-ms (`0.1`, deliberately gentle: it re-orders viable matches rather than blocking distant players from matching). The `max_latency` telemetry now records real estimated ms — the calibration target this work is about.

## Tier 3 (latency half)

The telemetry PR deferred *"Tier 3 (live latency buckets + sampled queue snapshots)"*. The capture half is now done (`max_latency` is a real estimate instead of always-zero), so this also adds the **analysis** half: a **"predicted vs. actual match latency"** calibration query that joins the matchmaker's queue-time estimate against the launch-time route latency already stored in `games.routes` (same one-way-ms units, same per-pair selection), so `WEIGHT_LATENCY` can be tuned from data. The unrelated "sampled queue snapshots" half stays in the backlog.

## Verification

- `cargo fmt` / `cargo clippy --all-targets --workspace -D warnings` clean; 19 Rust matchmaking tests pass (rewrote the two bucket tests as ping-based; added `latency_picks_lowest_combined_server`, `latency_no_shared_server_has_no_penalty`, and a 2v2 `latency_team_mode_uses_worst_pair_across_all_pairs`).
- The calibration example's `pg_validate()` still passes, confirming the prototype scorer reproduces the production matcher with the new latency math.
- `tsc` + `eslint` clean; 14 TS service tests pass (added one asserting pings are read from the rally-point service and forwarded).
- **Live integration:** drove the running Rust matchmaker over HTTP with two players whose crafted multi-server pings made server 1 (combined 120 → one-way 60ms) the right pick over server 2 (which player B individually prefers). The published `matchFound` carried `"maxLatency":60.0` exactly, and the requeue ticket round-tripped `serverPings` — proving per-pair selection through the real serde/search-loop/Redis path.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/server-rs/src/matchmaking/api.rs
+++ b/server-rs/src/matchmaking/api.rs
@@ -88,8 +88,9 @@ struct QueueRequest {
     mode_ratings: Vec<PlayerModeRatingDto>,
     /// The player's most recent round-trip pings (ms) to each rally-point server, as
     /// `[server_id, ping]` pairs. Used to estimate a candidate match's latency (see
-    /// [`crate::matchmaking::matchmaker::match_latency`]). May be empty if the client hasn't
-    /// measured any pings yet.
+    /// [`crate::matchmaking::matchmaker::match_latency`]). The Node.js side waits for the client to
+    /// measure its pings before enqueuing, so this is normally populated; an empty list is tolerated
+    /// and yields no latency penalty.
     #[serde(default)]
     server_pings: Vec<(u32, f32)>,
 }
--- a/server-rs/src/matchmaking/api.rs
+++ b/server-rs/src/matchmaking/api.rs
@@ -86,7 +86,12 @@ struct QueueRequest {
     /// Per-mode ratings. One entry per queued mode; the set of modes the player queues for is
     /// derived from these entries.
     mode_ratings: Vec<PlayerModeRatingDto>,
-    latency_bucket: Option<u8>,
+    /// The player's most recent round-trip pings (ms) to each rally-point server, as
+    /// `[server_id, ping]` pairs. Used to estimate a candidate match's latency (see
+    /// [`crate::matchmaking::matchmaker::match_latency`]). May be empty if the client hasn't
+    /// measured any pings yet.
+    #[serde(default)]
+    server_pings: Vec<(u32, f32)>,
 }
 
 #[derive(Debug, Clone, Deserialize, Serialize)]
@@ -103,7 +108,9 @@ struct QueueTicket {
     /// modes) on requeue.
     mode_ratings: Vec<PlayerModeRatingDto>,
     queue_time: u64,
-    latency_bucket: Option<u8>,
+    /// Per-server pings preserved from queue time, as `[server_id, ping]` pairs. Reused as-is on
+    /// requeue rather than re-measured (a requeue is rare and the pings are still a good estimate).
+    server_pings: Vec<(u32, f32)>,
     process_token: Uuid,
 }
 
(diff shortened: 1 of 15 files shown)


server-rs/src/matchmaking/api.rs
```diff
@@ -86,7 +86,12 @@ struct QueueRequest {
     /// Per-mode ratings. One entry per queued mode; the set of modes the player queues for is
     /// derived from these entries.
     mode_ratings: Vec<PlayerModeRatingDto>,
-    latency_bucket: Option<u8>,
+    /// The player's most recent round-trip pings (ms) to each rally-point server, as
+    /// `[server_id, ping]` pairs. Used to estimate a candidate match's latency (see
+    /// [`crate::matchmaking::matchmaker::match_latency`]). May be empty if the client hasn't
```
tec27: imo we should probably not let clients enter matchmaking without having measured latency now, since this is a factor in finding matches. It doesn't take *that* long to measure either

server-rs/src/matchmaking/matchmaker.rs
```diff
@@ -32,7 +35,12 @@ https://www.youtube.com/watch?v=Q8BX0nXfPjY
 
 const WEIGHT_RATING_VARIANCE: f32 = 0.005;
 const WEIGHT_WIN_PROB: f32 = 50.0;
-const WEIGHT_LATENCY: f32 = 5.0;
+/// Seconds of wait time we'll trade for a 1ms improvement in a match's estimated one-way latency.
```
tec27: I'm not sure why you did this over what I had designed. It does not make sense to look at latency on a millisecond by millisecond basis, it only changes anything if they cross a threshold such that we have to pick a lower turnrate or longer buffer. Anything under 125ms is identical, e.g. 30ms is no better than 100ms

My sheet calculates the latency value as: `MAX(CEILING(latencyMs * 0.9 / 30), 3) - 3`
And uses a weight of 30

server-rs/src/matchmaking/matchmaker.rs
```diff
@@ -84,6 +96,34 @@ fn effective_rating(player: &Player, mode: MatchmakingType) -> f32 {
...
 }
 
+/// Estimates the one-way latency (ms) of a candidate match by reproducing the route selection the
+/// game server performs at launch. The game meshes players with one rally-point route per pair,
+/// choosing for each pair the server that minimizes their combined round-trip ping (see
+/// `RallyPointService::createBestRoute` on the Node.js side); that pair's estimated one-way latency
+/// is `combined / 2`. A lockstep game is bottlenecked by its slowest link, so the match's latency
+/// is the maximum across all pairs.
+///
+/// Pairs that share no commonly-pinged server are skipped — their latency is unknown, not infinite
```
tec27: This makes it especially crazy to accept people who do not have any pings registered :)

tec27 (changes requested):

PR #1342, opened 2026-07-05
Add Twitch integration

2Pacalypse-:
Twitch integration: link a Twitch account, track who's live, and surface live streamers on the home page and profiles. Built as four stacked commits.

## Commits
1. **Add Twitch account linking** — OAuth linking, surfaced in a new "Connections" settings page.
2. **Add Twitch EventSub live-stream tracking** — subscribes to stream online/offline and records live state.
3. **Add Twitch live-streams feed to the home page** — a "Live streams" section next to live games / leagues.
4. **Add Twitch channel link and live badge to user profiles** — channel link + "Live" badge/avatar ring.

## Design notes
- **Auth**: ShieldBattery authenticates with a bearer JWT, not a cookie, so a classic server-side OAuth redirect can't identify the user on the callback. Linking is client-driven: the client opens the Twitch authorize URL, captures `code`/`state`, and hands them to an authenticated GraphQL mutation that does the secret exchange server-side. On the web this uses a popup + `/twitch/callback` relay; in the desktop app (`window.open` is denied there) a main-process `BrowserWindow` runs the flow via a `twitchOauthFlow` IPC.
- **Ownership**: entirely on server-rs (GraphQL mutations + an EventSub webhook route). The persistent link lives in a new `twitch_connections` table; the ephemeral "who's live" state lives in Redis (`twitch:live`). No user OAuth tokens are persisted — only an app token (client-credentials) plus the broadcaster's Twitch user id are needed.
- **EventSub webhook** (`POST /twitch/eventsub`): HMAC-SHA256 signature verification (unit-tested), timestamp replay rejection, message-id dedupe, challenge response, and online/offline handling. Boot reconciliation recreates missing/unhealthy subscriptions and deletes orphans.
- **Home feed filter**: StarCraft-only, matched on Twitch category **id** — `11989` (StarCraft) and `1664649323` (StarCraft: Remastered), confirmed against the live API. Ids are stable across capitalization/localization, so we match on them rather than display names. Both id and name are stored on each summary. The profile "Live" badge is category-agnostic by contrast.
- **Config**: adds optional `SB_TWITCH_CLIENT_ID` / `SB_TWITCH_CLIENT_SECRET` / `SB_TWITCH_EVENTSUB_SECRET` (all-or-none; the integration disables cleanly when unset, so dev/CI stay green). See `sample.env`.

## Verification
Statically green across the stack: server-rs build + clippy (all targets) + HMAC signature **unit tests**, sqlx query validation, `schema.graphql` + client `gql` regen, `tsc`, eslint, and `gen-translations`.

**Not yet live-verified** — the OAuth round-trip and EventSub delivery need a registered Twitch app, and the webhook needs a publicly reachable `<host>/twitch/eventsub` (localhost isn't reachable by Twitch; the Twitch CLI can simulate events locally). To bring it up: register an app at dev.twitch.tv/console with redirect URL `<SB_CANONICAL_HOST>/twitch/callback`, set the three `SB_TWITCH_*` vars, and restart server-rs.


app/app.ts
```diff
@@ -196,6 +196,82 @@ async function createScrSettings() {
...
 }
 
+/**
+ * Runs the Twitch OAuth authorization flow in a dedicated `BrowserWindow`. The renderer can't do
+ * this itself: `setWindowOpenHandler` denies `window.open` and pushes URLs to the system browser,
+ * which can't relay the result back to the app. We load the authorize URL, watch for the redirect
+ * back to our callback URL (taken from the authorize URL's `redirect_uri`), and resolve with the
+ * code/state -- or an error -- captured from it.
+ */
+function runTwitchOauthFlow(
```
tec27: I think this is *maybe* okay but requires uses to log in again (whereas their browser may already be logged in). An alternative I see a lot of apps do is to spin up an http server on localhost + do the oauth redirect to that (or possibly include enough things in the oauth state [securely] that you can redirect back to our canonical URL and it can update whatever account state) -- that is, you'd just open the auth page in their normal browser

client/users/user-profile.tsx
```diff
@@ -149,13 +153,15 @@ const TopSection = styled.div`
   align-items: center;
 `
 
-const AvatarCircle = styled.div`
+const TWITCH_PURPLE = '#9146ff'
```
tec27: imo we should pick a color for people being live in general that is not Twitch-specific, and it should go in the colors file, not here (red is the common thing people use)

server-rs/src/twitch.rs
```diff
@@ -0,0 +1,1545 @@
...
+const TWITCH_HELIX_EVENTSUB_URL: &str = "https://api.twitch.tv/helix/eventsub/subscriptions";
+
+const SUB_TYPE_STREAM_ONLINE: &str = "stream.online";
+const SUB_TYPE_STREAM_OFFLINE: &str = "stream.offline";
+
+/// Twitch category names we treat as "StarCraft: Brood War" for the home-page live-streams feed,
+/// matched case-insensitively against a stream's category. We match on name rather than Twitch's
+/// numeric game id because the ids aren't easily verifiable without live API access; both the id and
+/// name are stored in Redis, so this can be swapped for an id allowlist later.
+const STARCRAFT_CATEGORY_NAMES: &[&str] = &["starcraft", "starcraft: brood war"];
```
tec27: I'm pretty sure the IDs are constant so you should probably just pull the 2 categories (StarCraft + StarCraft: Remastered) (and this comment seems to be complaining about not having an API key so maybe you should get one of those to actually test this :P )

PR #1388, opened 2026-08-01
Add lobby settings mutability, the bench, and host move/swap

2Pacalypse-:
Stacked on #1387. Lobby redesign Phase 4: the largest new server surface before persistence — host-editable settings with slot reconciliation, the bench, and host move/swap. Ready checks (D14) are deliberately NOT included: Q3 is still open with tec27, so the 5s countdown stays.

## Settings mutability + reconciliation

The host can change map, game type/sub-type, unit limit, and observer allowance while gathering. Reconciliation on a layout change: **the host seats first** (a host can never end up benched), then players by join order, then observers displaced by an observers-off toggle, then current bench members; overflow goes to open observer slots (closed-first, `makeObserver`'s preference) and then the bench; computers re-seat last and drop if they no longer fit. Same-shape changes (map swap between identical layouts, observer toggle) leave every seat untouched — slot ids preserved, so clients see no churn. Seat pouring is **team-balanced** (via `findAvailableSlot`) rather than first-slot-first: filling team 0 to the brim on a TvB change would leave lobbies without opposing sides.

The `settingsChange` event carries the complete reconciled lobby (a rebuild doesn't diff well) plus the list of settings that actually changed; clients get a system chat message and a brief highlight on the changed values.

## The bench

Joins to a full lobby land on the bench (capped at `MAX_BENCH = 8` for joins; reconciliation never kicks, so a drastic shrink may exceed it), layout shrinks displace into it, and **whenever a joinable seat opens, the longest-waiting bench member is seated automatically** — which is also how a lobby whose last seated member leaves hands itself to whoever was waiting. Bench members are full members: they chat, hold their gameplay activity, keep the lobby alive, appear in the init user infos, and are cleanly released (activity + subscriptions) if a game launches without them.

Design calls worth reviewing:
- **Auto-seat on vacancy** isn't in the spec (D15 doesn't say); "waiting for a seat" seemed to demand it, and it resolves host handover and empty-seats-with-bench edge cases. Easy to make host-managed later if preferred.
- Join-while-full is currently reachable only via invite links/cards — the join list has always filtered full lobbies out (Phase 6a makes them visible-but-distinct).
- Bench members can't yet be kicked/banned from the UI (no menu on bench rows); server-side both work (the slot-id field carries their user id).

## Move/swap (D8)

Host moves any occupant via "Move to…" on the slot menu; an occupied destination swaps the two, with observer-boundary retyping and UMS forced-race exchange per direction. Controlled game types (team melee/FFA) allow human↔human swaps with `controlledBy` handed to the team's longest-seated remaining player. Self-move stays click-to-sit.

## Verification

- 152 lobby-area tests (reconciliation edges: shrink-to-obs closed-first, shrink-to-bench join order, same-layout identity preservation, UMS forced pour, TvB redistribution, computer drop, swap variants incl. controlled-team handoff and computer-into-obs rejection; service: bench cap, ban-from-bench, auto-seat ordering, host handover to bench) — full suite 1808 green, typecheck + lint clean.
- **Live-verified with 3 clients**: join-while-full via a whispered invite card → bench (region carried); auto-seat on leave; settings 1v1→Melee→1v1 with the third player displaced to bench (UI row + system messages on all clients); occupied-destination swap with host reference following; game launch with a benched member (seated players reached `playing`, benched member cleanly released).


server/lib/lobbies/lobby-service.ts
```diff
@@ -554,6 +599,227 @@ export class LobbyService {
...
+    }
+    updated = this._seatBenchOverflow(updated)
+
+    this.lobbies.set(updated.id, updated)
+    this._publishTo(updated, {
+      type: 'settingsChange',
+      changedSettings,
+      lobby: updated,
+    })
+    this._publishListChange('update', updated)
```
tec27, line 803: **Major:** a settings change never republishes the lobby preview. `_publishPreview` is only called from `_publishLobbyDiff` (line 1884), which this path bypasses, so the browser's preview channel keeps the pre-change seat layout while the list `update` refreshes the summary counts.

Scenario: someone in the lobby browser has this lobby selected (`lobby-list-reducer.ts` holds `previewTeams`); the host changes a 4-slot map to an 8-slot TvB. The rail shows the new `playerSlots`/`benchCount` next to the old 4-slot roster until some unrelated occupancy diff happens.

Fix: `const summary = Lobbies.toSummaryJson(updated); this._publishPreview(updated, summary)` alongside the list publish, and a service test asserting a preview publish on `updateSettings` (none of the eight current ones check it).

server/lib/lobbies/lobby-service.ts
```diff
@@ -980,23 +1303,54 @@ export class LobbyService {
...
+  /**
+   * Seats the members waiting on the bench, longest-waiting first, for as long as the lobby has
+   * both someone waiting and a slot they could have joined into directly. Also picks a new host if
+   * the lobby's is gone, which is how a lobby whose last seated member left is handed to whoever
+   * was waiting behind them.
+   */
+  private _seatBenchOverflow(lobby: Lobby): Lobby {
+    let updated = lobby
+    while (updated.bench.length > 0) {
+      const [teamIndex, slotIndex] = Lobbies.findAvailableSlot(updated)
```
tec27: **Major (design call):** `findAvailableSlot` falls back to open observer slots, so this seats a bench member as an observer whenever an observer slot opens. Combined with `applySettingsChange`'s layout-keeping branch (`lobby.ts:1142`, which pours the whole bench into the just-added, all-closed observer team) and `removeObserver` (line 1276: "Make player" takes a closed player slot, the vacated observer slot opens, and the bench fills it), three ordinary host actions convert people who joined to play into observers and drop them from the seat queue.

Probe P9: 1v1 full, one benched, host opens an observer slot for a spectator: the bench member becomes the observer immediately and the spectator lands on the bench. Probe P2: 1v1 full plus two benched, host ticks "Allow observers": both become observers in previously-closed slots, zero open observer slots remain.

This contradicts the "waiting for a seat" copy and the browser caption ("you'll wait on the bench for a seat"). If the observer-draining stays, at minimum make this loop use `findPlayerSeat` (player slots only) so an observer slot the host opens is not consumed by the bench, and keep the bench order for the next player seat.

server/lib/lobbies/lobby.ts
```diff
@@ -786,3 +921,339 @@ export function removeObserver(lobby: Lobby, slotIndex: number): Lobby {
...
+ * unseats, and finally whoever was waiting on the bench. Anyone left over goes to the observer team
+ * if there is one, and to the bench if there isn't: a settings change never removes anyone from the
+ * lobby. Computers are added back last, and only as far as the new layout has room for them.
+ *
+ * Members keep the slot ids they had, so clients can tell that someone moved rather than that one
+ * member left and another arrived.
+ */
+export function applySettingsChange(lobby: Lobby, next: LobbySettings): Lobby {
+  const keepsLayout =
+    lobby.gameType === next.gameType &&
```
tec27: **Major:** `keepsLayout` demands an identical `gameType`, so melee <-> ffa (and teamMelee <-> teamFfa at the same sub-type) goes through the full rebuild even though `getSlotsPerTeam` and `teamId` are identical. The rebuild starts from `createInitialTeams` (all open) and re-pours, including the bench (line 1163).

Probe P1: 8-slot melee, host closed 4 slots to run a 4-player lobby, flips Melee to FFA: zero closed slots remain; with anyone benched they get seated into the reopened slots.

Fix: define `keepsLayout` as same UMS-ness, same `hasControlledOpens`, and `keepsSlotsPerTeam`, instead of identical gameType/subType. No test covers a gameType change between same-shape types; worth adding one with closed slots and a non-empty bench.

server/lib/lobbies/lobby-service.ts
```diff
@@ -554,6 +599,227 @@ export class LobbyService {
...
+    }
+    if (sourceSlot === destSlot) {
+      throw new LobbyServiceError(LobbyServiceErrorCode.AlreadyInSlot, 'already in that slot')
+    }
+    const isMove = isSlotUnoccupied(destSlot)
+    if (
+      Lobbies.hasControlledOpens(lobby.gameType) &&
+      sourceSlot.type !== SlotType.Human &&
+      (destSlot.type === SlotType.ControlledOpen ||
+        destSlot.type === SlotType.ControlledClosed ||
```
tec27, line 851: **Minor:** this rejects an `observer` source into a `controlledOpen`/`controlledClosed` slot (and `swapSlots` requires both sides `human`), so in team melee/FFA the host cannot move an observer into a controlled team or swap an observer with a player. The same observer can `changeSlot` into that slot themselves (line 887) and "Make player" (`removeObserver`) puts them there, so the restriction only bites the host's move dialog, which mirrors it (`lobby-move-slot-dialog.tsx:64-72`) and greys out valid destinations.

Fix: treat `SlotType.Observer` like `SlotType.Human` in this guard and in `swapSlots`' controlled-type check; `occupantInSlot` already does the retype.

client/lobbies/lobby-reducer.ts
```diff
@@ -191,6 +192,66 @@ const lobbyHandlers = {
     draft.info.teams[teamIndex].slots[slotIndex] = player
   },
 
+  '@lobbies/updateSettingsChange'(draft, action) {
+    if (!draft.info.name) {
+      // Not in a lobby (e.g. this event trailed our own removal in a diff) - nothing to update
+      return
+    }
+
+    draft.info = castDraft(action.payload.lobby)
```
tec27, line 203: **Minor:** a settings change that shrinks the layout replaces `info` wholesale, so members it benches get no `benchAdd` (the server publishes only `settingsChange`, never a diff) and the chat shows only "The host changed the lobby settings: map". The displaced member's own client gives no notice that they lost their seat, and nobody sees the "is waiting for a seat" lines the join path produces.

Fix: diff `draft.info.bench` against `action.payload.lobby.bench` here and push a `LobbyBenchJoin` entry per newly benched user (or have the server emit `benchAdd` events alongside `settingsChange`).

server/lib/lobbies/lobby-service.ts
```diff
@@ -1453,14 +1824,14 @@ export class LobbyService {
...
-      }
+      // Everything else about a kept slot - a new position, or in-place changes like a controlled
+      // slot's controller handoff - is communicated by re-sending the whole slot
+      diffEvents.push({
+        type: 'slotChange',
+        teamIndex: newTeamIndex,
+        slotIndex: newSlotIndex,
+        player: newSlot,
+      })
       if (samePlace && oldSlot.race !== newSlot.race) {
```
tec27: **Minor:** now that every same-place slot whose object changed gets a `slotChange` (line 1829), a race change emits both `slotChange` and `raceChange` for the same slot; the client applies the whole slot and then re-assigns the race. Harmless, but this branch is dead weight. Either drop it or restrict the unconditional `slotChange` to position and `controlledBy` changes.

tec27 (commented):
**Verdict: merge after fixes, plus a rebase (182 commits behind master).** The reconciliation, swap, and bench math holds up: a 12-case probe over UMS/melee/team-melee/TvB transitions, observer toggles with the host observing, closed slots, and bench overflow found no duplicate slot ids, nobody seated twice or seated-and-benched, and the host always holding a slot. Authorization and the await-window races are handled. Ranked points, each with an inline thread: (1) `updateSettings` never republishes the lobby preview, so browser previewers keep the old seat layout; (2) bench members are silently converted to observers by three ordinary host actions (design call); (3) melee <-> ffa is treated as a full rebuild and reopens every host-closed slot; (4) `moveSlot` rejects observer moves that `changeSlot`/`removeObserver` allow.

**Rebase hazards (semantic, beyond the textual conflicts):**
- `client/lobbies/devonly/lobby-test.tsx:103` builds `Omit<LobbyProps, 'lobby'|'loadingState'|'chat'|'user'>`; master's `LobbyProps` added a required `commandContext: LobbyCommandContext`, so this literal fails `tsc` after a mechanical rebase. `view.tsx` also has to pass `commandContext` next to the new `onMoveSlot`/`onOpenLobbySettings`.
- `client/lobbies/browser/lobby-detail-rail.tsx`: this PR sets the join button to `disabled={joinRefused}`; master sets it to `... || joinPending` and adds `joinPending`/`pendingAsObserver` props. The correct merge is `disabled={joinRefused || joinPending}`. Taking either side alone is wrong.
- `server/lib/lobbies/lobby-service.ts`: master added `_deleteJoinCode(lobby.id)` in the empty-lobby branch of `_removeClientFromLobby` and in `_onGameLoaded`; both were restructured here. The cleanup has to land in the bench-empty branch too.
- `lobby-service.test.ts`: both sides added tests to the same describe blocks (mechanical). Master's `FriendActivityStatus.InLobby` registration and `regionManual` network fields sit on untouched lines and merge cleanly.

**Nits:**
- `client/lobbies/lobby-settings-dialog.tsx:90-100`: `useMemo(..., [])` plus an eslint-disable to capture once. `useState(() => ({ ... }))` expresses the same thing without the disable and matches the react-compiler guidance in AGENTS.md.
- `client/lobbies/lobby-settings-dialog.tsx:246`: `Object.values(GameType)` where `create-lobby.tsx` uses `ALL_GAME_TYPES`.
- `client/lobbies/browser/lobby-detail-rail.tsx:598-616`: the three new join captions use em dashes; repo preference is periods/commas in UI copy.

**Prettier note:** the `LobbyChangedSetting` formatting claim carried across the last nine review rounds is false on this head. The PR lockfile pins prettier 3.9.5 at top level, `eslint-plugin-prettier` resolves 3.9.5 (checked with `require.resolve`), and 3.9.5 accepts the two-line union; only the transitive 3.8.4 binary rejects it. `eslint` on all 30 changed files exits 0.

**Tooling run on the head:** `tsc` clean; `vitest run server/lib/lobbies client/lobbies common/lobbies` 316/316 passing across 14 files; `eslint` on every changed file clean; `gen-translations` (app and email) produces no drift. All 34 previously resolved inline threads were re-verified against this head and are genuinely fixed; none is re-opened.

---
_Adversarial review by Claude (Fable 5.1) at tec27's request: one reviewer per open PR, each on a worktree at this head (22cf6726e) with typecheck, eslint, vitest and gen-translations run._

PR #1389, opened 2026-08-01
Make lobbies persist through their games and regroup after.

2Pacalypse-:
Stacked on #1388. Lobby redesign Phase 5, the headline: **a lobby exists until it's empty; games are things a lobby does.**

## Server

- Starting a game no longer deletes the lobby: it enters an `inGame` lifecycle (carried on summaries — list, invite cards, the unauthenticated landing and link unfurl, which say "in game" instead of counting open seats), members stay subscribed and activity-registered (still blocked from matchmaking/second lobbies mid-game, which is intended), and the lobby goes back **onto** the list (an `add`, not an `update` — the countdown `delete`d it) marked in-game. Joins during the game always land on the bench, even with open seats; bench seating is deferred to regroup. Join codes outlive the game and die with the lobby.
- **Game-end detection** is a small `GameLifecycleEvents` emitter in `games/` (so `lobbies/` can listen without an import cycle) with two signals: per-user (`Finished`/`Error` status reports, result submissions — idempotent per user) and whole-game (`forceReconcileGame`, the funnel for the relay session closing, the known-complete trigger, and the liveness-probe sweep — emitted before its results-exempt early-out, since a comp-stomp lobby regroups like any other). A status report only counts for a game the reporter was actually in (players and observers alike; anyone else gets a 404), and it re-asserts the member's in-lobby presence so friends see them back in the lobby rather than online-but-nowhere.
- **Regroup** (last in-game member done, by report, disconnect, or the whole-game signal): run state cleared, bench overflow seated, seats and races kept, `regroup` event with the finished game's id. Seat, race, settings, and start changes are refused while in-game (`GameInProgress` 409); chat, leaving (which can itself trigger regroup), and bench-member kick/ban still work. Race changes are also refused once the countdown has snapshotted the game config.
- **Two backstops against a lobby stuck in-game.** A member's client that drops its socket mid-game keeps its seat for the same 60s grace presence uses, and the same client reconnecting resumes it (a different client is another app instance and doesn't). And a game that never produces any end signal at all — a solo-vs-computers game has no relay session, and a crashed StarCraft reports nothing — regroups the lobby at an 8h per-lobby deadline rather than wedging it until everyone leaves.
- Incidental fix: `getListedSummaries` no longer returns counting-down/loading lobbies (the list `delete` at countdown already implied it; the lifecycle field made the inconsistency visible).

## Client

- `gameStarted` keeps the lobby state (the old handler reset it) and no longer bounces the user off the lobby route; the in-game lobby shows an "In game · elapsed" status, dims in-game members' slots, and hides host controls.
- A member whose lobby survived its game never leaves it: when the game ends they're already looking at the lobby, and the regroup drops a "Game finished" card into lobby chat with a View-results link. Only someone no longer in the lobby moves on to the results screen.
- The browser shows running games: rows get an in-game chip with the true elapsed time (off the app's shared seconds ticker), and an in-game lobby is exempt from the hide-when-full filter since it's still joinable onto the bench. The unauthenticated landing gets the lifecycle but not the elapsed time.

## Landing in #1399 instead

The series score strip and the results card with replay live in the room UI (#1399). Still deferred beyond that: per-slot load rings, the 75s partial-abort refinement, and any idle-expiry beyond the existing empty-lobby deletion. Durability caveat stands: in-memory, a server restart drops lobbies.

## Verification

- Unit suite, typecheck and lint clean. The service tests cover start-keeps-lobby, per-user end idempotence, whole-game end, regroup keeping races, join-during-game benching, in-game rejections, the disconnect grace and same-client resume, the stuck-game deadline on a fake clock, observer game-end reports, and a second countdown after regroup.
- **Live-verified end to end**: create → join → launch → both clients hard-quit (worst case: no clean reports) → relay session-close drove the regroup → both members were still on the lobby with seats kept and the results card in chat → **started a second game from the same lobby** and reached `playing` again.


server/lib/lobbies/lobby-service.ts
```diff
@@ -1327,6 +1486,10 @@ export class LobbyService {
     removalType = REMOVAL_TYPE_NORMAL,
     seatFromBench = true,
   ) {
+    // Someone who disappears mid-game is out of it as far as the lobby is concerned, or the lobby
+    // would sit `inGame` forever waiting on a client that is never coming back.
+    const wasInGame = this.runStates.get(lobby.id)?.inGameUsers.delete(client.userId) ?? false
```
tec27, line 1702: **Major:** a socket disconnect during a game removes the member immediately, with no grace. The subscribe close callback (line 690, not in this diff) calls `_removeClientFromLobby`, which drops the user from `inGameUsers` here, opens their slot, releases their activity, and calls `_maybeRegroup`.

Scenario: 2v2, 30 minutes in. Player B's SB app websocket drops for 5 s (independent of the relay link; the game keeps running). B loses the seat and the activity registration; if B was the last one still marked in-game the lobby regroups under a running game and the host gets the start button back. B reconnects (`@network/connect` clears client lobby state) and is no longer a member; when the game ends B lands on the results screen, and the seat may already have gone to a bench member. Before this PR the lobby was gone at game start, so a blip was harmless. It also means the "reconnecting client gets `runState`/`elapsedMs`" path can never fire: a reconnect is an eviction.

Master's `ActivityStatusService` (`6367ffa99`, `server/lib/users/activity-status-service.ts:19`, `IN_GAME_DISCONNECT_GRACE_MS = 60_000`) added a grace period for exactly this. Suggested fix: while a user is in `runState.inGameUsers`, defer the close-driven removal behind a grace timer cancelled when a client of the same user re-subscribes to the lobby, or skip close-driven removal for in-game members entirely and rely on the per-user/whole-game signals plus the backstop. The test "the last in-game member disconnecting regroups the lobby" (`lobby-service.test.ts:591`) currently pins the undesirable behaviour and needs to change with it.

server/lib/lobbies/lobby-service.ts
```diff
@@ -1551,30 +1716,110 @@ export class LobbyService {
...
+  /**
+   * Hands a lobby over to the game it just launched: the lobby stays exactly as it is, with its
+   * members still in it and still holding their gameplay activity, and simply reports that a game
+   * is running. It goes back to gathering once that game is over for every member (see
+   * `_maybeRegroup`).
+   *
+   * The members registered as being in the game are the seated ones (players and observers alike) —
+   * the bench takes no part in it and is free to keep filling up while it runs.
+   */
+  _onGameStarted(lobbyId: SbLobbyId, gameId: string, inGameUsers: ReadonlyArray<SbUserId>) {
```
tec27, line 1941: **Major (semantic conflict with master `6367ffa99`, #1453):** after the first game, friends see every member of this lobby as Online rather than In lobby.

On master, `registerActiveClient(userId, client, InLobby)` records an InLobby activity in `ActivityStatusService`; `GameLoader.completeLoad` (master `server/lib/games/game-loader.ts:540`) calls `setInGame`, which replaces it; the Finished/Error status report calls `clearInGame` (master `server/lib/games/game-api.ts:509`, `activity-status-service.ts:135`), which deletes it and publishes Online. Master relied on the lobby tearing down at game start (see the comment on `completeLoad`); with `_onGameStarted` keeping the registry entry, nothing re-asserts InLobby after `clearInGame`, and the eventual `unregisterClientForUser` -> `clearActivity(userId, InLobby)` on leave is a no-op. Every subsequent game from the same lobby repeats it.

Fix: on `_onUserGameEnded` (or in `_maybeRegroup`) call `activityStatusService.setActivity(userId, FriendActivityStatus.InLobby)` for each member still registered to this lobby, and cover it with a test against the real `ActivityStatusService`.

server/lib/games/game-api.ts
```diff
@@ -472,9 +474,18 @@ export class GameApi {
...
+      // means the reporter's game is over - e.g. a crash mid-game - so it is accepted rather than
+      // rejected, or the game-end signal below would never fire for clients that die instead of
+      // finishing.
+      this.gameLoader.maybeCancelLoading(gameId, user.id)
+    }
+
+    if (status === GameStatus.Finished || status === GameStatus.Error) {
+      // The reporter's game is over one way or the other, so anything waiting on them (a lobby that
+      // regroups when its game ends) can stop waiting.
+      this.gameLifecycleEvents.emit('userGameEnded', { gameId, userId: user.id })
```
tec27, line 564: **Major (semantic conflict with master):** this emit only runs after the guards above, and master's version of this handler keeps the guards this PR relaxes. Master (`6367ffa99`) still 409s on `((Launching..Playing) || Error) && !isLoadingOrRecentlyLoaded` (master `game-api.ts:514-517`) and still 404s when `maybeCancelLoading` returns false, and it deliberately runs its own per-user end hook `activityStatusService.clearInGame(user.id, gameId)` before those guards (master `game-api.ts:509`, "so that a report which arrives too late to be interesting to the game loader still ends the user's in-game state"). A merge that keeps master's guard order leaves this emit below the 409, so an `Error` posted after load never signals and the crash-mid-game gap this commit claims to close reopens.

Fix: emit exactly where master calls `clearInGame`, before the guards, or better, drop the per-user half of `GameLifecycleEvents` and let `ActivityStatusService` (master's single source of truth for per-user in-game state, including the disconnect grace) publish the per-user end; keep `gameEnded` from `forceReconcileGame`. Master's client (`client/active-game/socket-handlers.ts:179`) already reports `finished` and `unknown` as `Finished`, so the client-side change in this PR should be dropped for master's.

server/lib/games/game-result-service.ts
```diff
@@ -645,6 +651,10 @@ export default class GameResultService {
    * response or interrupt the rest of a sweep.
    */
   async forceReconcileGame(gameId: string): Promise<void> {
+    // Whatever triggered this observed the game's end authoritatively (relay session closed,
+    // known-complete, liveness probe), so announce it before the results-exempt early-out below:
+    // a lobby regroups when its game ends whether or not that game ever reconciles results.
+    this.gameLifecycleEvents.emit('gameEnded', { gameId })
```
tec27, line 840: **Major (test gap):** none of the games-side emit points are tested. `game-result-service.test.ts` and `game-api.test.ts` only gain a constructor argument: nothing asserts that this emit happens before the `isResultsExempt` early-out (the whole premise of the comp-stomp regroup), that `submitGameResults` emits only after a stored report (line 544) and not on the `AlreadyReported` paths, or that `updateGameStatus` emits on Finished and on Error after load (there is no `updateGameStatus` test in this branch at all; master has a `GameApi#updateGameStatus` describe block to extend). The description says both live-found gaps are unit-covered; only the lobby half is. The lobby tests do drive the real emitter, so these three would complete the chain.

server/lib/lobbies/lobby-service.ts
```diff
@@ -1666,6 +1911,33 @@ export class LobbyService {
     }
   }
 
+  /**
+   * Ensures that the lobby is in a state where its contents can be rearranged: not on its way into a
+   * game, and not holding the roster of one that is running.
+   */
+  ensureLobbyNotTransient(lobby: Lobby) {
```
tec27, line 2160: **Minor (still open, raised in three earlier rounds):** `setRace` (line 1078, not in this diff, anchored here on the gate it should use) only calls `ensureLobbyNotLoading`, so it is reachable while the lobby has a game in progress, and `client/lobbies/lobby.tsx:514` keeps `canSetRace` live on a slot that is dimmed and tagged "In game". That contradicts the description ("mutations are rejected while in-game except chat, leaving, and bench kick/ban") and the dimmed presentation. Either route it through `ensureLobbyNotTransient` or document the exception on the method and keep the picker enabled deliberately.

server/lib/lobbies/lobby-service.ts
```diff
@@ -219,26 +247,112 @@ export class LobbyService {
...
+    })
+    this.gameLifecycleEvents.on('gameEnded', ({ gameId }) => {
+      try {
+        this._onGameEnded(gameId)
+      } catch (err) {
+        logger.error({ err }, "error handling the end of a lobby's game")
+      }
+    })
+
+    setInterval(() => {
```
tec27: **Minor:** the stuck-game backstop is a bare `setInterval` on `Date.now()`, created per instance and never cleared (each `new LobbyService()` in the test suite leaks another one; `unref` only hides it), and it cannot be driven from a test, so the `MAX_IN_GAME_MS` regroup path has no coverage. `GameResultService` and `ActivityStatusService` take the `Clock` abstraction for exactly this; inject it here, schedule through it, and add a `FakeClock` test. Also worth a note in the constant's comment: at the 8 h mark the regroup hands the host a start button while members' clients still have `activeGame.isActive` set.

server/lib/games/game-api.ts
```diff
@@ -472,9 +474,18 @@ export class GameApi {
...
     } else if (status === GameStatus.Error) {
-      if (!this.gameLoader.maybeCancelLoading(gameId, user.id)) {
-        throw new httpErrors.NotFound('game not found')
-      }
+      // During a load, an error report cancels the load (and assigns fault; a no-op otherwise).
+      // After the load is done the game is no longer the loader's concern, but the report still
+      // means the reporter's game is over - e.g. a crash mid-game - so it is accepted rather than
+      // rejected, or the game-end signal below would never fire for clients that die instead of
+      // finishing.
+      this.gameLoader.maybeCancelLoading(gameId, user.id)
```
tec27, line 549: **Minor:** ignoring the return value means `PUT /games/<anyId>/status` with `Finished` or `Error` now answers 204 and emits `userGameEnded` for any game id, whether or not the caller was in it (previously 404). The lobby handler bounds the damage via `runState.gameId` and `inGameUsers`, but a player can post `Finished` for their own running game to drop out of `inGameUsers` and, if they are the last one, regroup the lobby under a live game. Keep the 404 when the user has no `games_users` row for the id (the loader lookup is gone after load, so check the record), and only emit for a game the user actually holds.

server/lib/lobbies/lobby.ts
```diff
@@ -177,8 +178,15 @@ function toSummarySlotCounts(
...
  * order), so callers can compare serialized summaries to tell whether a change is one the list
  * would show at all.
+ *
+ * `elapsedMs` is only carried into the result when `lifecycle` is `inGame`, since it describes a
+ * running game's duration.
  */
-export function toSummaryJson(lobby: Lobby): LobbySummaryJson {
+export function toSummaryJson(
+  lobby: Lobby,
+  lifecycle: LobbyLifecycle = 'gathering',
```
tec27: **Minor (still open, raised twice):** every production caller passes `lifecycle`, so the `= 'gathering'` default only exists to let a future call site silently report an in-game lobby as gathering. Make it required and pass it explicitly from the tests.

server/lib/lobbies/lobby-service.ts
```diff
@@ -1107,17 +1248,18 @@ export class LobbyService {
...
-    this.ensureLobbyNotTransient(lobby)
 
     const [teamIndex, slotIndex, playerToKick] = findSlotById(lobby, slotId)
-    if (!playerToKick) {
-      const benched = findBenchedTarget(lobby, slotId)
-      if (!benched) {
-        throw new LobbyServiceError(LobbyServiceErrorCode.InvalidSlotId, 'invalid slot id')
-      }
+    const benched = playerToKick ? undefined : findBenchedTarget(lobby, slotId)
+    this.ensureHostCanRemove(lobby, !!benched)
```
tec27: **Nit (still open):** `ensureHostCanRemove` runs before the `!playerToKick` check (same in `banPlayer` at 1328), so an unknown slot id during a game reports `GameInProgress` rather than `InvalidSlotId`. Check existence first.

client/lobbies/browser/browser-parts.tsx
```diff
@@ -61,3 +61,16 @@ export function HostCrown({ tabIndex }: { tabIndex?: number }) {
     </Tooltip>
   )
 }
+
+/** Formats a game's length as `M:SS`, growing to `H:MM:SS` once it runs past an hour. */
+export function formatGameDuration(durationMs: number): string {
```
tec27: **Nit (still open):** this duplicates `getGameDurationString` in `common/games/games.ts` with different padding, so the same running game reads `5:23` in the browser chip and `05:23` in the lobby view (`ElapsedTime`). Use one formatter.

client/lobbies/lobby.tsx
```diff
@@ -691,13 +718,31 @@ class LobbyComponent extends React.Component<LobbyProps & WithTranslation> {
...
+  renderInGameStatus() {
+    const { runState, t } = this.props
+    if (!runState) {
+      return null
+    }
+
+    return (
+      <InfoItem>
+        <InfoLabel as='span'>{t('lobbies.lobby.status', 'Status')}</InfoLabel>
+        <InfoValue>
```
tec27: **Nit (still open):** bare `<InfoValue>`; every other row in this block (679-713) renders `<InfoValue as='span'>`.

client/lobbies/view.tsx
```diff
@@ -96,7 +97,11 @@ export function LobbyView(props: LobbyViewProps) {
   }, [isLeavingLobby, isActiveGame])
   useEffect(() => {
     if (!isActiveGame && prevIsActiveGame) {
-      if (prevGameClientGameId) {
+      if (inLobby) {
+        // The lobby survives its own game, so land back in it rather than the results screen --
+        // its regroup message links to the results for anyone who wants them.
+        navigateToLobby(lobbyId, currentLobbyName, replace)
```
tec27: **Nit:** this effect lives in `LobbyView`, which is only mounted on `/lobbies/:id`, so when it fires the user is already on this lobby's route and the call is a self-`replace`. A member who browsed to chat during the game is not navigated anywhere when it ends (same as before the PR). Fine to keep, but the description's "auto-returned to the lobby route" only holds for the never-left case; either drop the call or move the return-to-lobby navigation somewhere that is mounted during the game.

tec27 (commented):
**Verdict: needs rework against current master.** The lobby-side state machine (run state, per-user idempotence, gameId scoping, regroup bookkeeping, bench handling) is sound, but the branch is 182 commits behind master and two master changes invalidate its assumptions: #1453 (`6367ffa99`, friend activity status, which now owns per-user in-game state with a 60 s disconnect grace) and the reworked `updateGameStatus` handler. Ranked: (1) a mid-game socket blip evicts the member and can regroup the lobby under a running game; (2) friends see members as Online instead of In lobby after the first game on merged code; (3) the `userGameEnded` emit sits below guards that master keeps, so the Error-after-load fix does not survive the merge; (4) the games-side emit points have no tests despite the description claiming coverage.

**Merge-specific notes**
- Master `_onGameLoaded` also calls `_deleteJoinCode(lobby.id)`; this PR deletes `_onGameLoaded`. The join code has to survive the game and only go away on the empty-lobby path in `_removeClientFromLobby` (master already deletes it there).
- Master rewrote `server/lib/lobbies/lobby-summary-api.ts` (`toSummaryResponse` explicit picks, `getLiveLobbyWithHost`, join-code route). Its picks do not include `lifecycle`; the landing page and OG unfurl "In game" text depend on re-adding it to master's `toSummaryResponse`.
- Master `client/active-game/socket-handlers.ts:179` already reports `finished` and `unknown` as `Finished` through a separate non-fatal retry path. Drop this PR's edit to that file (it routes `finished` through the branch that clears the game config on 409/404) and take master's.
- `GameplayActivityRegistry.registerActiveClient` now takes a `GameplayActivityStatus` argument.
- `forceReconcileGame` on master gained a `NotFound` early-return in its catch; emitting before the `try` still works (a cancelled load's late `sessionClosed` emits for a game no lobby holds and is ignored).
- Master's new end-of-game paths (`resolveGameManually`, coordinator-driven `completeLoad`, `isLocalOnlyLoad`) do not add a whole-game path the emitter misses; only the per-user placement (F3) is affected.

**Still open from earlier rounds (nits)**
- `setRace` ungated mid-game and `canSetRace` live on a dimmed slot (raised three times; see inline).
- `toSummaryJson` still defaults `lifecycle = 'gathering'` (raised twice).
- Kick/ban of an unknown slot id during a game reports `GameInProgress` instead of `InvalidSlotId`.
- `formatGameDuration` duplicates `getGameDurationString` with different padding (`5:23` vs `05:23`).
- `renderInGameStatus` uses a bare `<InfoValue>` where every sibling row uses `as='span'`.

Tooling in a worktree at c4d3f6414: `tsc` clean, `vitest run server/lib/lobbies server/lib/games client/lobbies client/active-game common/lobbies` 24 files / 531 tests green, `eslint` clean on all 40 changed TS files, `gen-translations` produces no drift.

---
_Adversarial review by Claude (Fable 5.1) at tec27's request: one reviewer per open PR, each on a worktree at this head (c4d3f6414) with typecheck, eslint, vitest and gen-translations run._

Issue #1269, opened 2025-12-24
User stats


2Pacalypse-:
This issue should be used as a spec for implementing user stats (sub)page. First and foremost that means defining _what_ data that page should include and _how_ should it look like. After that is defined we can also discuss some potential implementation details.

But for now, here's an example of what the user stats page could look like.

Screenshots:

<img width="1919" height="1042" alt="Image" src="https://github.com/user-attachments/assets/4d2789fb-d826-4473-a603-6394989579dc" />

<img width="1920" height="1034" alt="Image" src="https://github.com/user-attachments/assets/b57bbce8-2e73-418f-9b33-4ecfd325fce9" />

Video:

https://github.com/user-attachments/assets/41db0e26-1dc3-4848-8f28-6e4d06f973e5

2Pacalypse-:
Ok, updated the post with the new version (the old version is still available in the post, just commented out so Github doesn't render it).

tec27:
Yeah this looks nicer, I mostly like this (maybe the Rivalries/Enemy Teams could be a 2 or 3-up card layout instead of a list for wider screens but yeah :) ). And then we'd have to figure out how to calculate all these things

tec27:
Yeah I think that looks pretty good 👍

ExcalZ:
I've put the Rating section up as a PR (#1391). Before starting on **Rivals** I measured what it
would actually cost, because it's the one section on the mock whose cost isn't obvious from
looking at it. It turns out to need a schema decision, so I'd rather raise it here than pick an
answer myself.

### Setup

1M games / 2.8M `games_users` rows / 20k players, seeded into a dev database. Target user has
12,600 games against ~4,000 distinct co-participants, so the long tail that makes a "top rivals"
list meaningful is present rather than a handful of buckets. Postgres 17 on stock JIT settings.

### Numbers

| | |
|---|---|
| Self-join alone — "who else was in my games" | **12.8 ms** |
| Full Rivals, teams read from `games.config->'teams'` | **291 ms** |
| …same, `jit = off` | **71.7 ms** |
| …same, rewritten as a single pass (no self-join on the expansion) | **70.6 ms** |
| …same, off a denormalized `team` column | **~20 ms** |

### The join is fine; resolving teams isn't

The self-join is already well served by existing indexes — `games_users_pkey` leads with
`user_id` for "my games", and `games_users_game_id_index` covers "who else was in this game."
[...]

PR #1390, opened 2026-08-01
Add a dev page prototyping the profile stats sections

ExcalZ:
Adds a devonly page at /dev/users/stats that mocks up the three sections proposed for the profile Stats tab: Rating, Rivals and Matchups. It runs on generated data, so nothing here is wired to real queries yet -- it exists to make the layout and the data shape reviewable before the server work starts.

Rating lists matchmaking modes ordered by lifetime games played, with the most-played expanded and the rest collapsed. Each mode charts rating or points over time. Two details the charts have to get right: the series is split into separate lines at each discontinuity, since rating only carries across a season that did not reset it while points restart every season, and the points view underlays ladder division bands. Those bands take their bounds from getDivisionsForPointsChange and their bonus pool from getTotalBonusPool rather than a local copy of the table, so they cannot drift away from the matchmaker. Band colours are sampled from the rank badge art.

Matchups renders a race matchup matrix filtered by mode, season and map. Both axes are unordered compositions, so a team is identified by the races on it rather than by who played which: 3x3 at 1v1, 6x6 at 2v2, 10x10 at 3v3. Cells under five games are de-emphasised and a coverage line reports how much of the grid a player has actually filled, which is the honest answer to whether the team matrices are worth building -- at 3v3 a 132-game player leaves 41 of 100 combinations empty and tops out at 8 games in the best one.

Adds recharts (3.x, the React 19 compatible line) as the charting library.

Refs #1269

### DIFF
--- a/client/dev.tsx
+++ b/client/dev.tsx
@@ -12,6 +12,7 @@ import { DevPageMetadata } from './page-metadata/devonly/routes'
 import { DevReplays } from './replays/devonly/routes'
 import { DevStarcraft } from './starcraft/devonly/dev-starcraft'
 import { DevTwitch } from './twitch/devonly/dev-twitch'
+import { DevUsers } from './users/devonly/routes'
 
 const Container = styled.div`
   width: 100%;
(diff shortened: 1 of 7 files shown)


ExcalZ:
Closing in favour of #1391, which combines both as you suggested.

I also moved the real code out of the dev page there. You were right that it had too much in
it -- it had grown its own `RatingChart`, its own mode card, and its own copies of
`splitOnDiscontinuity`, `bandOpacity` and the season band grouping, and those had already
drifted from the versions the real page uses. The dev page is down to constructed data and a
scenario picker; the components it renders are the ones that ship.

PR #1391, opened 2026-08-02
Chart rating history on the profile Stats tab

ExcalZ:
Adds the profile Stats tab, with the **Rating** section wired to real data, replacing the
coming-soon placeholder.

Supersedes #1390, which I've closed — this is the combined PR @tec27 asked for.

## Where the code lives

@tec27's other point was that the dev page had far too much in it, and that anything real
should sit next to the real page so the two share it. That was more true than it looked: the
dev page had grown its own `RatingChart`, its own mode card, and its own copies of
`splitOnDiscontinuity`, `bandOpacity` and the season band grouping. They'd already drifted —
the dev `splitOnDiscontinuity` took an `everyBoundary` boolean and closed over a module-level
`SEASONS`, so the chart's most correctness-sensitive decision existed twice in two shapes.

Real components and types now live beside the real page:

| | |
|---|---|
| `mode-stats-card` | the collapsible card; takes history as props rather than fetching |
| `rating-chart` | the chart itself |
| `rating-chart-data` | its shaping logic — where lines break, which bounds a band uses |

`user-profile-stats` keeps the queries and hands results to the card. The dev page is down from
~1,230 lines to ~215: constructed data, plus a ready/loading/error picker for the card. Those
last two states otherwise need a slow or broken backend to look at.

The Stats tab is lazy-loaded, so recharts only downloads for visitors who open it. In the Vite
build that's a 352 kB / 104 kB gzipped `user-profile-stats` chunk kept out of a 533 kB main
bundle.

## Server

`userRatingHistory(userId, matchmakingType)` and `matchmakingSeasons`, plus
`userSeasonDeltas(userId)`.

`matchmaking_rating_changes` is already one row per game, but it carries no `season_id`, so
seasons are derived by bucketing `change_date` against `matchmaking_seasons.start_date`. The
client needs that split: points restart every season, and MMR restarts whenever a season sets
`reset_mmr`, so a chart drawing either as one continuous line shows a fall the player never
took.

Histories past 400 points are downsampled with largest-triangle-three-buckets, chosen over
every-Nth sampling or time-bucket averaging because it preserves shape: the peak is the part
of a rating curve players care about. Downsampling runs per season and never across a
boundary, since a bucket spanning one would emit a point between two disjoint rating bands.
Endpoints are always kept so the chart agrees with the headline numbers.

`userSeasonDeltas` compares the current rating against the last rating held *before* this
season started, deliberately spanning the reset — after a reset everyone starts from the same
place, so a within-season figure mostly measures how much you've played. It's a separate
query because every mode's header shows this, including collapsed cards whose history is
never fetched.

Also adds `getAllDivisionsWithBounds` to `common/matchmaking`, replacing an enumeration that
went through `getDivisionsForPointsChange(solo, 0, MAX_SAFE_INTEGER, …)`. That only resolved
to Bronze1..Champion because Champion's high bound is `Infinity`; had it ever become finite,
the call would have returned `[Unrated, Bronze1]` and the bands would have silently become one
flat wash over the chart.

## Client

One card per ranked mode, most-played first and open by default. The mode list and header
figures come from the profile's existing ladder data rather than a new request —
`getRankedTypesByActivity` already orders modes by games played. A mode's history is fetched
only when its card is expanded.

The chart splits the series at each discontinuity, and the points view underlays division
bands per season at the bonus pool that applied during that season. Bounds and colours come
from `getAllDivisionsWithBounds` and `getDivisionColor` rather than local copies. Over an
all-time range the x-axis is labelled by season roll rather than evenly spaced dates.

Expanding and collapsing animates the body's height. It's driven from state rather than
`AnimatePresence`: with react-compiler, `AnimatePresence` doesn't appear to observe the child
swap, so its exit animation never runs.

The shaping logic — where a line breaks, which bounds a band uses, where the ticks go — is in
`rating-chart-data.ts` rather than the component, so it can be tested without a measured
container. Those decisions are the ones that go wrong in ways that still look plausible on
screen, so they're the ones worth pinning down.

## Verification

Rating history was inserted directly into a dev database; no real games were played against
this. The seed covers seasons that reset MMR and seasons that don't, so the discontinuity
handling is exercised rather than assumed.

For read cost I seeded 1M rows across 2000 users, since at toy sizes the planner makes choices
it won't make in production. At that size it uses the primary key (`user_id, matchmaking_type`)
and returns a 12,600-game history in 3.8 ms; end-to-end that query is 59.2 ms median, against
7.4 ms for deltas and 2.9 ms for seasons. The downsampled payload for that history is 390
points across 17 seasons, in order, with endpoints preserved and allocation proportional to
each season's game count. Nothing here wants a new index.

`cargo check` covers the SQL via sqlx's compile-time verification. Unit tests cover season
bucketing, endpoint preservation, season isolation, and spike survival on the server, and the
chart's splitting, band bounds, and axis ticks on the client.

## Scope

**Rating** only. Rivals and Matchups are out of this PR entirely per review — they had no
queries behind them, and they'll be built against real data and the real componentry when their
schema questions are settled.

The dev page keeps the Rating harness, whose loading and error scenarios otherwise need a slow
or broken backend to see.

I did measure Rivals first, against 1M games / 2.8M `games_users` rows. The self-join itself is
cheap and already indexed — `games_users_pkey` covers "my games" and `games_users_game_id_index`
covers "who else was in this game" — at 12.8 ms for a 12,600-game history. What costs is
ally-vs-opponent: `games_users` has no team column, so it has to come from
`games.config->'teams'`, which means fetching a config row for every game the user has played to
recover one bit per participant. That's 72 ms warm and materially worse cold, and it isn't a
query-shape problem — a single-pass rewrite measured identical. A `team smallint` on
`games_users` takes it to ~20 ms. That's a schema change rather than something to assume here,
so the numbers are written up in #1269.


client/users/mode-stats-card.tsx
```diff
@@ -0,0 +1,288 @@
...
+  font-variant-numeric: tabular-nums;
+`
+
+const Delta = styled.div<{ $positive: boolean }>`
+  ${labelMedium};
+  font-variant-numeric: tabular-nums;
+  color: ${props => (props.$positive ? 'var(--theme-positive)' : 'var(--theme-negative)')};
+`
+
+const Caret = styled(MaterialIcon).attrs({ icon: 'expand_more' })<{ $open: boolean }>`
```
tec27: need to use `styledWithAttrs` for these

client/users/rating-chart.tsx
```diff
@@ -0,0 +1,174 @@
...
+  RatingChartSeason,
+  RatingMetric,
+  bandOpacity,
+  buildSeasonBands,
+  seasonAxisTicks,
+  splitOnDiscontinuity,
+  valueExtent,
+} from './rating-chart-data'
+
+export type { RatingChartPoint, RatingChartSeason, RatingMetric } from './rating-chart-data'
```
tec27: not a fan of re-exports, just have things import them from the other file

client/users/rivals-section.tsx
```diff
@@ -0,0 +1,215 @@
+import { useTranslation } from 'react-i18next'
```
tec27: If we're not shipping a real version of this it shouldn't be included in this PR. This has a bunch of code that seems to not use our actual componentry (e.g. avatar, etc.) so we need to not do that, but it will be easier to use the proper thing if it is actually hooked up

tools/sqlx-prepare.mjs
```diff
@@ -0,0 +1,38 @@
+#!/usr/bin/env node
```
tec27: Not sure why this was added, the syntax in `pnpm run sqlx-prepare` worked fine because pnpm has shell emulation

package.json
```diff
@@ -122,6 +122,7 @@
     "pino-tee": "^0.4.0",
     "prom-client": "^15.1.3",
     "pug": "^3.0.4",
+    "recharts": "^3.10.0",
```
tec27: this is client-only so it should be in devDependencies

client/users/user-profile.tsx
```diff
@@ -62,6 +61,7 @@ import { ExpandableRankDisplays } from './expandable-rank-displays'
 import { ConnectedMatchHistory } from './match-history'
 import { MiniMatchHistory } from './mini-match-history'
 import { UserProfileSeasons } from './user-profile-seasons'
+import { UserProfileStats } from './user-profile-stats'
```
tec27: we should probably lazy-load this component since recharts is (I'm sure) a large library

tec27 (changes requested):

tools/sqlx-prepare.mjs
ExcalZ: It does have shell emulation and it is enabled (`pnpm config get shell-emulator` → `true`), but its shell doesn't implement the constructs this script used. On Windows, pnpm 11.18.0:

| what I ran | result |
| --- | --- |
| `[ "${NODE_ENV:-}" != "production" ] && echo A \|\| echo B` | `command not found: [` → `B` |
| `test "${NODE_ENV:-}" != "production" && echo A \|\| echo B` | `command not found: test` → `B` |
| `if [ … ]; then … ; else … ; fi` | `command not found: fi`, exit 127 |

So the original always fell through to the `||` branch: it printed "In production, not sqlx-preparing" and exited 0 in dev *and* production, never running cargo. It also swallowed failures — I replaced the cargo call with `(exit 3)` and the script still exited 0.

CI doesn't catch it because the `sqlx-prepare` job runs `cargo sqlx prepare --check` directly in `./server-rs` rather than going through the pnpm script, so it only bites on a Windows dev machine. That's how I ran into it: `pnpm run sqlx-prepare` looked like it succeeded while quietly writing nothing.

Happy to solve it differently if you'd rather not have the extra file — `cross-env-shell`, or dropping the `NODE_ENV` guard entirely, would both work. I went with a script mostly because it also fixes the swallowed exit code.

ExcalZ:
Pushed in 7d48c73. Thanks — all six addressed:

- **Rivals / Matchups removed.** Agreed on both counts, and the same argument applied to Matchups so that went too. The dev page keeps only the Rating harness, since its loading and error scenarios otherwise need a slow or broken backend to see. When these come back they'll be built against real queries and the real componentry — Rivals is still waiting on the schema question in #1269.
- **`styledWithAttrs`** on the caret.
- **Re-export dropped** from `rating-chart`; callers take those types from `rating-chart-data` directly.
- **`recharts` → devDependencies.**
- **Stats tab lazy-loaded.** You were right about the size — it's the biggest chunk in the build at ~4 MiB, and every profile visitor was paying for it regardless of which tab they opened. Follows the `LoadableAdminUserPageLayout` pattern already in that file.
- **`sqlx-prepare`** — replied on that thread with what I measured; short version is the emulator is on but doesn't implement `[`, `test` or `if`, so the old script no-opped in both dev and production on Windows and swallowed cargo failures too. Glad to take a different fix if you'd prefer.

tools/sqlx-prepare.mjs
ExcalZ: You're both right and I was wrong — sorry for the noise. Reverted in 517ede3, so `sqlx-prepare` is back to exactly what it was and the extra file is gone.

What I missed: `@yarnpkg/shell` doesn't implement `[` as a builtin, it resolves it from `PATH`. On Linux/macOS that finds `/usr/bin/[`, so it works. On Windows it works too, as long as Git's `usr\bin` is on `PATH` — I confirmed that by putting it on `PATH` and re-running:

| | result |
| --- | --- |
| dev | `WOULD_RUN_CARGO` ✅ |
| `NODE_ENV=production` | `WOULD_SKIP` ✅ |

Git for Windows doesn't add `usr\bin` to `PATH` in its default install, which is the only reason I saw it fail. Environment, not the script. I should have checked that before asserting it was broken.

One unrelated thing I did notice while testing, take it or leave it: because the shape is `cond && cargo … || echo`, a **cargo failure** lands on the `||` branch, so the script prints "In production, not sqlx-preparing" and exits 0. I simulated it by swapping the cargo call for `(exit 3)` and the script still exited 0, with that message. That one isn't platform-specific. Not worth anything in this PR — happy to open a separate issue if you think it's worth a look, otherwise ignore it.

ExcalZ:
Merged master in (b6ee18b) — the Vite change conflicted on `pnpm-lock.yaml`. I resolved it by taking master's lockfile and re-running `pnpm install` rather than hand-editing a generated file. Merge rather than rebase so your review threads stay anchored.

Also correcting a number I gave above: I said recharts was "~4 MiB", which came from the **webpack dev** output and isn't representative. In the production Vite build it's a **352 kB / 104 kB gzipped** `user-profile-stats` chunk, against a 533 kB main bundle. Still worth splitting out — it's roughly a 40% addition to the entry bundle for a tab most visitors never open — but an order of magnitude less than I implied. The build does confirm the split lands as its own chunk.

ExcalZ:
Came back to the automated review above — I'd replied to @tec27's comments without reading it properly, and two of its findings were real. Fixed in f3b73a5.

**1. Reset in a skipped season didn't break the line.** Confirmed against `getMatchmakingRating` (`server/lib/matchmaking/models.ts`), which walks seasons newest-first and `break`s on `row.reset` whether or not the player has a rating in that season — so a reset voids everything older regardless. `splitOnDiscontinuity` only tested the season being arrived in, so playing S1, sitting out a resetting S2, and returning in S3 joined S1→S3 into one run and drew exactly the fall the split exists to prevent. It now checks every season in the gap. My existing tests all had games in every season, which is why none of them caught it — there's one for it now, and it fails against the old code.

**2. Downsampling was selecting on rating only.** Also real. `lttb` used `p.rating` for the bucket average, `point_a_y`, and the triangle area, and the client charts *both* metrics from that one array. Points restart each season while rating carries across, so the two curves aren't correlated and the points view was getting a series sampled to preserve a different shape — only past the 400-point threshold, i.e. exactly the histories that need it. `lttb` now returns indices and takes a value accessor; each metric gets half the budget and the union is kept, so the result still fits the budget. Two tests: one that a spike in *either* metric survives, one that the budget is respected.

The packaging and scope points it raised were already handled while addressing @tec27's review — recharts moved to `devDependencies`, and Rivals/Matchups (plus their i18n keys and the `ALL_MAPS` sentinel it flagged) are out of the PR entirely.

Also took its two minor notes: dropped the `useMemo` calls in `rating-chart` per AGENTS.md, and stopped rendering the load error inside a styled component called `Downsampled`.

ExcalZ:
Thanks — four fixed in 4ae067f and 40c28ce, one pushed back on.

**1. Suspense — fixed.** Real, and worse than it reads: the boundary it throws to is the one added when the tab was lazy-loaded, so expanding any card blanked the whole Stats tab and took the expand animation with it. `context: { suspense: false }`, matching the Twitch query on the same page.

**2. Division bands — working as intended, not changing.** Bands within a season being flat is deliberate. The sloped version was built and rejected: bands that expand across the season read as the *player* moving when they haven't, and the boundaries drifting under the line made the chart much harder to read at a glance. The tradeoff you describe is real — a genuinely Champion week-2 game can be drawn low — but a stable horizontal reference beat an accurate moving one when both were on screen. The line and its endpoint are what players read; the bands are there to say roughly where the tiers sit.

**3. Label — fixed, computation kept.** You called this right: the number is the one we want (comparing against where you finished, deliberately spanning the reset), but "this season" describes something else. It now reads "vs. last season", so the reset case no longer shows a climbing player a large loss.

**4. Duplicate seasons API — fixed.** Simpler than suggested, even: `user-profile.tsx` already reads `s.matchmakingSeasons.byId` and passes it to `SummaryPage` a few lines from where the Stats tab renders, so the Stats tab now takes the same prop. The resolver, the GraphQL type and the per-mode refetch of the whole list are gone.

**5. AnimatePresence comment — fixed.** You're right, and I should have checked before writing it. Exit animations work in 18 other files here. The comment now says what the code does — the body stays mounted so it can be measured for the `height: auto` collapse — with no claim about the compiler.

**Minor (budget overshoot) — fixed, and thank you for it.** It was real. Worth noting my first attempt at it was wrong twice: the fix didn't bound anything (a season smaller than its budget returns all its points regardless), and the test passed against the *unfixed* code because the test helper sets `points = rating * 4`, so both LTTB passes pick identical indices and the union masks the overshoot. With uncorrelated data it produces 487 for a 400 budget. Now the per-season floor is reserved before the remainder is shared out, and `lttb` honours a 2-point budget as endpoints — 387 for the same input.

ExcalZ:
Both real, both verified against the source. Fixed in 751ec7f and 083c690.

**1. Full season list missing on this route — my regression, thank you for catching it.**

This one is mine from the previous round. Dropping the `matchmakingSeasons` resolver was right, but I moved the chart onto `s.matchmakingSeasons.byId` assuming the list was already there. `user-api.ts:678` returns `seasons: [toMatchmakingSeasonJson(currentSeason)]` — one entry — and `/ladder` is the only page that fetches the rest, exactly as you describe.

The part I find worth noting: nothing errors. The `crossedReset` guard I'd added defensively two rounds ago — "an unknown season shouldn't happen, break the line rather than risk drawing a fall" — turns a missing season list into a line that breaks at *every* boundary. My own safety net converted this into the precise bug the split exists to prevent, silently.

Fixed by dispatching the existing cached `getMatchmakingSeasons` rather than putting the list back into GraphQL, since not duplicating that API was @tec27's point and it still stands. The tab holds until it lands — a chart drawn on a partial list looks plausible and is wrong, which is worse than a moment of loading.

**2. Modes not played this season — fixed.**

Also correct, and the fix was cheaper than expected: `userSeasonDeltas` already grouped over every row the user has, it just discarded modes with no prior rating to compare against. It's now `userRankedModes`, returning every mode ever played with all-time totals and a nullable delta, and the card list comes from that instead of `profile.ladder`.

On your related point about the header figures: they now use current-season numbers where the player has them and fall back to all-time where they don't, because 0–0 sitting above a chart of a thousand games reads as a bug. That does mean the header's meaning varies per card, which I don't think should be settled quietly in a review thread — flagging it rather than picking. Happy to make them all-time throughout, or label them, whichever @tec27 prefers.

ExcalZ:
All three fixed in f2ded884b.

**1. Default-open card couldn't be collapsed — real, and exactly as described.** `undefined` was doing double duty: clicking the first card's header called `setOpen(undefined)` on state that was already `undefined`, so React bailed out and nothing moved, and collapsing any *other* card fell back to `modes[0]` and silently re-opened the first one along with its history fetch. Untouched is now `null` and distinct from an explicit close. You're right that the dev harness hid it — it seeds the state with a concrete mode, so the untouched case never existed there.

**2. Failed seasons fetch spun forever — also real.** `onError: () => setSeasonsLoaded(false)` set the flag to the value it already held, so the behaviour the comment above it promised was unreachable and the tab just sat there. It's `'loading' | 'ready' | 'error'` now and the error branch renders `users.profile.stats.loadError`. I reused that existing string rather than adding a key: it reads correctly for this failure too, and the alternative was a new string in five locales for a case that says the same thing.

While in there, the `NoRanked` styled div became `Message` and is shared by both tab-level messages, rather than a second identical component next to it.

**3. Shared date formatters — done.** `monthDay` for the axis, `longTimestamp` for the tooltip. The tooltip one matters more than the per-call `Intl.DateTimeFormat` construction did: a date alone labels every game in a day identically, and a day is often dozens of them.

Worth noting the tick fallback is narrower than it looks — when `ticks` is passed, every value in it came from `seasonAxisTicks` and so always hits `tickLabels`. The formatter fallback therefore only runs in the single-season view, where the span is one season and `monthDay` dropping the year is right rather than lossy. I left a comment on the axis saying so, since that isn't obvious from the call site.

**No test for (1).** It's parent-component state that would need urql, redux and i18n stood up to exercise, and there are two component tests in the whole client. The shaping logic that can be tested without rendering still is. Happy to add one if you'd rather have the coverage than the mocks.

PR #1392, opened 2026-08-02
Move web build to vite.

tec27:


tec27:
Applied `experimental.renderBuiltUrl` returning `{ relative: true }` for non-HTML host types.

Confirmed the mechanism in the built output — the preload helper changes from:

```js
var e = function(e) { return `/scripts/` + e }      // absolute, resolved against the document
var e = function(e, t) { return new URL(e, t).href } // resolved against the importing module
```

Also stood up a two-origin simulation (shell served from one port, assets from another with CORS) and confirmed no asset request reaches the shell's origin.

One caveat on the impact claim: I could not produce a runtime repro of the double-fetch. Exercising a lazy route fetches exactly one chunk with no preload deps, and I couldn't find a dynamic import in this build that emits a non-empty dep list — so the symptom may not be reachable here today. The change is still strictly more correct (it restores webpack's `publicPath: 'auto'` behaviour) and costs nothing, so I've kept it.

tec27:
Round 2 — took four of the five, left one deliberately.

**Fixed**

- **Optimizer tests** (raised twice, fair). Eight cases against a temp fixture root and a stub plugin context: aliased `graphql as gql`, fragments, relative import paths from nested files, interpolated templates, `graphql` imported from elsewhere, unknown query strings, and the name collision the aliasing exists to prevent. Also added `build-plugins/` to tsconfig `include` — only the plugin itself was reachable (transitively via `vite.config.ts`), so the test file was invisible to typed linting.
- **Missing shell build** now throws naming `pnpm run build-web-client` rather than a bare ENOENT per request.
- **svgr excludes node_modules**, restoring the old webpack rule's behaviour.
- **`browserslistToTargets` comment** now says what the code does. Checked the resolved list: Samsung Internet is the only dropped engine, at versions 29/30 (Chromium 125/127), so it sits well above the Chrome 103 floor and wasn't constraining the target. Folding it into `chrome` by number would be wrong — Samsung 29 is not Chrome 29 — so if it ever becomes binding the fix is to name it in the browserslist query.

**Not changed**

- **Dev-server file exposure.** Accurate, but restricting `fs.allow` risks breaking legitimate module serving for a dev-only, loopback-by-default surface, and it's Vite's normal posture. I did add `.env-build` to `fs.deny`, since Vite's default covers `.env`/`.env.*` but not that spelling.

Also worth noting for the earlier `ASSET_BASE` point about bundler-emitted asset URLs (CSS `url()`, `new URL(..., import.meta.url)`): the `renderBuiltUrl` change in the previous round covers those too — they're non-HTML host types, so they now resolve relative to the importing module rather than the document.

tec27:
Good catch, and that one was mine — `.env-build` was a fix I added last round that made things strictly worse.

`server.fs.deny` replaces Vite's default rather than extending it, so a one-entry list un-protected `.env`, private keys, `.npmrc` and `.git` on a server that serves the whole repo root. I added it as a "zero-risk one-liner" without checking replace-vs-extend semantics, which is exactly the assumption worth checking on a security-relevant option.

Now restates the default set (`['.env*', '*.{crt,pem,key,p12,pfx,cer,der}', '.npmrc', '.yarnrc.yml', '**/.git/**']`) with `.env*` generalizing `.env` + `.env.*` so it still covers `.env-build`. Verified against the resolved config rather than the source, and left a note to re-check on Vite upgrades since the default is now duplicated here.

tec27:
Round 4 — both inline nits fixed, and the filter one was the best catch of the review.

**Code filter vs. aliased imports.** Correct, and worse than a nit: the handler was written to follow the local binding, but the filter meant it never ran for those files. My test passed only because it called the handler directly, so it was actively giving false confidence. Fixed the filter to match the bare word, and changed the test to run the filter first the way rolldown does — it now fails against the old `graphql(` filter, which I confirmed by reverting.

**`babel-plugin-polyfill-corejs3`** dropped; the only remaining reference is the comment explaining why the approach was abandoned.

**Leftover `require()` in electron-gated branches** — confirmed safe, no change. `client/index.jsx` was converted because `module.hot` on the line below it threw at runtime under ESM; the `require` there was collateral, not the cause. Verified the remaining ones are dropped from the web bundle:

```
redux-devtools:               0 chunks
active-game/socket-handlers:  0 chunks
download/ipc-handlers:        0 chunks
system-bar/ipc-handlers:      0 chunks
```

They only survive because `IS_ELECTRON` folds to `false`, so they'll need converting when the renderer moves to Vite — that's recorded as a Phase 2 item in the migration doc rather than done here, since it can't be verified without the Electron build.

One CI note: an `integration-tests` run failed on this commit's predecessor at the *Playwright browser download* step (`cdn.playwright.dev` returned 403, "service not available in your location"), before any test executed. The sibling run on the identical SHA passed. Infrastructure, not the diff.

PR #1413, opened 2026-08-22
Remember and restore scroll positions on user profile pages

2Pacalypse-:
Adds the view-state store + `useScrollMemory` hook (Track 1, item 2 of the scroll-restoration plan) and applies it to profile pages.

- **`view-state-store.ts`**: module-level store for transient per-page view state — one shared LRU (capacity 50) with a per-namespace TTL, typed per consumer via `createViewStateStore<T>(namespace, { maxAgeMs })`. Deliberately outside React/Redux/Jotai: writes never trigger renders, and it is only touched from effects. Chat restoration can later add its own namespace with richer entries.
- **`history-entry-key.ts`**: patches `pushState`/`replaceState` to stamp a per-entry "visit key" into `history.state`, mirroring the granularity of the browser's own scroll restoration. Only entries pushed with a null state get stamped, so the app's existing sentinel values in `history.state` (settings overlay, navigation menus and traps) keep their identity comparisons. Pushes that stay on the same pathname and all replaces inherit the current key, so search-param updates, overlay entries, and in-place username spelling corrections read as a continuation of the same visit.
- **`useScrollMemory(ref)`**: saves an element's scroll position on leave and restores it pre-paint on return (30 min TTL), keyed by the current entry's visit key — traversing back/forward to an entry resumes where the user left off, while following a link to a previously visited page starts at the top like any fresh visit. It composes with the #1412 reset: the page-level restore effect fires after `CenteredContentContainer`'s internal reset, so memory-enabled pages win while everything else keeps resetting to top. The saved value comes from a passive scroll listener rather than a cleanup-time `scrollTop` read, since same-route navigations commit the new DOM before cleanups run and the live value can already be clamped by shorter content.

The Summary tab (redux-cached) restores exactly; match history/stats tabs still land at top until the game-list window cache (Track 1 item 4) lands.

Verified live in the Electron app: back/forward traversal to a profile entry restores the position exactly, while following a link to the same profile lands at the top like a fresh visit. Unit tests cover the store (LRU eviction, recency refresh on read, TTL, namespacing) and the entry-key stamping (fresh key on cross-pathname pushes, inheritance on same-pathname pushes and replaces, sentinel pass-through, popstate resync).

### DIFF
--- a/client/index.jsx
+++ b/client/index.jsx
@@ -11,6 +11,7 @@ import './dom/window-focus'
 import i18n, { detectedLocale, initI18next } from './i18n/i18next'
 import { getBestLanguage } from './i18n/language-detector'
 import log from './logging/logger'
+import { installHistoryEntryKeys } from './navigation/history-entry-key'
 import { fetchJson } from './network/fetch'
 import registerSocketHandlers from './network/socket-handlers'
 import { setServerConfig } from './server-config-storage'
@@ -22,6 +23,7 @@ window.__webpack_nonce__ = window.SB_CSP_NONCE
 
 enableArrayMethods()
 enableMapSet()
+installHistoryEntryKeys()
 
 window.addEventListener('error', event => {
   const messageText = event.error?.message ?? event.message
--- a/client/navigation/history-entry-key.test.ts
+++ b/client/navigation/history-entry-key.test.ts
@@ -1,0 +1,121 @@
+import { afterEach, beforeEach, describe, expect, test } from 'vitest'
+import {
+  getHistoryEntryKey,
+  installHistoryEntryKeys,
+  resetHistoryEntryKeysForTesting,
+} from './history-entry-key'
+
+describe('history-entry-key', () => {
+  beforeEach(() => {
+    history.replaceState(null, '', '/a')
+    installHistoryEntryKeys()
+  })
+
+  afterEach(() => {
+    resetHistoryEntryKeysForTesting()
+  })
+
+  test('getHistoryEntryKey returns undefined for null state', () => {
+    expect(getHistoryEntryKey(null)).toBeUndefined()
+  })
+
+  test('getHistoryEntryKey returns undefined for a sentinel string state', () => {
+    expect(getHistoryEntryKey('SOME:sentinel')).toBeUndefined()
+  })
+
+  test('getHistoryEntryKey returns undefined for an object without the stamp', () => {
+    expect(getHistoryEntryKey({ foo: 'bar' })).toBeUndefined()
+  })
+
+  test('installing stamps the initial entry with a key', () => {
+    const key = getHistoryEntryKey()
+    expect(typeof key).toBe('string')
+  })
+
+  test('push with null state to a different pathname mints a fresh key', () => {
+    const firstKey = getHistoryEntryKey()
+
+    history.pushState(null, '', '/b')
+
+    const secondKey = getHistoryEntryKey()
+    expect(secondKey).toEqual(expect.any(String))
+    expect(secondKey).not.toBe(firstKey)
+  })
+
+  test('push with null state to the same pathname inherits the current key', () => {
+    const firstKey = getHistoryEntryKey()
+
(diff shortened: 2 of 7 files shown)


2Pacalypse-:
Here's a more detailed Claude's description of what the PR does:

>This PR makes user profile pages remember where you were scrolled to, so that pressing Back (or Forward) puts you right back where you left off instead of at the top. Here's how the pieces fit together, in plain terms.
>
>The problem it solves
>
> Normally when you scroll down someone's profile, click a match, and then hit Back, the profile loads scrolled to the top — you lose your place. Browsers do this restoration automatically for normal pages, but ShieldBattery is a single-page app with its own scrolling containers, so the browser can't help. This PR rebuilds that behavior by hand.
>
>The key idea: one "visit key" per history entry
>
>The subtle part is deciding when to restore. If you scroll down Alice's profile, then later click a fresh link to Alice's profile, you should start at the top — that's a new visit. But hitting Back to the same profile should restore. The pathname alone can't tell those two apart, so the PR keys everything by history entry (each spot in the browser's back/forward stack), which is exactly how browsers do their own scroll restoration.
>
>history-entry-key.ts (new file) makes that possible. At app startup it wraps history.pushState and history.replaceState so that every new history entry gets stamped with a random unique ID (the "visit key") stored in history.state. Rules for the stamping:
>
>- A push to a different pathname gets a brand-new key — new visit, fresh scroll.
>- A push that keeps the same pathname (changing only search params, or opening an overlay on top of the page) reuses the current key — you're still on the same visit, scroll continuity holds.
>- Any replace reuses the current key too — this covers things like the URL being corrected to a username's canonical casing.
>- Some parts of the app put their own value into history.state (sentinels that mark, say, an open settings overlay). Those are left completely untouched — no stamp — so the code that compares those values by identity keeps working. Such entries simply have no key, and scroll memory does nothing while one is current.
>
>A popstate listener keeps track of which key is "current" as you traverse back/forward.
>
>The storage: view-state-store.ts (new file)
>
>A small module-level map (deliberately outside React/Redux/Jotai — writing to it must never cause a re-render) that holds transient per-view state. It's shared and bounded:
>
>- Max 50 entries total across all features, evicting the least-recently-used one when full.
>- Each feature gets its own namespace ("scroll" here) so keys can't collide.
>- Entries expire after a configured age — 30 minutes for scroll positions — so a stale position from an hour ago doesn't get restored.
>
>It's built generically because later PRs (per the chat roadmap) will store more than scroll positions in it — message windows, drafts, etc.
>
>The hook: useScrollMemory in router-hooks.ts
>
>This ties the two together. A page passes it a ref to its scrolling element. The hook:
>
>1. Looks up the current visit key.
>2. In a layout effect (so it happens before paint — no visible jump), checks the store for a saved position under that key and scrolls there if found.
>3. Listens to scroll events to keep track of the latest position.
>4. On cleanup (navigating away), saves that last-seen position under the visit key.
>
>Two careful details:
>
>- The saved value comes from the scroll listener, not from reading scrollTop at cleanup time. Reason: when you navigate between two profiles, React reuses the same mounted component, so by the time cleanup runs the new page's (possibly shorter) content is already in the DOM and scrollTop may have been clamped down. The listener's last value is always from before that swap.
>- The restore runs in the page's layout effect, which fires after the container's built-in "reset to top on navigate" effect — so the restore wins over the reset, with no flicker.
>
>The consumer: user-profile.tsx
>
>The smallest part: the profile page creates a ref, passes it to useScrollMemory, and attaches it to its CenteredContentContainer scroller. That's the whole integration — three lines. And index.jsx calls installHistoryEntryKeys() once at startup so the stamping is active before any navigation happens.
>
>There are also two test files covering the entry-key stamping rules and the store's LRU/expiry behavior.
>
>In one sentence: every spot in the back/forward stack gets a unique ID, the profile page saves its scroll position under that ID when you leave, and restores it before paint when you come back to that same spot — while fresh link clicks always start at the top, just like native browser behavior. Per the roadmap, this is the foundation piece; the next step is applying the same mechanism to the ladder's virtualized table.

PR #1446, opened 2026-08-30
Persist chat reading positions and dividers across reloads.

2Pacalypse-:
Stacked on #1445. Makes the reading-position restore survive a reload or app restart.

- One localStorage entry per user (`{userId}|chatRestore`) maps place keys to `{anchor, unreadLineTime?, savedAt}`, written when a conversation is left and on `beforeunload` (the message list owns the unload save and calls back after recording, avoiding listener-ordering assumptions). Same 45-minute TTL as the in-memory store, capped at 20 places, oldest evicted.
- A conversation left at the bottom deletes its entry; an entry is never written without an anchor — a divider alone would be consumed the moment its conversation opened.
- Hydration is idempotent and happens both where the list first asks for its position (layout phase) and in activation (passive phase); the restored divider rides the activate action and obeys the same freeze/consume rules as one frozen in-session.
- The persist thunks bail unless the place is actually activated, so a cold load cannot wipe the entry it is about to restore. Logging out clears the in-memory view state so another account cannot inherit positions.

Live-verified: reload mid-backlog restores the exact anchor and the divider (with the server read position already past it) via a single `aroundTime`; catching up and reloading pins to the bottom with the entry removed.

### DIFF
--- a/client/auth/action-creators.ts
+++ b/client/auth/action-creators.ts
@@ -13,6 +13,7 @@ import { DialogType } from '../dialogs/dialog-type'
 import { dispatch, type ThunkAction } from '../dispatch-registry'
 import { maybeChangeLanguageLocally } from '../i18n/action-creators'
 import logger from '../logging/logger'
+import { clearViewState } from '../navigation/view-state-store'
 import { RequestHandlingSpec, abortableThunk } from '../network/abortable-thunk'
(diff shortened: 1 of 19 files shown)


2Pacalypse-:
Not worth.

PR #1447, opened 2026-08-30
Persist chat and whisper message drafts across reload via localStorage.

2Pacalypse-:
Unsent message input already survived conversation switches in-session; it now also survives a reload or app restart.

One localStorage entry per user (`{userId}|chatDrafts`) keyed by the existing per-place storage keys. Keystroke bursts coalesce into a single write (300ms trailing edge), flushed on unmount and `beforeunload`; sending or clearing deletes the entry rather than storing an empty string. Entries are pruned on write (empties dropped, capped at 50 places, oldest first). Lobby chat stays draft-free as before.

Live-verified: a typed draft survives a full reload and is restored into the input; clearing it removes the stored entry.


2Pacalypse-:
Not worth.

PR #1500, opened 2026-09-09
Play a rare transmission-glitch burst on the splash logotext.

intrepidcanadian:
A temporary, easily-reverted visual experiment — proposed as a **one-month trial** to see whether it draws interest to the landing page, after which it can be kept, tuned, or reverted (it's two isolated CSS-keyframe changes with no logic or dependencies).

## What it does

- **Splash logotext**: every ~7 seconds the "ShieldBattery" wordmark plays a ~400ms transmission-glitch burst — a chromatic split in the brand's blue and amber plus two horizontal slices that tear out and snap back. It idles the other 93% of the cycle so it reads as a rare flicker of interference rather than a looping animation. Hovering the wordmark speeds the cycle up.
- **Bottom-links gear logo**: the same interference, but **hover-only**, since that logo also sits on surfaces logged-in players see constantly.

Both are pure CSS on layered copies of the existing SVGs (no new assets, no JS timers), are `aria-hidden` on the duplicate layers, and are fully disabled under `prefers-reduced-motion`.

## Screenshots

Mid-burst frames of the splash logotext:

Splash logotext glitch burst, slice displaced right
Splash logotext glitch burst, bottom slice tearing left

The gear logo while hovered:

Gear logo glitching on hover

## Verification

Verified live against the dev server (T0 + T1): lint/typecheck clean, burst frames inspected by pausing the animations at multiple points in the cycle, hover behavior confirmed on both the wordmark and the gear, and layout confirmed unchanged at rest.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/client/home/bottom-links.tsx
+++ b/client/home/bottom-links.tsx
@@ -1,5 +1,5 @@
 import { useTranslation } from 'react-i18next'
-import styled from 'styled-components'
+import styled, { keyframes } from 'styled-components'
 import { Link } from 'wouter'
 import BlueskyLogo from '../icons/brands/bluesky.svg?react'
 import DiscordLogo from '../icons/brands/discord.svg?react'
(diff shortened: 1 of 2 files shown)


intrepidcanadian:
it would just be like... "Why is Shield Battery logo glitching?"

a fun feature for like a month or so.

2Pacalypse-:
Cool idea, but without having something behind the reason for glitching, it is just a gimmick.

intrepidcanadian:
sometimes you need to send a dragoon to fake the dark templar

PR #1513, opened 2026-09-11
Add command and argument autocomplete to the chat input.

### LINKED ISSUE #1479: Command autocomplete and signature help
## Why

Slash commands are only usable in practice if people can discover them and get their arguments
right without memorizing syntax. The unified typeahead and signature help were designed by 2Pac,
2026-07-10, and refined through a mockup review and feedback rounds with 2Pac on
2026-09-03 that fixed the palette on the existing Material popover and moved signature help into
the text field's own supporting-text strip. This issue follows the command framework and the
`/me` emote work in the delivery order, and it provides the single typeahead surface (command
names, then arguments) that the later user-lookup, friend/block/moderation, and
whispers-everywhere commands plug into through their argument providers.

## Current behavior

Written against c104da263 (2026-09-07).

No command-related code exists on master: there is no `client/messaging/commands/` directory, no
`runChatCommand`, no `CommandArg` type, and no signature-help component anywhere in the client
(a repository-wide search for each of these at this commit returns nothing; the only hit for
"slash command" is an unrelated comment in `game/src/bw_scr/dialog_hook.rs` about the in-game
chat box). Autocomplete today is two independent, hardcoded suggestion features living entirely
inside `client/messaging/message-input.tsx`.

**Mention typeahead.** `MessageInput` holds four pieces of state for it: `userMentionStartIndex`,
`userMentionMatchedText`, `matchedUsers`, and `virtuallyFocusedMentionIndex`
(`message-input.tsx:194-197`), plus a single `UFuzzy` instance shared with the emote matcher
(`new UFuzzy({ intraIns: Infinity, intraChars: '.' })`, line 206). A `selectionchange` listener on
the underlying `<textarea>` (added in a `useEffect`, lines 235-338) re-derives the suggestion list
on every caret move: it first checks for a bare `@` with nothing typed after it (using the
`baseMentionableUsers` prop, lines 275-286), otherwise it locates the start of the current
`@word` by searching `message.slice(0, selectionStart)` for `/(?<=^|\s)@\S*$/` (line 290), parses
it with the shared `matchUserMentions` regex from `common/text/user-mentions.ts` (the same regex
used to render mentions in sent messages), and fuzzy-filters `mentionableUsers` by name with
`fuzzy.filter(...)` (lines 307-311), capped to `MAX_MENTIONED_USERS` (10, line 45). Results render
in a `Popover` anchored to the input container's top-left corner minus 8px
(`useElemAnchorPosition(containerElem, 'left', 'top')`, line 210, `anchorY={(anchorY ?? 0) - 8}`,
`originX='left' originY='bottom'`, lines 549-576) so it opens upward above the input, with
`focusOnMount={false}` so real DOM focus stays in the textarea. Inside it, a `MenuList` with
`virtualFocus` tracks a visual-only active index through `onActiveIndexChange` (no real DOM focus
moves), and each `MenuItem` shows a `ConnectedAvatar`, fading both the icon and the text when
`!user.online` (lines 78-101, 566-574). Accepting a suggestion (Enter, checked in
`onEnterKeyDown` at lines 420-427, or Tab, in a `TextField` `onKeyDown` handler at lines 530-544,
or a click) replaces the matched `@partial` span with `@name ` and repositions the caret. Escape
dismissal is not custom code here; it comes for free because the shared `Popover` component
already closes on `Escape` (`client/material/popover.tsx`, its own key listener handles the key).
Neither `Popover` is given an ARIA `role`, and `MenuList`/`MenuItem`
(`client/material/menu/menu.tsx`, `client/material/menu/item.tsx`) render no `role` or `aria-*`
attributes at all, so today's typeahead has no combobox/listbox semantics.

Note: `client/messaging/mention-hooks.tsx` does not implement any part of this typeahead. It only
exports `useMentionFilterClick`, a hook used by rendered messages so shift-clicking a username
mentions that user in the compose box instead of opening their profile overlay. The typeahead
trigger detection, matching, and rendering described above live entirely in
`message-input.tsx`.

**Emote-shortcode suggestions.** `client/messaging/emote-suggestions.ts` is a second, independent
matcher: `EMOTE_QUERY_REGEX` (`/(?<=^|\s):(?<query>[\w+-]{2,})$/`, line 16) requires a `:`
preceded by whitespace or the start of the message, followed by at least two more characters.
`searchUnicodeEmojis` (lines 93-153) tiers matches as exact, prefix, substring, then fuzzy (the
fuzzy tier reuses the same `UFuzzy` settings as mentions), and `orderEmoteSuggestions`
(lines 159-167) sorts by that tier and then by how often each emoji has been inserted before,
tracked in `localStorage` under `sb.emoteUsage` (`recordEmoteUsage`/`getEmoteUsage`, lines 55-68),
capped at `MAX_EMOTE_SUGGESTIONS` (10, line 5). `MessageInput` wires this into the same
`selectionchange` effect (lines 248-269): on a match it loads `getUnicodeEmojiEntries()`
asynchronously, guarding against the query having changed by the time it resolves
(`latestEmoteQueryRef`), then opens a second, separately-stated `Popover`/`MenuList` pair
(`emotesOpen`/`matchedEmotes`/`focusedEmoteIndex`, lines 199-202, 578-604) built from the
identical `Popover` + `MenuList` + `MenuItem` shapes and anchored the same way as the mention
popover, showing the emoji character as the row icon.

**Coexistence.** Mentions and emotes are two entirely separate state machines with no shared
controller; both are re-evaluated on every `selectionchange` inside one effect. They cannot both
have a match at once in practice, since a caret position can only be preceded by one of `@\S*$`
or `:[\w+-]{2,}$`, but nothing in the code enforces that structurally. Where both
`onEnterKeyDown` (lines 411-457) and the `Tab` handler (lines 530-544) check state, the emote
check always runs first, so if both were ever open at once the emote suggestion would win.

**TextField's reserved supporting-text strip.** `client/material/text-field.tsx` renders an
`InputError` (`client/material/input-error.tsx`) below the field whenever `allowErrors` is true,
which is the default (`text-field.tsx:246`, rendered at `text-field.tsx:459`). `InputError`'s
container is a `min-height: 20px` div with `padding: 2px 12px` (`input-error.tsx:7-8`) that shows
`errorText` in `--theme-error` `bodySmall` with a fade/slide transition, and reserves the 20px
even with no error text to show. `MessageInput`'s `StyledTextField` passes `allowErrors={false}`
(`message-input.tsx:509`), so the chat input opts out of this strip entirely today: there is no
reserved space below the input and no supporting-text slot currently rendering for it.

## Desired behavior

- One typeahead controller with pluggable providers replaces the two hardcoded state machines in
  `MessageInput`: `@` keeps triggering the existing user provider unchanged; `/` at the very start
  of the input triggers a command provider that fuzzy-matches command names and aliases; once the
  input parses as a recognized command, further typing triggers an argument provider driven by
  that argument's `CommandArg.suggest` (for example `/kick` suggests channel members, `/join`
  suggests known channels).
- The command palette and the argument palette render through the same Material `Popover` +
  `MenuList`/`MenuItem` combination the mention typeahead already uses (same anchoring, same
  virtual focus behavior): no bespoke row components and no key-hint header row above the list.
- A command that does not exist on the current surface (a moderation command typed in a whisper)
  is omitted from the palette. A command that exists here but cannot run for this user (`/kick`
  for a non-moderator) is listed greyed out with the reason the registry's `getUnavailableReason`
  supplies, the same treatment the `/help` dialog gives it; selecting it completes the name, and
  submitting prints that reason.
- Signature help, the command's derived usage string with the currently-active parameter
  emphasized, renders inside the `TextField`'s reserved supporting-text strip (the slot
  `InputError` occupies today), not a separate bar above the input. It advances to the next
  parameter as the user types past a space and stays in sync with the palette's own selection.
- Argument suggestions for a `user` argument use the same online/offline styling the mention
  typeahead already applies (faded avatar and text for offline users).
- Keyboard: ArrowUp/ArrowDown move the active row, Tab or Enter accepts the active row, Escape
  dismisses the open palette without submitting the message, and Space advances signature help to
  the next parameter (accepting an unambiguous argument suggestion first, if one is active),
  matching the existing Tab/Enter acceptance pattern used for mentions.
- The open palette, command or argument, exposes ARIA combobox/listbox semantics: the input
  carries `role="combobox"` with `aria-expanded` and `aria-activedescendant`, and the open list
  carries `role="listbox"` with each row `role="option"`, kept in sync with the active row.
- With no palette open, sending a message behaves exactly as it does today. Resolving aliases,
  rejecting an unresolved `/foo`, and running the command are the framework issue's
  responsibility; this issue only ships the discovery and typing UI.

## Decisions

- Unavailable commands are shown greyed out with their reason, matching `/help`; only commands
  that do not exist on the surface are left out. Revised by 2Pac, 2026-09-11 (the 2026-07-10 lock
  hid them entirely).
- The command and argument palettes reuse the mention typeahead's exact Material `Popover` +
  `MenuList`/`MenuItem` shape, with no custom row rendering and no key-hint header row. Locked by
  2Pac, 2026-09-03 (mockup review, confirmed across the following feedback rounds the same day).
- Signature help renders inside the `TextField`'s reserved supporting-text strip rather than a
  separate bar pinned above the input. Locked by 2Pac, 2026-09-03.
- Working direction from the 2026-09-03 design review (not formally locked): autocomplete ships
  after the command framework and the command catalog are in place, not before.
- Rejected: hiding unavailable commands from the palette as if they did not exist (2026-07-10,
  reversed 2026-09-11; see Decisions).
- Rejected: a boxed signature-help bar pinned above the input, because it would shift layout near
  the chat's at-bottom scroll logic. 2026-09-03 mockup review.
- Rejected: custom-built palette rows for commands and arguments, in favor of reusing the existing
  mention popover components as-is. 2026-09-03.

## Verification

Tier: T0 (lint, typecheck, unit tests for the provider and typeahead logic) plus T2, driven with
the real Electron app (verify-app skill), since this is client-only UI with no cross-user
interaction to check.

Recipe: with a channel, a whisper, and a lobby reachable from one client, and both a moderator and
a non-moderator view of the same channel (a second account, or a permission toggle), type `/`
alone and confirm the palette lists the commands that exist on the current surface: moderation
commands normal as a moderator, greyed out with the no-permission reason as a plain member;
whisper-only or channel-only commands absent on the wrong surface. Type `/w` and confirm it
narrows to whisper-related commands and aliases by fuzzy match. Type `/kick ` (or another command
with a `user` argument) in a channel where the account holds moderation permission and confirm
the argument palette lists channel members with the same online/offline styling as `@mentions`,
and that the supporting-text strip below the input shows the derived usage with the `user`
parameter emphasized, advancing to the next parameter after a space. Drive the palette with
ArrowUp/ArrowDown, Tab, Enter, and Escape and confirm each behaves as specified. Use the
verify-app skill's DOM snapshot capability, not screenshots, to confirm `role="combobox"`,
`aria-expanded`, and `aria-activedescendant` on the input, and `role="listbox"` with
`role="option"` rows on the open palette. Finally, close the palette and send both an ordinary
message and a message starting with an unresolved `/foo` to confirm neither is affected by
anything this issue adds.

## Out of scope

#1477 owns the command registry, the
typed-argument schema, the parser, and alias resolution that this issue's providers read from.

#1484 owns the visual design of command answer cards and
the `/help` dialog; this issue only needs the palette and signature-help chrome, not those answer
surfaces.

## Links

None beyond the parent issue.

2Pacalypse-:
Slash commands are only usable if people can discover them and get their arguments right without memorizing syntax. This replaces the two hardcoded typeaheads in `MessageInput` (mentions and emotes) with one controller and pluggable providers, and adds two more providers on top of the command framework: a command palette for `/` at the start of the input and an argument palette driven by the argument the caret is in. The palette reuses the mention typeahead's Material `Popover` + `MenuList`, carries combobox/listbox ARIA semantics, and lists commands the way the `/help` dialog does.

Fixes #1479

## Acceptance criteria

- [x] One typeahead controller with pluggable providers replaces the two hardcoded state machines; `@` keeps triggering the user provider unchanged, `/` at the start triggers the command provider (fuzzy on names and aliases), and a recognized command triggers the argument provider through `CommandArg.suggest` (or the kind default). Verified in the app: `/` lists the surface's commands, `/w` narrows to whisper via its alias, `/kick ` offers channel members, `/join ` offers joined channels; `@cl` and `:fir` still open their palettes and Enter/Tab accept as before. Unit tests cover provider matching, ranking tiers, caret location and kind defaults.
- [x] Command and argument palettes render through the same `Popover` + `MenuList` as mentions, same anchoring and virtual focus, no key-hint header. Command rows use the help dialog's presentation (shared `CommandUsage`/`CommandAliases` pieces): amber name, dimmed arguments, alias chips, description in an aligned second column. Verified in the app by screenshot.
- [x] A command that does not exist on the surface is omitted, and so is one that exists but cannot run there (the registry's `getUnavailableReason`): the name palette, the argument palette and the `/help` dialog all leave it out, while typing it out and submitting still prints the reason as an only-you line (runner path unchanged, unit-tested). Verified in the app: as a plain member of #modtest26, `/` lists help/join/whisper/leave/me, `/ki` narrows to nothing, `/kick ` opens no argument palette and `/help` omits kick and ban; as the channel owner `/kick` and `/ban` are listed, `/kick ` offers the members and `/help` lists them. `/close` is absent in channels and lobbies and present in whispers; `/kick`, `/ban`, `/leave` are absent in whispers; `/leave` is present in lobbies.
- [x] `user` argument suggestions use the mention typeahead's online/offline styling. Verified in #ShieldBattery: `/kick cl` lists offline members with the faded text color and avatar opacity 0.38, identical to the `@cl` rows.
- [x] `/whisper` (and its aliases) suggests real targets: open whisper sessions (most recent first), then friends (online first), then the channel's members, each once and never yourself. Verified as claude-1 in #ShieldBattery: `/w ` lists claude-2, claude-admin, claude-3 (sessions) then test5, 2Pacalypse-, claude-10… (members, claude-1 absent); in a whisper and a lobby the sessions and friends are still offered.
- [x] Keyboard: ArrowUp/ArrowDown move the active row (wrapping), Tab or Enter accept, Escape dismisses without submitting, Space accepts an unambiguous argument suggestion first. Verified in the app for each key; Escape left `/join ` in the input and sent nothing.
- [x] ARIA: the textarea carries `role="combobox"`, `aria-expanded`, `aria-controls` and `aria-activedescendant`; the open list is `role="listbox"` with `role="option"` rows whose `aria-selected` follows the active row. Verified through DOM evaluation: `aria-activedescendant` tracked `<listId>-0` → `-1` → `-6` as the highlight moved, and exactly one row had `aria-selected="true"` at each step.
- [x] With no palette open, sending behaves as today; unresolved `/foo` stays the framework's business. Verified: "hello from admin" was sent as an ordinary message and `/foo bar` printed "Unknown command /foo. Type /help for a list of commands." with the palette closed the whole time.

## Dropped in review

- Signature help (the command's usage in the `TextField` supporting-text strip with the active part emphasized) is gone: it made the input jump in height and the palette already shows the same usage. The `supportingText` prop it needed on `TextField`/`InputError` is gone with it, and `ArgCaret` no longer carries a signature.
- Greyed rows for commands the user can't run (lock icon and reason under the description, in both palettes and the `/help` dialog) are gone too: those commands are hidden instead, so `CommandUnavailableReason` and the reason fields on the dialog payload and the palette rows went with them.

## Drift from the issue's pinned commit

The issue was written before the framework (#1510) and `/me` (#1512) merged. `client/messaging/commands/` exists on master with the schema, parser, registry, runner and `/help` dialog, and `MessageInput` already takes a `commands` prop; the palette lists commands the way the `/help` dialog does. `CommandArg` had no `suggest` field and `CommandUserEntry` had no `online` flag; both are added here. The mention and emote matchers no longer shared one UFuzzy instance.

## Calls made

- `suggest` is an optional per-argument override on `CommandArg`; the kind decides the default (`user` → the surface's members, `channel` → known channels with joined ones first, `enum`/`subcommand` → their own values, other kinds nothing). `/whisper` is the first command to use it; `/unban` and `/unfriend` will.
- `ArgSuggestion.user.online` and the palette's user visual are optional booleans: a whisper-session target has no tracked presence, so its row is neither marked online nor faded.
- `CommandUserEntry` gains `online` so the argument palette can fade offline users; `channel.tsx` already passed entries that carry it.
- Command rows are a dedicated `CommandMenuItem` (built on the now-exported `MenuItemButton` chrome from `material/menu/item.tsx`) inside a grid `MenuList` variant with subgrid rows, so usage and description columns line up across rows the way the dialog's do. Every other palette keeps the plain `MenuItem` rows.
- `MessageInput` is now a plain function component taking `ref` as a prop instead of `React.forwardRef`. react-dom 19.2 only refreshes `useEffectEvent` handlers on plain function components; inside the `forwardRef` wrapper the `selectionchange` handler stayed frozen at its mount-time closure, so every palette (mentions included) only ever offered the users that existed when the input mounted. That is what made `/whisper` look like it only knew about yourself on a fresh channel visit.
- Enter on an exact match (`/help`, `/kick claude-1`) sends instead of re-accepting, in the command and argument palettes only; mentions keep their existing Enter behavior. Space accepts a single suggestion only once something has been typed towards it (never right after the bare `/` or an empty argument) and never when the typed text already is the suggestion.
- Mentions and emotes became providers of the same controller and render through one popover instead of two.
- Every palette caps at 10 rows, as today; `/help` remains the full list.
- A command the user can't run here gets no argument suggestions either, since it isn't offered by name; submitting it is what tells them why. The surface-and-permission filter is one `getRunnableCommands` helper shared by both palettes and `/help`, and `locateCommandCaret` searches whatever commands it is handed.

## Verification

T0: `pnpm run lint`, `pnpm run typecheck`, `pnpm exec vitest run client common`, `pnpm run check-circular` pass; `pnpm run gen-translations` changed nothing.

T2 (verify-app, Electron instances as claude-1 / claude-admin over playwright-cli): everything listed under the acceptance criteria above, driven with playwright-cli and inspected through DOM evaluation plus screenshots of the command palette and the `/help` dialog. The stale-palette regression was reproduced on a fresh app reload (first visit to #ShieldBattery: `/w `, `/kick ` and `@cl` all offered only claude-1 while the input's props already carried 17 users) and confirmed fixed the same way after the `forwardRef` removal. Re-driven after hiding gated commands, as claude-1 (plain member) and claude-admin (owner) in #modtest26: the name palette, the argument palette and the `/help` dialog agree on what each account can run.


client/messaging/commands/command-provider.ts
```diff
@@ -0,0 +1,174 @@
...
+
+      return {
+        start: caret.argStart + token.start,
+        matchedText: token.text,
+        suggestions,
+        submitOnExact: true,
+        // An argument nothing has been typed for yet is being skipped over, not completed, when
+        // the user presses space; a single row is only worth accepting once it has been typed
+        // towards.
+        spaceAcceptsSingle: query.length > 0,
```
tec27: **Major:** Space and Enter accept a single fuzzy row over a fully typed, valid target that is merely not in the pool. `exact` only compares against pool values, and whisper targets / channel names are open sets (anyone on the server, any channel), so a complete literal name is routinely "not exact" while one pool entry prefix- or fuzzy-matches it.

Scenario: #ShieldBattery has a member `ZergRush`. Type `/w Zerg` (meaning a user `Zerg` who is not a member, friend or open session) and press Space to start the message. One row (`ZergRush`), not exact, `spaceAcceptsSingle` true, so the token is replaced: `/w ZergRush `. Eyes are on the message; `gg wp` + Enter goes to the wrong person. `/join sb` + Space becomes `/join sb-dev ` the same way. Enter has the same root cause (`/w Zerg` + Enter rewrites instead of opening the session), though that one is visible before anything is sent. The existing test at `command-provider.test.ts:215-220` encodes exactly this path.

Fix: only let Space accept when the value set is closed (`enum`, `subcommand`, and `user` args whose runner resolves strictly against `members`, i.e. kick/ban); never for `channel` or for an open `suggest` pool like whisper's. Consider also letting Enter send rather than swap when the typed token already passes `isValidUsername`/`isValidChannelName`.

client/messaging/message-input.tsx
```diff
@@ -175,468 +219,420 @@ export interface MessageInputProps {
...
+        matchedText: match.matchedText,
+        suggestions,
+        submitOnExact: !!match.submitOnExact,
+        spaceAcceptsSingle: !!match.spaceAcceptsSingle,
       })
+
+      if (suggestions.length > 0) {
+        openPalette(event)
+      } else {
+        closePalette()
```
tec27: **Major:** the parent `activeIndex` survives a close that isn't a dismissal, so the accepted row and the highlighted row diverge on the next open. `setActiveIndex(0)` only runs when the provider id changes; this `closePalette()` (and the no-match one above at the `!result` branch) leaves the index alone. `Popover` unmounts its subtree after the exit animation (`popover.tsx:231`), so `MenuList` comes back with its own index at 0 and only reports changes via `onActiveIndexChange`, while Enter/Tab and `aria-activedescendant` keep using the stale parent value.

Scenario: `@c` (5 rows), ArrowDown twice, type a space (regex no longer matches, palette closes, index still 2), wait ~0.4s, Backspace. Row 0 is highlighted, `aria-activedescendant` names row 2, Enter accepts row 2. Same via typing a char with no matches and deleting it. This flaw existed in the old `virtuallyFocusedMentionIndex` too, but this is the state machine the PR rewrote and the ARIA now reports the wrong row.

Fix: reset `activeIndex` on every close path in `onSelectionChange` (route both `closePalette()` calls through one helper that also resets), or make the parent the single source of truth and pass the index down to `MenuList`.

client/messaging/message-input.tsx
```diff
@@ -175,468 +219,420 @@ export interface MessageInputProps {
...
+        allowErrors={false}
+        maxLength={CHAT_MESSAGE_MAXLENGTH}
+        showDivider={showDivider}
+        disabled={!!chatRestriction}
+        trailingIcons={[
+          <EmotePickerButton key='emotes' disabled={!!chatRestriction} onInsert={insertAtCaret} />,
+        ]}
+        inputProps={{
+          autoComplete: 'off',
+          role: 'combobox',
```
tec27: **Minor:** `role="combobox"` (and `aria-expanded`) on a `<textarea>` is not conformant. ARIA-in-HTML allows no explicit role on `textarea` (axe `aria-allowed-role` flags it; `input[type=text]` is the element that may take `combobox`), and `aria-expanded` is not supported on the implicit `textbox` role. `aria-autocomplete`, `aria-controls` and `aria-activedescendant` are all allowed on `textbox`, so dropping `role` and `aria-expanded` keeps the useful wiring. Screen readers still voice the pattern as written, so this is conformance/lint rather than a broken experience.

client/messaging/message-input.tsx
```diff
@@ -175,468 +219,420 @@ export interface MessageInputProps {
...
-            )
-            const matchedUsers = matchedUserIndexes?.map(i => mentionableUsers[i]) ?? []
-
-            setUserMentionStartIndex(userMentionStartIndex)
-            setUserMentionMatchedText(userMention.text)
-            setMatchedUsers(matchedUsers.slice(0, MAX_MENTIONED_USERS))
+// A plain function component rather than a forwardRef one: react-dom only refreshes
+// `useEffectEvent` handlers on plain function components, and a forwardRef wrapper would leave
+// `onSelectionChange` frozen at its mount-time closure.
+export function MessageInput({
```
tec27, line 243: **Minor:** nothing exercises `MessageInput` itself; every new test is a pure-function test (`typeahead.test.ts`, `command-provider.test.ts`, `command-suggestions.test.ts`, the `locateArgAtCaret` block). Enter/Tab/Space/Escape, `submitOnExact`, `spaceAcceptsSingle` at the input level, the index reset rules and the async request guard are all uncovered, and both major findings live in this file. A small RTL test around the component with a fake provider covering the key paths and the reopen-with-same-provider case would have caught them. The caret tests also only cover an unclosed quote (not a closed one followed by the caret) and no non-ASCII names.

client/messaging/message-input.tsx
```diff
@@ -175,468 +219,420 @@ export interface MessageInputProps {
...
-            const matchedUserIndexes = fuzzy.filter(
-              mentionableUsers.map(u => u.name),
-              userMention.groups.username,
-            )
-            const matchedUsers = matchedUserIndexes?.map(i => mentionableUsers[i]) ?? []
-
-            setUserMentionStartIndex(userMentionStartIndex)
-            setUserMentionMatchedText(userMention.text)
-            setMatchedUsers(matchedUsers.slice(0, MAX_MENTIONED_USERS))
+// A plain function component rather than a forwardRef one: react-dom only refreshes
```
tec27: **Nit:** the constraint is wider than forwardRef. In react-dom 19.2.7 the before-mutation pass refreshes effect-event impls only for `case 0` (FunctionComponent); `case 11` (ForwardRef) and `case 15` (SimpleMemoComponent) both `break` (`react-dom-client.development.js:13872-13886`). Wrapping this component in `React.memo` would silently bring the frozen `onSelectionChange` back, so say so here.

client/messaging/commands/command-schema.ts
```diff
@@ -1,12 +1,32 @@
...
+   * When the value names a user: whose avatar the row shows, faded along with the text while the
+   * user is known to be offline. `online` is undefined when the client doesn't track that user's
+   * presence.
+   */
+  user?: { id: SbUserId; online?: boolean }
+}
+
+export interface ArgSuggestDeps {
+  context: CommandContext
+  /** Reads the store as of when the palette was opened; suggestions are not kept live. */
```
tec27: **Nit:** this doc is wrong: `getState` is called from `createCommandArgProvider.match` on every `selectionchange` (`command-provider.ts:135`), so suggestions are re-read per keystroke, not "as of when the palette was opened".

tec27 (commented):
**Verdict: merge after fixes.** The controller/provider refactor holds up, but two behaviours need fixing first: (1) Space (and Enter) in the argument palette accept a single fuzzy match over a fully typed, valid target that is simply not in the pool, so `/w Zerg` becomes `/w ZergRush ` on the space bar and the whisper goes to the wrong person (same for `/join`); (2) the parent `activeIndex` survives a non-dismissal close, so after the palette reopens with the same provider the highlighted row, the row Enter/Tab accept, and `aria-activedescendant` disagree. Lower priority: `role="combobox"` on a `<textarea>` is not an allowed role, and nothing tests `MessageInput` itself, which is where both bugs live.

Nits (inline): the forwardRef comment should also name `React.memo` (react-dom 19.2.7 skips SimpleMemoComponent fibers too); the `ArgSuggestDeps.getState` doc says the store is read once per palette open, but it is read on every `selectionchange`.

Verified: tsc, eslint on the changed files, vitest (71 files / 975 tests) and gen-translations (no drift) all pass on the branch; the forwardRef/`useEffectEvent` claim is confirmed against react-dom 19.2.7 source (before-mutation pass refreshes effect events for FunctionComponent only). Mention and emote behaviour is otherwise unchanged keystroke for keystroke; no semantic conflicts with master.

---
_Adversarial review by Claude (Fable 5.1) at tec27's request: one reviewer per open PR, each on a worktree at this head (3ff7553bb) with typecheck, eslint, vitest and gen-translations run._

